PR #5, opened 2018-10-06
adds a --force option to override existing projects

dariooddenino:
With this pr, `spacchetti local-setup` exits with an error in case of existing `packages.dhall` or `psc-package.json` files.

A `--force` option has been added to overwrite existing projects.

There's a bit of a lazy implementation of a `unsafePathToText` to print file paths in the error message, but I suppose it's good enough for now.


justinwoo:
Thanks

PR #9, opened 2018-10-09
clean command

dariooddenino:
Adds a `spacchetti clean` command that nukes the .psc-package folder.

I also added a clearer message in case the user runs `insdhall` without a `packages.dhall` file.


justinwoo:
thx

PR #17, opened 2018-10-25
aeson + psc-package upgrade

dariooddenino:
This pr uses aeson to generate the `psc-package.json`.

In case there's an existing `psc-package.json` and `--force` is not used, `spacchetti local-setup` keeps the existing dependencies.

I also changed a bit some output lines, but I feel like they can improved a lot.

Now the workflow to create a new project is:

```
pulp --psc-package init
spacchetti local-setup
spacchetti insdhall
psc-package install
pulp build
```

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -3,10 +3,12 @@
 module Main where
 
 import qualified Data.Text as Text
+import qualified Data.Text.Encoding as Text
 import Data.Version (showVersion)
 import qualified Paths_spacchetti_cli as Pcli
-import qualified Templates as Templates
+import qualified Templates
 import qualified Turtle as T
+import Data.Aeson
 
 -- | Commands that this program handles
 data Command
@@ -37,8 +39,10 @@ insDhall = do
   code <- T.shell ("cat ./packages.dhall | dhall-to-json --pretty > " <> packagesJson) T.empty
 
   case code of
-    T.ExitSuccess ->
-      T.echo $ T.unsafeTextToLine ("wrote packages.json to " <> packagesJson)
+    T.ExitSuccess -> do
+      T.echo $ T.unsafeTextToLine $
+        "Wrote packages.json to " <> packagesJson
+      T.echo "Now you can run `psc-package install`."
     T.ExitFailure n ->
       T.die ("failed to insdhall: " <> T.repr n)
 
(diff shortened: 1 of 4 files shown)


app/Templates.hs
```diff
@@ -1,12 +1,31 @@
...
 import Data.FileEmbed
 import System.FilePath ((</>))
+import qualified Data.Text.Lazy as LT
+import qualified Data.Text.Lazy.Encoding as LT
+import qualified Data.Text as T
+import Data.Aeson
+import Data.Aeson.TH
+import Data.Aeson.Encode.Pretty
+
+data PscPackage = PscPackage { name :: T.Text
```
justinwoo: cant we use normal formatting here instead of weird column align

app/Main.hs
```diff
@@ -51,35 +55,45 @@ unsafePathToText p = case T.toText p of
...
+  if hasPscPackage && not force
+    then do
+      pscPackage <- T.readTextFile pscPackageJsonPath
+      case eitherDecodeStrict $ Text.encodeUtf8 pscPackage of
+        Left e -> T.die $ "The existing psc-package.json file is in the wrong format: " <>
+          Text.pack e
+        Right p -> do
+          T.writeTextFile pscPackageJsonPath $
+            Templates.encodePscPackage $ p { Templates.set = "local", Templates.source = "" }
+          T.echo "An existing psc-package.json file was found and upgraded to spacchetti."
```
justinwoo: to use local package sets?

app/Main.hs
```diff
@@ -51,35 +55,45 @@ unsafePathToText p = case T.toText p of
...
-        T.die $ "Found " <> unsafePathToText p <> ": there's already a project here. "
-             <> "Run `spacchetti local-setup --force` if you're sure you want to overwrite it."
-
 localSetup :: Bool -> IO ()
 localSetup force = do
-  T.unless force $
-    checkFiles [ pscPackageJsonPath, packagesDhallPath ]
-
-  T.touch pscPackageJsonPath
+  -- packages.dhall file
```
justinwoo: seems like these commented things should be their own functions to be called

app/Templates.hs
justinwoo: this should also probbaly go in a Types module since we will probably end up needing thsi later

app/Main.hs
dariooddenino: What do you mean?

justinwoo: "was found and upgraded to use local package sets"

app/Main.hs
```diff
@@ -51,40 +56,58 @@ unsafePathToText p = case T.toText p of
...
+      case eitherDecodeStrict $ Text.encodeUtf8 pscPackage of
+        Left e -> T.die $ "The existing psc-package.json file is in the wrong format: " <>
+          Text.pack e
+        Right p -> do
+          T.writeTextFile pscPackageJsonPath $
+            Templates.encodePscPackage $ p { Types.set = "local", Types.source = "" }
+          T.echo "An existing psc-package.json file was found and upgraded to use local package sets."
+          T.echo $ "It's possible that some of the existing dependencies are not in the default " <>
+                   "spacchetti package set."
+          T.echo ""
```
justinwoo: Why the extra

app/Main.hs
```diff
@@ -51,40 +56,58 @@ unsafePathToText p = case T.toText p of
...
+    then do
+      pscPackage <- T.readTextFile pscPackageJsonPath
+      case eitherDecodeStrict $ Text.encodeUtf8 pscPackage of
+        Left e -> T.die $ "The existing psc-package.json file is in the wrong format: " <>
+          Text.pack e
+        Right p -> do
+          T.writeTextFile pscPackageJsonPath $
+            Templates.encodePscPackage $ p { Types.set = "local", Types.source = "" }
+          T.echo "An existing psc-package.json file was found and upgraded to use local package sets."
+          T.echo $ "It's possible that some of the existing dependencies are not in the default " <>
```
justinwoo: Should just be one long string

app/Main.hs
dariooddenino: Too much clutter, same as below

justinwoo: below??

PR #18, opened 2018-10-26
Added override & add instructions and examples

JordanMartinez:
It's a bit wordy, but I don't think anyone will need to ask questions afterwards.


justinwoo:
Thanks!

PR #19, opened 2018-10-27
implement init, install, sources, build - wip/test

justinwoo:


justinwoo:
```
[bash:~/Code/test-spacc] 普 spacchetti init
Set up a local Spacchetti project.
Try running `spacchetti install`
[bash:~/Code/test-spacc] 普 spacchetti install
generating spacchetti.json
installing 3 dependencies.
console already installed.
effect already installed.
prelude already installed.
installation complete.
[bash:~/Code/test-spacc] 普 spacchetti build
Compiling Record.Unsafe
Compiling Data.Symbol
Compiling Type.Data.RowList
Compiling Type.Data.Row
Compiling Data.NaturalTransformation
Compiling Data.Boolean
Compiling Control.Semigroupoid
Compiling Data.Show
Compiling Control.Category
Compiling Data.Void
Compiling Data.Unit
Compiling Data.HeytingAlgebra
Compiling Data.Semiring
Compiling Data.Semigroup
Compiling Data.Ring
Compiling Data.CommutativeRing
Compiling Data.Eq
Compiling Data.BooleanAlgebra
Compiling Data.Ordering
Compiling Data.EuclideanRing
Compiling Data.Ord.Unsafe
Compiling Data.Ord
Compiling Data.DivisionRing
Compiling Data.Field
Compiling Data.Function
Compiling Data.Bounded
Compiling Data.Monoid
Compiling Data.Functor
Compiling Control.Apply
Compiling Control.Applicative
Compiling Control.Bind
Compiling Control.Monad
Compiling Prelude
Compiling Data.Monoid.Additive
Compiling Data.Semigroup.First
Compiling Data.Semigroup.Last
Compiling Effect
Compiling Data.Monoid.Disj
Compiling Data.Monoid.Conj
Compiling Data.Monoid.Multiplicative
Compiling Data.Monoid.Dual
Compiling Data.Monoid.Endo
Compiling Effect.Uncurried
Compiling Effect.Class
Compiling Effect.Console
Compiling Effect.Unsafe
Compiling Test.Main
Compiling Effect.Class.Console
Compiling Main
Build succeeded.
[bash:~/Code/test-spacc] 普 node -e "require('./output/Main').main()"
Hello Spacchetti!
```

justinwoo:
@f-f

app/Main.hs
```diff
@@ -99,13 +123,166 @@ makePscPackage force = do
...
+makeSpacchetti :: Bool -> IO ()
+makeSpacchetti force = do
+  T.unless force $ do
+    hasSpacchettiDhall <- T.testfile spacchettiDhallPath
+    T.when hasSpacchettiDhall $ T.die
+       $ "Found " <> unsafePathToText spacchettiDhallPath <> ": there's already a project here. "
+      <> "Run `spacchetti initSpacchetti --force` if you're sure you want to overwrite it."
+  T.touch spacchettiDhallPath
+  T.writeTextFile spacchettiDhallPath Templates.spacchettiDhall
+  T.void $ T.shell ("dhall format --inplace " <> spacchettiDhallText) T.empty
```
f-f: Let's depend on `dhall` so that we don't have to have it installed (way less friction for the user)

app/Main.hs
```diff
@@ -99,13 +123,166 @@ makePscPackage force = do
...
+makeSpacchettiJson :: IO ()
+makeSpacchettiJson = do
+  T.mktree $ T.fromText spacchettiDir
+  code <- T.shell dhallToJsonCmd T.empty
+  case code of
+    T.ExitFailure n -> T.die ("making spacchetti.json failed: " <> T.repr n)
+    T.ExitSuccess -> pure ()
+
+readSpacchettiJson :: IO Types.SpacchettiConfig
+readSpacchettiJson = do
```
f-f: We don't need to go through JSON if we depend on `dhall`

app/Main.hs
```diff
@@ -99,13 +123,166 @@ makePscPackage force = do
...
+      { Types.version=version
+      , Types.repo=repo
+      }
+  )
+  = spacchettiDir <> name <> "/" <> version
+
+getGlobs :: [(Types.PackageName, Types.PackageDefinition)] -> [Text.Text]
+getGlobs = map fn
+  where
+    fn pair = (getDir pair) <> "/src/**/*.purs"
```
f-f: Do we want to hardcode the globs? I think it would make sense to make them configurable in the `spacchetti.dhall`

app/Main.hs
```diff
@@ -99,13 +123,166 @@ makePscPackage force = do
...
+      T.mktree . T.fromText $ dir
+      Dir.withCurrentDirectory (Text.unpack dir) $ do
+        let
+          cmd = Text.intercalate " && "
+            [ "git init"
+            , "git remote add origin " <> repo
+            , "git fetch origin " <> version
+            , "git -c advice.detachedHead=false checkout FETCH_HEAD"
+            ]
+        code <- T.shell cmd T.empty
```
f-f: Not sure if we want to print out all `git` output?

app/Main.hs
```diff
@@ -1,35 +1,59 @@
...
   -- | Do the Ins-Dhall-ation of the local project setup, equivalent to:
   -- | ```sh
   -- | NAME='local'
   -- | TARGET=.psc-package/$NAME/.set/packages.json
-  -- | mkdir -p .psc-package/$NAME/.set
+  -- | mktree -p .psc-package/$NAME/.set
   -- | dhall-to-json --pretty <<< './packages.dhall' > $TARGET
   -- | echo wrote packages.json to $TARGET
   -- | ```
   | InsDhall
```
f-f: How about renaming this (or aliasing if you want to keep backwards compatibility for a while) to `psc-package-install`?

app/Main.hs
```diff
@@ -1,35 +1,59 @@
...
   -- | Deletes the .psc-package folder
   | Clean
```
f-f: This is not strictly psc-package-related, if we clean also the `.spacchetti` dir

app/Main.hs
```diff
@@ -99,13 +123,166 @@ makePscPackage force = do
...
+  ( Types.PackageName name
+  , Types.PackageDefinition
+      { Types.version=version
+      , Types.repo=repo
+      }
+  )
+  = spacchettiDir <> name <> "/" <> version
+
+getGlobs :: [(Types.PackageName, Types.PackageDefinition)] -> [Text.Text]
+getGlobs = map fn
```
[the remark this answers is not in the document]
f-f: What do you mean?

app/Main.hs
justinwoo: Sounds horrible for our build times, and the other commands already use dhall executables. You'll have to depend on both Dhall and Dhall-JSON if you want to do this

app/Main.hs
```diff
@@ -57,13 +81,13 @@ unsafePathToText p = case T.toText p of
...
+makePackagesDhall :: Bool -> IO ()
+makePackagesDhall force = do
   T.unless force $ do
-    hasDhall <- T.testfile packagesDhallPath
-    T.when hasDhall $ T.die
+    hasPackagesDhall <- T.testfile packagesDhallPath
+    T.when hasPackagesDhall $ T.die
        $ "Found " <> unsafePathToText packagesDhallPath <> ": there's already a project here. "
-      <> "Run `spacchetti local-setup --force` if you're sure you want to overwrite it."
+      <> "Run `spacchetti [command] --force` if you're sure you want to overwrite it."
```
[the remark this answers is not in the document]
justinwoo: Maybe worthwhile

app/Main.hs
justinwoo: Mmmmmmmmmm, maybe that's a good place to put that

app/Main.hs
justinwoo: yeah, might as well hide it in the end

app/Main.hs
justinwoo: This was a mistake with iedit

app/Main.hs
justinwoo: Same comment as above

app/Main.hs
f-f: I think we don't strictly need `dhall-json`, but anyways build-times is only a CI problem?

f-f:
This stuff is still simple so it's possible to get it right without many guardrails, but I'd say we'd want tests to check that the commands do the right things.  
However, since 95% of the stuff happens in `IO` (and we cannot do much about it without adding 300 more lines of code), it's probably best to have end-to-end tests that check that files are in the right places and contain the right stuff and so on.  
@justinwoo I guess a bunch of perl scripts might be the right tool for this?

app/Main.hs
f-f: So should this be `PscPackageClean` or should it clean both `.psc-package` and `.spacchetti`?

f-f: Actually we could leave these commands as they are in this PR and tackle the issue later

app/Main.hs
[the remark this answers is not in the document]
justinwoo: in `build` there's applications of `surroundQuote`, but we also need these quoted forms in reality for `sources` output

justinwoo:
> @justinwoo I guess a bunch of perl scripts might be the right tool for this?

:100: personally i think i might start putting a bunch of executable perl scripts in a repo and pull them in via a simple derivation

justinwoo:
wip branch to be deleted later

PR #20, opened 2018-10-29
Switch to prebuilt docker image for Linux build on Travis 

f-f:
The Linux Travis build now runs all stack commands in a Docker image we built for this purpose, [haskell-lavello](https://github.com/f-f/docker-haskell-lavello), which contains all the prebuilt dependencies we need for the project.

This means that adding new dependencies to the build is now free, at the small upfront cost of pulling the image from Docker Cloud (~7mins).

Note that the osx build still uses native stack commands, but build times there seem to be manageable so far.


justinwoo:
timed out?

justinwoo:
passed, we could merge this

PR #34, opened 2018-12-09
Rename to `spago`

### LINKED ISSUE #23: New name proposal
As we discussed some time ago with @justinwoo, the name of this repo/program has a bit of a name clash with the [spacchetti package set repo](https://github.com/justinwoo/spacchetti), and this is super confusing.

So I'd like to propose a new name for this repo and tool: **spago**
(IPA: */ˈspaɡo/*)

Why?
- It means "twine" in Italian, and it would symbolize the new role of this tool (well at least after #21), to "tie everything together"
- It is different than `spacchetti`
- It kind of reminds of it at the same time (they have the same root)
- It is actually shorter to type at the command line, but unique enough

Sample image to visualize it:
spago

### LINKED ISSUE #26: Port commands from `purp`
After #21 the feature set of this overlaps quite a lot with [`purp`](https://github.com/justinwoo/purp), so we might as well just port that in (since it's 100 lines)

### LINKED ISSUE #32: Fetch git repos in parallel
We should parallelize the fetching of git repos on `install`, as `psc-package` already does

f-f:
Fix #23

This is mostly shuffling around stuff, things happening:
- rename to `spago`
- rewrite the README with a bunch more instructions on various commands
- add a `stylish-haskell` config
- split the responsibilities of working with a `psc-package` project and a `spago` project
- prefix the `psc-package`-related commands with `psc-package-` as mentioned in https://github.com/spacchetti/spacchetti-cli/issues/31#issuecomment-444475865
- fix #32 by fetching the repos concurrently
- fix #26 by implementing the commands from `purp`

### DIFF
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -1,15 +1,29 @@
-module Spago where
+module Spago
+  ( initProject
+  , install
+  , sources
+  , build
+  , test
+  , bundle
+  , makeModule
+  , printVersion
+  , ModuleName(..)
+  , TargetPath(..)
+  , WithMain(..)
+  ) where
 
 import           Control.Concurrent.Async (forConcurrently_)
-import           Control.Exception        (throwIO, try)
+import           Control.Exception        (SomeException, throwIO, try)
 import qualified Data.List                as List
 import qualified Data.Map                 as Map
+import           Data.Maybe               (fromMaybe)
 import           Data.Text                (Text)
 import qualified Data.Text                as Text
 import           Data.Version             (showVersion)
 import qualified Dhall.Format             as Dhall.Format
 import qualified Dhall.Pretty             as Dhall.Pretty
 import qualified Paths_spago              as Pcli
+import           System.IO                (hPutStrLn)
 import qualified System.Process           as Process
 import qualified Turtle                   as T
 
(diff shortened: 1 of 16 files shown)


app/Spago.hs
```diff
@@ -139,7 +149,7 @@ install = do
   config <- ensureConfig
   let deps = getAllDependencies config
   echo' $ "Installing " <> Text.pack (show $ List.length deps) <> " dependencies."
-  _ <- traverse getDep deps
+  _ <- forConcurrently_ deps getDep
```
justinwoo: I think this is good, but Travis toasters can't handle this. Not very much worth looking into at the momen tthough.

f-f: Yeah, and if need be this is quite easy to limit, e.g. see [how `psc-package` does this](https://github.com/purescript/psc-package/blob/648da70ae9b7ed48216ed03f930c1a6e8e902c0e/app/Main.hs#L269)

justinwoo: Seems like something where http://hackage.haskell.org/package/async-pool-0.9.0.2/docs/Control-Concurrent-Async-Pool.html should be solving the problem

PR #37, opened 2018-12-10
local psc-package test

justinwoo:
this is all i need


justinwoo:
Yeah, we could add a script for it to be run sometime

Issue #47, opened 2018-12-18
Rudimentary windows support


btrepp:
The issue you were already expecting :). I'll open this and document trying to get it running on windows. Maintainers: don't feel compelled to respond yet, I'll try to do as much as I can myself.

Running in cmd.exe, can't even get to running the install commands. Looks like the install script assumes bash, curl/wget and tar to be present.

UPDATE: Looks like npm install would never work, that script downloads binaries that have been prebuilt. Next steps for me getting this to work would be getting stack+haskell and trying to build it myself.

UPDATE2: stack install fails due to win32/turtle version mismatchs. Stack is implying to add ` Win32-2.5.4.1@sha256:e623a1058bd8134ec14d62759f76cac52eee3576711cb2c4981f398f1ec44b85` to extra deps. Will see what we get from that. (looks like https://github.com/Gabriel439/Haskell-Turtle-Library/issues/323 is the upstream issue)

UPDATE3: This built with the above change. Though stack install did have to be ran multiple times, but I do have a spago.exe :)

Trying yarn
> yarn global add purescript-spago
> yarn global v1.9.4
> [1/4] Resolving packages...
> [2/4] Fetching packages...
> [------------------------------------------------------------------------------------------------------------------------------------------------] 0/501(node:11552) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
> info fsevents@1.2.4: The platform "win32" is incompatible with this module.
> info "fsevents@1.2.4" is an optional dependency and failed compatibility check. Excluding it from installation.
> [3/4] Linking dependencies...
> [4/4] Building fresh packages...
> error C:\Users\beau.trepp\AppData\Local\Yarn\Data\global\node_modules\purescript-spago: Command failed.
> Exit code: 1
> Command: .\install
> Arguments:
> Directory: C:\Users\beau.trepp\AppData\Local\Yarn\Data\global\node_modules\purescript-spago
> Output:
> '.\install' is not recognized as an internal or external command,
> operable program or batch file.
> info Visit https://yarnpkg.com/en/docs/cli/global for documentation about this command.

Trying npm
> npm install -g purescript-spago
> C:\Users\beau.trepp\AppData\Roaming\npm\spago -> C:\Users\beau.trepp\AppData\Roaming\npm\node_modules\purescript-spago\spago
> 
[...]

PR #48, opened 2018-12-19
Initial windows support changes

btrepp:
Solves getting compilation working, and making bundle work on windows as per #47 

PR'd for code review purposes, I'm not sure if these are the best solutions and won't cause issues for non windows machines.

Unfortunately the encoding issue seems trickier, I'm not sure whether embedding the template or writing it to disk is causing the bad characters to be created, and my research into this has failed (so I might need a haskell guru to help fix this part). I could just change packages.dhall to be ascii, but I feel like this is weak solution as dhall supports unicode, so we should be able to work with that template file.

### DIFF
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -278,7 +278,7 @@ bundle withMain maybeModuleName maybeTargetPath = do
         WithoutMain -> ""
 
       cmd
-        = "purs bundle './output/*/*.js'"
+        = "purs bundle \"output/*/*.js\""
         <> " -m " <> moduleName
         <> main
         <> " -o " <> targetPath
--- a/stack.yaml
+++ b/stack.yaml
@@ -9,3 +9,4 @@ extra-deps:
 - repline-0.2.0.0
 - async-pool-0.9.0.2
 - serialise-0.2.1.0
+- Win32-2.5.4.1@sha256:e623a1058bd8134ec14d62759f76cac52eee3576711cb2c4981f398f1ec44b85


app/Spago.hs
```diff
@@ -278,7 +278,7 @@ bundle withMain maybeModuleName maybeTargetPath = do
         WithoutMain -> ""
 
       cmd
-        = "purs bundle './output/*/*.js'"
+        = "purs bundle output/*/*.js"
```
f-f: Does it work if we keep the single quotes here? They are here to prevent the shell from expanding them (though I'm not sure it would happen for how we run commands)

stack.yaml
```diff
@@ -9,3 +9,4 @@ extra-deps:
 - repline-0.2.0.0
 - async-pool-0.9.0.2
 - serialise-0.2.1.0
+- Win32-2.5.4.1@sha256:e623a1058bd8134ec14d62759f76cac52eee3576711cb2c4981f398f1ec44b85
```
f-f: It looks like version [`2.8.2.0` of Win32](http://hackage.haskell.org/package/Win32-2.8.2.0) is out, so we might want to try to build with it (since this 2.5 version is 6 months old)

There's a bug open upstream about Win32 support in Turtle ([here](https://github.com/Gabriel439/Haskell-Turtle-Library/issues/323)). It's not affecting us, but it probably means that some versions of Win32 don't build with some versions of Turtle.

app/Spago.hs
btrepp: I'd say that's the difference, Having the shell expand them gets bundle to work on windows. I think I can also pass in quotes, but using backslashes. 

This was my punt at well, shell globbing hopefully working for both?. Alternatively it would have to be some OS detection, or maybe there is a haskell path library that can handle this.

stack.yaml
btrepp: Settings this to `- Win32-2.8.2.0` gives

> Error: While constructing the build plan, the following exceptions were encountered:
> 
> In the dependencies for Cabal-2.2.0.1:
>     Win32-2.8.2.0 from stack configuration does not match >=2.2.2 && <2.7  (latest matching version
>                   is 2.6.2.0)
> needed due to spago-0.6.0.0 -> Cabal-2.2.0.1
> 
> In the dependencies for directory-1.3.1.5:
>     Win32-2.8.2.0 from stack configuration does not match >=2.2.2 && <2.7  (latest matching version
>                   is 2.6.2.0)
> needed due to spago-0.6.0.0 -> directory-1.3.1.5
> 
> In the dependencies for process-1.6.3.0:
>     Win32-2.8.2.0 from stack configuration does not match >=2.2 && <2.7  (latest matching version
>                   is 2.6.2.0)
> needed due to spago-0.6.0.0 -> process-1.6.3.0
> 
> In the dependencies for turtle-1.5.12:
>     Win32-2.8.2.0 from stack configuration does not match >=2.2.0.1 && <2.6  (latest matching
>                   version is 2.5.4.1)
> needed due to spago-0.6.0.0 -> turtle-1.5.12
> 
> Some different approaches to resolving this:
> 
>   * Set 'allow-newer: true' in C:\sr\config.yaml to ignore all version constraints and build anyway.
> 
>   * Consider trying 'stack solver', which uses the cabal-install solver to attempt to find some
>     working build configuration. This can be convenient when dealing with many complicated
>     constraint errors, but results may be unpredictable.
> 
>   * Recommended action: try adding the following to your extra-deps
>     in C:\VisualStudio_Projects\purescript\spago\stack.yaml:
> 
> Win32-2.6.2.0@sha256:515518df0728294c99822c65b04be10a242933f40240b125c23469f45c8b11da

btrepp:
So on the encoding issue, commenting out that line causes spago init to successfully complete (I think before the error I get was the format line you mentioned crashing). On disk (opened with vim) the packages.dhall is still corrupted.

My other thoughts might be how the file is being embedded in Templates.hs (this appears to be literate haskell? well outside of my realm now :)).

app/Spago.hs
f-f: My point here is that `purs` supports globbing, and it's probably much faster than shell globbing (it's probably worth benchmarking the two cases tho)

We could use [`</>`](http://hackage.haskell.org/package/filepath-1.4.2.1/docs/System-FilePath-Windows.html#v:-60--47--62-) to form the path, since it supports Windows

f-f:
@btrepp oh right: we use `file-embed` to read the files (and yes, that uses Template Haskell); it looks like there's an issue open already [about Unicode encoding](https://github.com/snoyberg/file-embed/issues/27)

stack.yaml
f-f: Makes sense. How about trying `Win32-2.6.2.0`?

btrepp: > Error: While constructing the build plan, the following exceptions were encountered:
> 
> In the dependencies for turtle-1.5.12:
>     Win32-2.6.2.0 from stack configuration does not match >=2.2.0.1 && <2.6  (latest matching
>                   version is 2.5.4.1)
> needed due to spago-0.6.0.0 -> turtle-1.5.12
> 
> Some different approaches to resolving this:
> 
>   * Set 'allow-newer: true' in C:\sr\config.yaml to ignore all version constraints and build anyway.
> 
>   * Consider trying 'stack solver', which uses the cabal-install solver to attempt to find some
>     working build configuration. This can be convenient when dealing with many complicated
>     constraint errors, but results may be unpredictable.
> 
>   * Recommended action: try adding the following to your extra-deps
>     in C:\VisualStudio_Projects\purescript\spago\stack.yaml:
> 
> Win32-2.5.4.1@sha256:e623a1058bd8134ec14d62759f76cac52eee3576711cb2c4981f398f1ec44b85
> 
> Plan construction failed.

app/Spago.hs
btrepp: @kRITZCREEK suggestion works. Have rebased this commit to match.

templates/packages.dhall
```diff
@@ -118,4 +118,4 @@ let overrides = {=}
 
 let additions = {=}
 
-in  upstream ⫽ overrides ⫽ additions
+in  upstream // overrides // additions
```
btrepp: This is the simplest workaround for the unicode problem, using ASCII in the template here. Users can still use unicode in the file once it's been created, this workaround keeps the UX nice (until a better solution is found).

btrepp: That works perfectly :)

f-f: Awesome! I'll merge that in then, so we can just rebase this one :)

f-f:
Thanks @btrepp! 👏

PR #51, opened 2018-12-20
Fix Unicode file embedding: embedStringFile → embedFileUtf8

f-f:
Might fix some of the Unicode issues in #47 and #48


f-f:
So it looks like this fixes the issues from #48, merging

PR #53, opened 2018-12-21
Fix #50: include src while building tests

### LINKED ISSUE #50: `test` command doesn't work correctly
When I run `test` command, I end up with a bunch of `Module Foo.Bar was not found` errors. It just doesn't see modules in `src` folder. So I guess the answer to [this](https://github.com/spacchetti/spago/blob/master/app/Spago.hs#L253) question is "yes" :)

f-f:
@starper could you try if this fixes the issue?

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -131,7 +131,7 @@ main = do
     Init force                 -> Spago.initProject force
     Install limitJobs          -> Spago.install limitJobs
     Sources                    -> Spago.sources
-    Build                      -> Spago.build Nothing
+    Build                      -> Spago.build
     Test modName               -> Spago.test modName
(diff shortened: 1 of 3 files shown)


justinwoo:
Test project test should be added

f-f:
@justinwoo happens [here](https://github.com/spacchetti/spago/blob/master/test/spago-test.py#L44)

justinwoo:
The tests haven't been updated? They should pull in the Main module to ensure that this behaves as expected.

starper:
@f-f It does. Thanks.

justinwoo:
Good to merge? @f-f

f-f:
@justinwoo I have a commit to fix the tests, I’ll push it and cut a release later today

PR #56, opened 2018-12-23
Don't overwrite files when doing init, but just skip cp if present

f-f:


app/Spago.hs
```diff
@@ -97,16 +98,25 @@ makeConfig force = do
...
   T.mktree "src"
   T.mktree "test"
-  -- TODO fail if these files exist
-  T.writeTextFile "src/Main.purs" Templates.srcMain
-  T.writeTextFile "test/Main.purs" Templates.testMain
-  T.writeTextFile ".gitignore" Templates.gitignore
+  -- But the other files in the template are just skipped if already there.
+  -- Because you might want to just init a project with your own source files,
+  -- or just migrate a psc-package project
+  copyIfNotThere "src/Main.purs" Templates.srcMain
```
justinwoo: IfNotExists

app/Spago.hs
```diff
@@ -97,16 +98,25 @@ makeConfig force = do
...
+  copyIfNotThere "test/Main.purs" Templates.testMain
+  copyIfNotThere ".gitignore" Templates.gitignore
   echo "Set up a local Spago project."
   echo "Try running `spago install`"
+  where
+    copyIfNotThere dest srcTemplate = do
+      let destPath = T.fromText dest
+      (T.testfile destPath) >>= \case
+        True  -> echo ("Found " <> surroundQuote dest <> ", not copying the sample one")
```
justinwoo: this file will not be overridden

app/Spago.hs
f-f: Thank you, I knew there was a more fitting name 😄

app/Spago.hs
f-f: That's the point?

justinwoo: the wording refers to "the sample" which doesn't readily have a relationship to the found file

PR #62, opened 2018-12-31
Add initial repl command implementation

### LINKED ISSUE #46: `spago repl` command
It might be useful to be able to start a REPL from `spago`

vyorkin:
Just a quick working prototype of the `repl` command.

closes #46

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -23,6 +23,9 @@ data Command
   -- | Get source globs of dependencies in spago.dhall
   | Sources
 
+  -- | Start a REPL.
+  | Repl [TargetPath] [T.Text]
+
   -- | Build the project paths src/ and test/
   --   or the specified target paths
   | Build [TargetPath] [T.Text]
@@ -72,6 +75,7 @@ parser
   T.<|> sources
   T.<|> listPackages
   T.<|> build
+  T.<|> repl
   T.<|> test
   T.<|> bundle
   T.<|> makeModule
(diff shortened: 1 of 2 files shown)


app/Main.hs
```diff
@@ -140,6 +148,7 @@ main = do
     Sources                    -> Spago.sources
     Build                      -> Spago.build
     Test modName               -> Spago.test modName
+    Repl                       -> Spago.repl []
```
vyorkin: It should be possible to pass options to the `purs repl` command (like `spago repl --port 3200`)

f-f: Awesome! We could look at how `psc-package` does it, and use something similar to their [`passthroughArgs`](https://github.com/purescript/psc-package/blob/648da70ae9b7ed48216ed03f930c1a6e8e902c0e/app/Main.hs#L651) combinator for gathering the args, and then handle them generically as they do in [`exec`](https://github.com/purescript/psc-package/blob/648da70ae9b7ed48216ed03f930c1a6e8e902c0e/app/Main.hs#L400)

app/Spago.hs
```diff
@@ -253,6 +255,14 @@ newtype TargetPath = TargetPath { unTargetPath :: T.Text }
 
 data WithMain = WithMain | WithoutMain
 
+repl :: [Text] -> IO ()
+repl opts = do
+  config <- ensureConfig
+  let
+    deps  = getAllDependencies config
+    globs = getGlobs deps <> ["src/**/*.purs"]
```
bouzuya: `    globs = getGlobs deps <> ["src/**/*.purs", "test/**/*.purs"]`

https://github.com/spacchetti/spago/blob/62567e626971e6c712213ce10a346ead63f3fbf7/app/Spago.hs#L242

app/Main.hs
f-f: @vyorkin I went on and implemented the passthrough in #66, so once that's merged you can rebase on master and reuse the `passthroughArgs` option to pass args to `purs`, so e.g. this would work:
```bash
spago repl -- --port 3200
```

vyorkin: thanks! sorry for delay, I hope I'll get to it today evening

f-f:
@vyorkin I'd like to merge this ASAP in order to avoid further merge conflicts, so if it's fine by you I'll just merge what we have here

PR #66, opened 2019-01-06
Add option to pass arguments through to purs

### LINKED ISSUE #49: Add a way to specify an output folder
With `psc-package` you can do something like this:
```
psc-package build -- -o path/to/folder
```
and with `pulp` you can use `--build-path` or `-o` flags. It would be nice if `spago` had something like this.

f-f:
With this we pass args through to `purs`, and this is now possible: `spago build -- -o myOutput`

Fix #49


justinwoo:
I almost wish this were another flag, but the precedent is already set with many tools.

PR #79, opened 2019-01-19
Fixes spago test in windows

btrepp:
So here's a weird issue on windows.
`spago test` never ran any tests, and reported tests succeeded.

Seems to be due to how the command is quoted, swapping them around works, but I'm not certain if this
may cause other clients to break.

### DIFF
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -291,7 +291,7 @@ test maybeModuleName paths passthroughArgs = do
     T.ExitFailure n -> die $ "Tests failed: " <> T.repr n
   where
     moduleName = fromMaybe (ModuleName "Test.Main") maybeModuleName
-    cmd = "node -e 'require(\"./output/" <> unModuleName moduleName <> "\").main()'"
+    cmd = "node -e \"require('./output/" <> unModuleName moduleName <> "').main()\""
 
 
 prepareBundleDefaults :: Maybe ModuleName -> Maybe TargetPath -> (ModuleName, TargetPath)


f-f:
@btrepp looks good, I'll test locally to see if there's any regressions. But we could also add a test that tries to match on the test output (we do match the output of some tests, e.g. see [the `fixtures` folder](https://github.com/spacchetti/spago/tree/master/test/fixtures), so adding one should be easy) so we're sure that the tests run

The CI fails here because I opened #81 that adds the AppVeyor build, will be fixed as soon as we merge and rebase

f-f:
@btrepp I was wondering if [tests](https://github.com/spacchetti/spago/blob/master/test/spago-test.py) would run at all on Windows? (they require `python3` and `purs`, see [here](https://github.com/spacchetti/spago/blob/master/.travis.yml#L88) for how to run them)

btrepp:
@f-f For CI Testing? Running locally gives.

>  python ..\spago-test.py
> Expecting success from: "spago init"
> Expecting failure from: "spago init"
> Expecting success from: "rm spago.dhall packages.dhall"
> Expecting success from: "spago init"
> Expecting success from: "rm src/Main.purs"
> Expecting success from: "spago init -f"
> FAILURE: Spago should always succeed in doing init with force
> Program output:
> spago: .\packages.dhall: openFile: permission denied (Permission denied)

So I think somethings gone awry (probably with file paths). I'll open a new issue to track it.

Issue #90, opened 2019-01-28
Hashes in existing projects break with new Dhall version


f-f:
I setup a project with the currently released Spago version:

```bash
$ spago version
0.6.3.0

$ spago init
Set up a local Spago project.
Try running `spago install`
```

Then I upgrade Spago to the latest master, and if I try to install, it fails:
```bash
$ spago version
0.6.4.0

$ spago install
spago:
↳ ./packages.dhall
  ↳ https://raw.githubusercontent.com/spacchetti/spacchetti/20181209/src/mkPackage.dhall sha256:8e1c6636f8a089f972b21cde0cef4b33fa36a2e503ad4c77928aabf92d2d4ec9
This decoded version is not supported: 4.0.0
```

What happened here?  
In [this commit](https://github.com/spacchetti/spago/commit/48ec1e86eacebf10aa597aedff06dce653dfa7cc) we upgraded to the latest Dhall version (1.19 → 1.20), which also upgraded the Dhall language version (4.0.0 → 5.0.0), rendering the hashes in the project file non-compatible with the new version, which cannot decode them.

There is already an issue open about a related problem: https://github.com/dhall-lang/dhall-lang/issues/335

Once the fix will be implemented in Dhall it will be easy to "upgrade the clients" if the hash of the expression is the same (but the version unsupported), I'll investigate if in the meanwhile we can patch it nicely and "autoupgrade" the projects

f-f:
@Gabriel439 would you have any suggestion on how to go about fixing this? The goal here is to migrate people's configs as automatically as possible.

 I considered the following flow:
- if there's a hash and a cache hit, try to `decodeWithVersion` it. If it's an old version substitute the new current version of the standard in the Haskell-CBOR `Term` and try to `decode` anyways.
- If this last decoding is successful, refreeze the import with the current standard version

Would this makes sense?

I took a stab at a modified version of `Dhall.Binary.decodeWithVersion`, but `Dhall.Binary.decode` is not exported (and I'm not sure it'd be a good idea to export it?), so I'm unsure how to proceed

Gabriel439:
@f-f: I think I just need to hurry up and standardize https://github.com/dhall-lang/dhall-lang/issues/335.  That seems like the correct solution to this

PR #93, opened 2019-02-02
Add command to upgrade the package-set if the upstream is Spacchetti

### LINKED ISSUE #39: Automate updating the spacchetti version
We should update automatically the spacchetti version and sha in `packages.dhall` every time spacchetti releases. This should be done in the `release` step of `spacchetti` Travis.

Steps for automating this:
- sed the new release name in the urls in `packages.dhall`, more sed to remove the sha265:..
- run `dhall freeze --inplace templates/packages.dhall`

### LINKED ISSUE #73: `spago spacchetti-upgrade` command
It is common that every once in a while you might want to sync to the upstream package-set.

We could have a `spago spacchetti-upgrade` command, that would:
- check that we are using a package set from a spacchetti release
- check if we're on the latest release
- if we are not:
  - get the latest release
  - fix the import URL
  - and recompute the Dhall hash for the new import

Related to #39

f-f:
This PR adds the command `spago spacchetti-upgrade`, to upgrade your package set to the latest release if the upstream is Spacchetti (fix #73).

Running it would look something like this:
```bash
$ spago spacchetti-upgrade
Found the most recent tag for "spacchetti": 20190131
Trying to read "packages.dhall"
Package-set upgraded to latest tag "20190131"
Fetching the new one and generating hashes.. (this might take some time)
Done. Updating the local package-set file..
```

This also fixes #39, as CI will now fail if our `packages.dhall` template is not on the latest version of spacchetti. Excerpt from Travis log in case of failure:

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ services:
 
 matrix:
   include:
-  # BBuild on Linux with docker
+  # Build on Linux with docker
   - env: BUILD=stack ARGS=""
     compiler: ": #stack default"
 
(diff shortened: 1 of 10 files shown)


f-f:
Implementation is correct, but fails because of #90 (and the fix for that will land in dhall master in some days)

PR #95, opened 2019-02-02
Start running tests on AppVeyor

### LINKED ISSUE #91: Spago testing fails on windows
Running the python tests seems to fail on windows.

I'll spend some time later investigating it further.

> python ..\spago-test.py
> Expecting success from: "spago init"
> Expecting failure from: "spago init"
> Expecting success from: "rm spago.dhall packages.dhall"
> Expecting success from: "spago init"
> Expecting success from: "rm src/Main.purs"
> Expecting success from: "spago init -f"
> FAILURE: Spago should always succeed in doing init with force
> Program output:
> spago: .\packages.dhall: openFile: permission denied (Permission denied)

f-f:
Fix #91

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,9 +7,10 @@ branches:
 environment:
   STACK_ROOT: C:\sr
   STACK_VERSION: 1.9.3
-  PYTHON: "C:\\Python35-x64"
-  PYTHON_VERSION: "3.5.0"
   PYTHON_ARCH: "64"
+  PYTHON_VERSION: 3.6
+  PYTHON: C:\\Python36-x64
+  TEST_WHEELINSTALL_ENV: wheelinstall-py36
   
   # Workaround a gnarly bug https://github.com/haskell/cabal/issues/5386
   # See: https://www.fpcomplete.com/blog/2018/06/sed-a-debugging-story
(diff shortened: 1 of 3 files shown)


btrepp:
So It seems the tests fail the first time they are ran, but succeed the second time. The cause is the run_for command, upping the time from 0.5s to 10 makes it always succeed (which I'm not sure why, maybe the python libraries don't send signals if the sub-process has gone away).

Further looking into it, CTRL+C looks like it gets sent to everything, so python seems to also get a ctrl+c event.

f-f:
@btrepp yeah, apparently you can use that CTRL_C signal only with `os.kill`. There's an [excellent Stack Overflow answer about this](https://stackoverflow.com/questions/7085604/sending-c-to-python-subprocess-objects-on-windows), where there is at least one solution that works, so we might want to try that.

Feel free to push to this branch if you want to give it a shot :)

More context on this test: here we are testing that Spago cleans after itself when downloading stuff.
When installing it will, in parallel, create folders for the deps and run some git commands. If something goes wrong before the git commands are completed, it should remove the folder it created (because we assume that if the folder is there then the package is properly installed).
Now, we are simulating the "something goes wrong" condition with a ^C on Linux (this is an actual use case, as you might wish to stop the install while it runs), and we have to do it in such a short time that Spago has the time to create the directories but not to complete the git commands, and 0.5s is a good value for that.

PR #96, opened 2019-02-03
Fix #88: implement handling of local packages in the package-set

### LINKED ISSUE #88: `spago link` command
Coming from using bower I'm missing `bower link`, spago is making it easier to consume others libraries but using bower it's damn handy to be able to link against local versions of packages when developing libraries/verifying changes etc.

Two options I can think of thus far.

1) Dropping a symlink to .spago/package/local/
2) Playing with packages .dhall to link in the other package.

The first is simple, but if your package has other deps it could be a fun little dependency hell.

For the latter, dhall is powerful enough that you can do things like

> let tiled = ../tiled/spago.dhall
> 
> let mkPackage =
>       https://raw.githubusercontent.com/spacchetti/spacchetti/20190105/src/mkPackage.dhall sha256:90974a8e07650e49a4197d4ce5b59e82740fd8469c8c1b9793939845ca8faad9
> 
> let upstream =
>       https://raw.githubusercontent.com/spacchetti/spacchetti/20190105/src/packages.dhall sha256:04da9c924e05cb6b43447ffd13408d01992f3172eb80d5d94e7f1b14c76a5123
> 
> let overrides = {=}
> 
> let additions = 
>     { tiled  = 
>         mkPackage 
>           tiled.dependencies
>           "..\\..\\..\\..\\tiled"
>           "master"
>     }
> 
> let tiledset = tiled.packages    
> 
> in  upstream ⫽ overrides ⫽ additions // tiledset

This works, with the caveat that if you update the repo's master, you need to delete `.spago\tiled\master\` to get spago install to copy everything again.

I can also appreciate that it's better to guide everyone to contribute more to the package sets, but this is more of a situation where having my own forked spachetti set/ overrides seems a tad overkill, especially as I'm still experimenting with what the 'package' should look like, so there's little purposing in publishing any of it.. yet.

f-f:
Implements support for local packages, as described in https://github.com/spacchetti/spago/issues/88#issue-403519960

Basically we now try to distinguish if "URL"s in the package-set are referring to remote repos, or local filesystem path. In the latter case instead of trying to clone them, we just include their sources in the build (if they are listed as a dependency ofc). This basically mirrors the behaviour of `stack`.

Practical example: let's say I find a bug in simple-json. I manage to fix it locally, but I also want to see if the fix works for my project.

Since simple-json is already in the package-set, I can add the local package to the overrides in `packages.dhall`:
```haskell
...

let overrides = { simple-json = upstream.simple-json // { repo = "../purescript-simple-json" } }

...
```

Note that if I `list-packages`, it is correctly listed as a local package:
```
$ spago list-packages
...
signal                v10.1.0   Remote "https://github.com/bodil/purescript-signal.git"
sijidou               v0.1.0    Remote "https://github.com/justinwoo/purescript-sijidou.git"
simple-json           v4.4.0    Local "../purescript-simple-json"
simple-json-generics  v0.1.0    Remote "https://github.com/justinwoo/purescript-simple-json-generics.git"
smolder               v11.0.1   Remote "https://github.com/bodil/purescript-smolder.git"
...
```

After adding it to the `dependencies` in `spago.dhall`, if I do `spago install`, it will not download it, but only its dependencies:
```
$ spago install
Installing 42 dependencies.
...
Installing "distributive"
Installing "refs"
Installing "identity"
Skipping package "simple-json", using local path: "../purescript-simple-json"
Installing "control"
Installing "enums"
Installing "st"
...
```

### DIFF
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -36,7 +36,7 @@ import qualified Turtle                        as T hiding (die, echo)
 import qualified PscPackage
 import           Spago.Config                  (Config (..))
 import qualified Spago.Config                  as Config
-import           Spago.Spacchetti              (Package (..), PackageName (..))
+import           Spago.Spacchetti              (Package (..), PackageName (..), Repo (..))
 import qualified Spago.Templates               as Templates
 import           Spago.Turtle
 
@@ -77,16 +77,27 @@ initProject force = do
 
 
 -- | Returns the dir path for a given package
+--   If the package is from a remote git repo, return the .spago folder in which we cloned
+--   Otherwise return the local folder
 getPackageDir :: (PackageName, Package) -> Text
-getPackageDir (PackageName{..}, Package{..})
+getPackageDir (PackageName{..}, Package{ repo = Remote _, ..})
   = spagoDir <> packageName <> "/" <> version
+getPackageDir (_, Package{ repo = Local path })
+  = path
 
 getGlobs :: [(PackageName, Package)] -> [Text]
 getGlobs = map (\pair -> getPackageDir pair <> "/src/**/*.purs")
 
 
+-- | If the repo points to a remote git, fetch it in the .spago folder.
+--   If it's a local directory do nothing
 getDep :: (PackageName, Package) -> IO ()
-getDep pair@(PackageName{..}, Package{..} ) = do
+getDep (PackageName package, Package { repo = Local path }) =
+  echo $ "Skipping package "
+      <> surroundQuote package
+      <> ", using local path: "
+      <> surroundQuote path
+getDep pair@(PackageName{..}, Package{ repo = Remote repo, ..} ) = do
   exists <- T.testdir $ T.fromText packageDir
   if exists
     then do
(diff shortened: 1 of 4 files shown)


app/Spago/Spacchetti.hs
```diff
@@ -15,7 +17,7 @@ newtype PackageName = PackageName { packageName :: Text }
 -- | A spacchetti package.
 data Package = Package
   { dependencies :: [PackageName] -- ^ list of dependency package names
-  , repo         :: Text          -- ^ the git repository
+  , repo         :: Repo          -- ^ the git repository
```
justinwoo: comment is out of date

app/Spago.hs
```diff
@@ -77,16 +77,27 @@ initProject force = do
 
 
 -- | Returns the dir path for a given package
+--   If the package is from a remote git repo, return the .spago folder in which we cloned
```
justinwoo: docstring?

app/Spago/Spacchetti.hs
```diff
@@ -25,3 +27,17 @@ instance FromJSON Package
 
 type Packages = Map PackageName Package
 
+data Repo
```
justinwoo: should have some basic comments

app/Spago.hs
f-f: Haskell docstrings don't have the `|` in multiple lines, [see here](https://www.haskell.org/haddock/doc/html/markup.html)

f-f:
@justinwoo note that in his example is an addition, not an override, so you need to specify the dependencies somehow: either listing them or pulling them from a spago config if the package you're importing has one

justinwoo:
So in this case you'd want to have a new mkPackage function that can make local library package definitions? Where one of the args passed in would be either the path or the imported record?

f-f:
@justinwoo could you make an example of how it would look like?

I though about having a `mkLocalPackage` that takes only two args (name and deps), but in the end it's not worth it, because a usecase could be that once you upload your package, you just swap the local package with the URL of the new git repo. In the end I'm not even sure we need `mkPackage`, since it's just producing a record.

justinwoo:
mkLocalPackage would be useful enough. mkPackage is also fine overall since you know that's what you're making rather than just some random record that you hope is fine, and also since you want to ensure that records are of type Package.

PR #99, opened 2019-02-06
Add explanation about npm packages

f-f:
Explain why my proposal here doesn't work:
https://discourse.purescript.org/t/post-purescript-package-management-explained/615/7

This thread is also highly informative on the matter:
https://github.com/purescript/purescript/issues/631#issuecomment-255542135

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,14 +25,13 @@ PureScript package manager and build tool powered by [Dhall][dhall] and
     - [Adding and overriding dependencies](#adding-and-overriding-dependencies)
     - [Upgrading the Package Set](#upgrading-the-package-set)
   - [Building, bundling and testing a project](#building-bundling-and-testing-a-project)
-- [Can I use this with `psc-package`?](#can-i-use-this-with-psc-package)
-  - [`psc-package-local-setup`](#psc-package-local-setup)
-  - [`psc-package-insdhall`](#psc-package-insdhall)
 - [FAQ](#faq)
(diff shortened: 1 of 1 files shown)


chexxor:
Nice explanation. I believe it's likely that this won't prevent/end future requests for spago to integrate with other build/dep management tools, but this at least gives a common starting point for those discussions. I think it's possible there's still some ways to support these other tools in a limited scope, so perhaps those future discussions will have proposals for them.

If you'd like to more absolutely block future requests for support for other dep tools, I suggest putting a more pointed sentence in there to state so.

README.md
```diff
@@ -486,6 +456,55 @@ the `ksf-login` component and output it in the `index.js` of the component's fol
...
+#### Why can't `spago` also install my npm dependencies?
+
+A common scenario is that you'd like to use things like `react-basic`, or want to depend
+on JS libraries like ThreeJS. 
+In any case, you end up depending on some NPM package.
+
+And it would be really nice if `spago` would take care of installing all of these
+dependencies, so we don't have to worry about running npm besides it, right?
+
+While this scenarios are common, they are also really hard to support.
```
JordanMartinez: Small typo: `these` not `this`

f-f:
@chexxor: integrating with other tools is hard and it's a lot of work, but it's an actual UX issue.
I would love myself to not have to use `npm` and `yarn` and `lerna` and whatever else. Setting up the JS+PS build for [our monorepo at work](https://github.com/KSF-Media/affresco) was very painful.

So I wouldn't like to exclude this possibility, as someone might come up with a clever solution and PR it 😄

PR #107, opened 2019-02-10
Check that purs version is at least what the package-set specifies

### LINKED ISSUE #101: Check that purs version is at least what the package-set specifies
As detailed in https://github.com/spacchetti/spago/issues/98#issuecomment-461131045, we should run `purs --version` every time we want to `install` packages, to check that we're running at least the compiler version specified by the package-set (see https://github.com/spacchetti/spacchetti/issues/63)

If the version isn't there we should skip the check (see https://github.com/spacchetti/spago/issues/94 for when we could fail instead), if the version is there we should fail if it's not SemVer-compatible (e.g. `0.11.x` is not compatible with `0.12.x`, but `0.12.x` is compatible with `0.12.y`, if `x > y` and the compiler has `x`), so that the user can install the right version of the compiler.

f-f:
Fix #101

Spago will now check the currently installed `purs` version when doing `install`

If you're on an old package-set that doesn't support this check you'll get a warning:
```
WARNING: the package-set version you're on doesn't check if the version of the
PureScript compiler installed on your system is compatible.
If your build fails you might want to upgrade your set by running this command:
`spago spacchetti-upgrade`
```

If you don't have the minimum version installed you'll get an error:
```
spago: Oh noes! It looks like the PureScript version installed on your system is

installed version:   0.12.1
package-set version: 0.12.2

Please upgrade your `purs` version. You can do it by running:
`npm update purescript`
```


justinwoo:
I hope this warning message only happens once or something, and instead of "ASAP" it's probably more like "maybe you should", since if it builds, it works.

f-f:
> since if it builds, it works.

@justinwoo good point. Maybe it's ok to remove the warning altogether?

justinwoo:
Warn once is fine, if you find a way to do it. Warning again on failed builds if this condition is true is also worth doing once i think

f-f:
"Warning once" is hard, so I'll warn every time but make the warning less scary and smaller 😄

PR #108, opened 2019-02-10
Add `spago verify` command

### LINKED ISSUE #14: `spago verify` command
Need a verify command that will attempt to install and build a given dependency. This can be done in almost any way, but a full set verify should install one package while verifying at a time.

f-f:
@justinwoo could you check that the implementation is somehow correct? It's basically extracted from psc-package, but I'm a bit confused by how it's performed there

Fix #14


app/Spago.hs
```diff
@@ -250,21 +281,55 @@ listPackages = do
...
+          verifyPackages packages toVerify
+  where
+    verifyPackages :: PackageSet -> [(PackageName, Package)] -> IO ()
+    verifyPackages packageSet packages = do
+      echoStr $ "Verifying "
+              <> show (length packages)
+              <> " packages, this might take a while.."
+      traverse_ (verifyPackage packageSet) (fst <$> packages)
+
+    verifyPackage :: PackageSet -> PackageName -> IO ()
```
justinwoo: i don't quite understand why this should be nested in here instead of being top level? it's actually bothered me that we have `verify` to verify an entire set instead of a given specific package

app/Spago.hs
```diff
@@ -250,21 +281,55 @@ listPackages = do
...
-    deps = getAllDependencies config
-    globs = getGlobs deps
-  _ <- traverse echo globs
+  deps <- getProjectDeps config
+  _ <- traverse echo $ getGlobs deps
   pure ()
 
 
+verify :: Maybe Int -> Maybe PackageName -> IO ()
+verify maybeLimit maybePackage = do
```
justinwoo: below comment would remove the need for Maybe whatever

app/Spago.hs
```diff
@@ -160,41 +164,12 @@ getDep pair@(PackageName{..}, Package{ repo = Remote repo, ..} ) = do
...
-
-  config <- Config.ensureConfig
 
-  let deps = getAllDependencies config
+fetchPackages :: Maybe Int -> [(PackageName, Package)] -> IO ()
+fetchPackages maybeLimit deps = do
   echoStr $ "Installing " <> show (List.length deps) <> " dependencies."
   Async.withTaskGroup limit $ \taskGroup -> do
-    asyncs <- for deps $ \dep -> Async.async taskGroup $ getDep dep
+    asyncs <- for deps $ \dep -> Async.async taskGroup $ fetchPackage dep
```
justinwoo: no change needed:

oh, i missed this but my guess is that a taskgroup only approximates what you truly want here, that only a batch gets worked at a time. doesn't really matter though, because forking a pool of workers to read a shared resource mvar is a pain in the ass (see various implementations of waitPool on hackage)

app/Main.hs
```diff
@@ -169,6 +178,7 @@ main = do
     ListPackages                          -> Spago.listPackages
     Sources                               -> Spago.sources
     Build limitJobs paths pursArgs        -> Spago.build limitJobs paths pursArgs
+    Verify limitJobs maybePackage         -> Spago.verify limitJobs maybePackage
```
justinwoo: commented below, but i would really rather do away with the optional arguments and either add `verify-pkg packagename` or `verify-set`

f-f: Current interface is:
```bash
# whole set
spago verify
# single package
spago verify simple-json
```

With `verify-set` it would be:
```bash
# whole set
spago verify-set
# single package
spago verify simple-json
```

I'm ok with having the `Maybe` there and skip adding a new command, as "less command surface" is better (less confusing for the user)

app/Spago.hs
f-f: Are the use-cases for `verify $package` and `verify-set` similar?

justinwoo: almost not at all, because you will verify-set locally but only run verify in CI

f-f: I.e. `verify-set` is to verify the consistency of the whole package-set right?

justinwoo: Funny thing is that it might not, since some packages might be orphans with no other shared references. But yes, overall it's meant to do that, whereas verifying a package does the reverse deps/referrers check because you want to make sure the it doesn't break them without having to check the whole set.

f-f: Thanks for the explanation 👍 Let's go with `verify-set` then 🙌

f-f:
@kedashoe I added a function to get the "transitive dependencies" here, so I thought that it would be useful to add in `list-packages`.

I renamed the `deps` flag to `filter`, and it works like this:
```bash
# Direct dependencies, i.e. only the ones listed in spago.dhall
$ spago list-packages --filter=direct

# Transitive dependencies, i.e. all the dependencies of your dependencies
$ spago list-packages -f transitive
```

app/Spago.hs
f-f: I left the `Maybe` in, so we can contain all the functionality (i.e. for both `verify` and `verify-set`) in this `verify` block

PR #144, opened 2019-03-16
Don't create a configuration automatically if not found

### LINKED ISSUE #139: Don't create spago.dhall on failed spago build
I often mistakenly run `spago build` from some subdirectory. This, correctly, fails with a message about missing `packages.dhall`. However, it also creates a new `spago.dhall` file in the current directory.
It would be nice if it didn't.

f-f:
Example of running spago in a directory where there's no configs:

```bash
$ spago build
spago: There's no "spago.dhall"in your current location.

If you already have a spago project you might be in the wrong subdirectory,
otherwise you might want to run `spago init` to initialize a new project.
```
Fix #139

/cc @JamieBallingall

### DIFF
--- a/test/spago-test.py
+++ b/test/spago-test.py
@@ -78,12 +78,12 @@ run_for(0.5, ['spago', 'install', '-j', '3'])
 time.sleep(1)
 
 expect_success(
-    ['spago', 'install'],
+    ['spago', 'install', '-j', '10'],
     "Subsequent installs should succeed anyways"
 )
 
 expect_success(
-    ['spago', 'install', 'simple-json', 'foreign'],
+    ['spago', 'install', '-j', '10', 'simple-json', 'foreign'],
     "Spago should be able to add dependencies"
 )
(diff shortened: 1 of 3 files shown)


test/spago-test.py
```diff
@@ -77,22 +78,28 @@
 time.sleep(1)
 
 expect_success(
-    ['spago', 'install', '-j', '10'],
+    ['spago', 'install'],
```
f-f:
```suggestion
    ['spago', 'install', '-j', '10'],
```

test/spago-test.py
```diff
@@ -77,22 +78,28 @@
...
     "Subsequent installs should succeed anyways"
 )
 
 expect_success(
-    ['spago', 'install', '-j', '10', 'simple-json', 'foreign'],
+    ['spago', 'install', 'simple-json', 'foreign'],
```
f-f:
```suggestion
    ['spago', 'install', '-j', '10', 'simple-json', 'foreign'],
```

PR #149, opened 2019-03-16
Run build upon bundle and make-module

### LINKED ISSUE #146: Automatically build in commands that require the project to be built
It looks like it's common to forget to call `spago build` before calling `spago bundle` or `spago make-module`.
From the UX point of view this doesn't make sense, because if these commands require the project to be built, then spago should build it if it's not built yet.

Implementation note: this basically requires plugging `Spago.Build.build` into `bundle`, `make-module` and `test`. This means that these commands should also accept the options that the `build` command accepts

FintanH:
Fixes #146 

**What's in this PR**
* Adds call to `build` in `bundle`
* Since `make-module` calls `bundle` it will build automatically too
* As a consequence I added args for passing the paths and and purs args to the commands for `bundle` and `make-module`
* Update to 0.7.1.0
* Update README to show the path and pass-through being used

PR checklist:
- [x] Updated the version number [in `package.yaml`](https://github.com/spacchetti/spago/blob/master/package.yaml#L2) according to SemVer (`0.x.y.z`, `x` changes on breaking changes, `y` on additions, `z` on patches)

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,12 @@ PureScript package manager and build tool powered by [Dhall][dhall] and
     - [Verifying your additions and overrides](#verifying-your-additions-and-overrides)
     - [Upgrading the Package Set](#upgrading-the-package-set)
     - [Caching the Package Set](#caching-the-package-set)
-  - [Building, bundling and testing a project](#building-bundling-and-testing-a-project)
+  - [Building and testing a project](#building-and-testing-a-project)
+  - [Bundling a project into a single JS file](#bundling-a-project-into-a-single-js-file)
+    - [1. `spago bundle`](#1-spago-bundle)
+    - [2. `spago make-module`](#2-spago-make-module)
+    - [3. `spago build` + whatever JS bundler](#3-spago-build--whatever-js-bundler)
+  - [Documentation](#documentation)
 - [FAQ](#faq)
     - [Hey wait we have a perfectly functional `pulp` right?](#hey-wait-we-have-a-perfectly-functional-pulp-right)
     - [I miss `bower link`!](#i-miss-bower-link)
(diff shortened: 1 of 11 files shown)


FintanH:
@f-f I didn't check if there's a test for this. I'll check now, but lemme know if there's one :)

justinwoo:
This could use a flag for optionally skipping build, because this will add to my build times if I don't reduce redundant build calls.

FintanH:
@justinwoo So I was under the impression that `bundle` and `make-module` _need_ `build` to have run before and `build`ing something a second time is a no-op. So will it actually add much to build times, out of curiosity?

justinwoo:
Unfortunately, since the compiler will go through parsing a bunch of externs files and looking at modification times. I'm not sure if it goes so far as trying to parse file contents, but it is still quite slow (in the range of 1 - 10s)

FintanH:
Cool, I can add a flag so!

justinwoo:
Thxx

README.md
```diff
@@ -49,7 +49,7 @@ PureScript package manager and build tool powered by [Dhall][dhall] and
 
 ## What does all of this mean?
 
-`spago` aims to tie together the UX of developing a PureScript project.  
+`spago` aims to tie together the UX of developing a PureScript project.
```
f-f: It looks like "whitespace removal" is a common feature in editors by now, but these two spaces are Markdown syntax (newline without new paragraph) 😅

Could you undo the whitespace changes to the README?

README.md
```diff
@@ -454,6 +454,14 @@ Bundle succeeded and output file to index.js
 $ node .
 ```
 
+Since `spago bundle` also requires the source project to be built, it will do so before
+bundling. Due to this, we can specify the source paths, and the pass-through arguments to the
+`purs` compiler.
+
+```bash
+$ spago bundle --main Main --to index.js --path 'another_source/**/*.purs' -- -o myOutput/
```
f-f: I'd say we could skip this text: the reason is that if someone needs to pass these flags they will do it already when calling `build`, and they'll want to do it anyways (so they don't need this text); the rest of the people (those not passing the "build" flags) don't care already about them, so they don't need this text (as all of this information is anyways there when you do `spago bundle -h`)

I'd say we could just mention something like "`spago bundle` will build your project if it hasn't been built already"

README.md
```diff
@@ -470,6 +478,14 @@ $ node -e "console.log(require('./index).main)"
 [Function]
 ```
 
+Similar to `spago bundle`, `spago make-module` requires the source project to be built, and
+will do so for us. We can also specify the source paths, and the pass-through arguments to
+the `purs` compiler.
+
+```bash
+$ spago make-module --main Main --to index.js --path 'another_source/**/*.purs' -- -o myOutput/
```
f-f: Same as above

justinwoo:
That would often end up with you having stale builds if it works the first time around by retrying. Less surprising to have a flag to purposefully skip builds.

f-f:
@justinwoo good point 👍 Cool, let's have something like a `--no-build` flag

README.md
FintanH: ◀️

README.md
FintanH: Ok, I removed these and added a section on how to skip the build

FintanH:
@f-f @justinwoo So I wasn't sure about naming conventions but added a flag to skip the build 👌I think this would be a breaking change because `bundle` because it will expect the flag. So I think we could:
1. Make it default to false, i.e. should build
2. Bump major version
3. Option C?

I just need to add the test case, but need to head out for a bit :)

f-f:
@FintanH looks great to me!
It's not a breaking change, since not adding the flag defaults to false (IMO we have a "breaking change" when a behaviour is removed. In this case no user will be broken by this addition)

FintanH:
Oh I wasn't aware of the defaulting behaviour!

FintanH:
@f-f should be good to go 👌

f-f:
@FintanH many thanks! 👏🎉

PR #171, opened 2019-04-07
Better handling of Dhall AST

### LINKED ISSUE #158: Handling of the `packages` config key is surprising
Suppose I have a project that contains several PureScript apps. I'd like to use `spago` to pin a single package set (+ additions and overrides) and have all my apps pull dependencies from that set. I'd like/expect my directory structure to look like

```
.
├── A
│   ├── spago.dhall
│   ├── src
│   │   └── Main.purs
│   └── test
│       └── Main.purs
├── B
│   ├── spago.dhall
│   ├── src
│   │   └── Main.purs
│   └── test
│       └── Main.purs
└── packages.dhall
```

Where `A/spago.dhall` looks something like:

```
{ name =
    "A"
, dependencies =
    [ "effect", "console" ]
, packages =
    ./../packages.dhall
}
```

But when I run `spago build` inside `A` a fresh `packages.dhall` is generated in that directory. 

It turns out that setting the `packages` key to anything other than `./packages.dhall` is kinda redundant, which I understand is for `psc-package` interop reasons. 

In order to achieve what I want (correct me if I'm wrong here) I'd have to have change `A/spago.dhall` to the standard:

```
{ name =
    "A"
, dependencies =
    [ "effect", "console" ]
, packages =
    ./packages.dhall
}
```

Then `A/packages.dhall` should just reference the shared package set

```
./../packages.dhall
```

Which isn't bad, just a bit unexpected. 

Discuss :)

### LINKED ISSUE #162: Parse error when `packages.dhall` doesn't follow the template format
The following `packages.dhall` file will throw a surprising parse error:

```
let overrides = {=}

let additions = {=}

let mkPackage =
      https://raw.githubusercontent.com/purescript/package-sets/psc-0.12.3-20190312/src/mkPackage.dhall sha256:0b197efa1d397ace6eb46b243ff2d73a3da5638d8d0ac8473e8e4a8fc528cf57

let upstream =
      https://raw.githubusercontent.com/purescript/package-sets/psc-0.12.3-20190312/src/packages.dhall sha256:379a4e0b19b1c1977d053d6cdd9403c579014f2cbdeab8b72700a1ae438b6897

in  upstream // overrides // additions
```

If this _must_ be an error, it probably shouldn't be described as a parse error. Cos it is valid Dhall and therefore should parse. Something more along the lines of _"packages.dhall should follow the standard format so that we can implement much of the functionality"_  would be more appropriate?

The relevant code is: https://github.com/spacchetti/spago/blob/master/app/Spago/PackageSet.hs#L117-L139

### LINKED ISSUE #163: Erroneous package-set warning when importing `packages.dhall`
Spago project:

```
{- spago.dhall 
-}
{ name =
    "my-project"
, dependencies =
    [ "effect", "console" ]
, packages =
    ./packages.dhall
}
```

```
{- packages.dhall

Note that `upstream` references another file

Also note that I'd prefer this file to consist of only `./my-packages.dhall`, but the current
parsing logic doesn't allow that...

-}
let mkPackage =
      https://raw.githubusercontent.com/purescript/package-sets/psc-0.12.3-20190312/src/mkPackage.dhall sha256:0b197efa1d397ace6eb46b243ff2d73a3da5638d8d0ac8473e8e4a8fc528cf57

let upstream = ./my-packages.dhall -- IMPORTANT

let overrides = {=}

let additions = {=}

in  upstream ⫽ overrides ⫽ additions
```

```
{- my-packages.dhall 
-}
let mkPackage =
      https://raw.githubusercontent.com/purescript/package-sets/psc-0.12.3-20190312/src/mkPackage.dhall sha256:0b197efa1d397ace6eb46b243ff2d73a3da5638d8d0ac8473e8e4a8fc528cf57

let upstream =
      https://raw.githubusercontent.com/purescript/package-sets/psc-0.12.3-20190312/src/packages.dhall sha256:379a4e0b19b1c1977d053d6cdd9403c579014f2cbdeab8b72700a1ae438b6897

let overrides = {=}

let additions = {=}

in  upstream ⫽ overrides ⫽ additions
```

Running `spago build` throws the following warning:

```
WARNING: the package-set version you're on doesn't check if the version of the
PureScript compiler installed on your system is compatible.
If your build fails you might want to upgrade your set by running this command:
`spago package-set-upgrade`
```

Which I'm pretty sure it shouldn't :)

f-f:
Fix #158 
Fix #162 
Fix #163

This PR reworks how the `config.dhall` and `packages.dhall` files are handled when reading their unevaluated form (i.e. as Dhall AST).
A small summary of what's in it:
- move from pattern matching on the shape of the AST to instead traversing the tree and searching for patterns (using e.g. `Control.Lens.Combinators.transformOf`), and this gives us less error conditions and more freedom for the user to edit the config as they wish
- In fact, the shape of `packages.dhall` is now not constrained at all, while `spago.dhall` still needs to be a record containing the `dependencies` key (we need this for `spago install $dep`)
- In case the packages file is not there Spago will not be happy (e.g. it wants to check that the upstream tag is compatible, and wants to check that the packages file is frozen, etc) but continue execution, emitting small 1-line warnings instead of the current 4-line warning
- The only exception to the above is when doing `spago package-set-upgrade`, in which case Spago will error out if the file is not there

/cc @jmackie could you try this out and see if it behaves as you'd expect?

### DIFF
--- a/app/Spago/Config.hs
+++ b/app/Spago/Config.hs
@@ -103,33 +103,33 @@ makeConfig force = do
       Left err -> echo $ Messages.failedToReadPscFile err
       Right pscConfig -> do
         echo "Found a \"psc-package.json\" file, migrating to a new Spago config.."
-        withConfigAST (pure . updateName (PscPackage.name pscConfig))
         -- try to update the dependencies (will fail if not found in package set)
         let pscPackages = map PackageName $ PscPackage.depends pscConfig
         config <- ensureConfig
-        withConfigAST (addRawDeps config pscPackages)
+        withConfigAST (\e -> addRawDeps config pscPackages
+                            $ updateName (PscPackage.name pscConfig) e)
 
 
-updateName :: Text -> Expr -> Maybe Expr
+updateName :: Text -> Expr -> Expr
 updateName newName (Dhall.RecordLit kvs)
-  | Just _name <- Dhall.Map.lookup "name" kvs = Just $ Dhall.RecordLit
+  | Just _name <- Dhall.Map.lookup "name" kvs = Dhall.RecordLit
     $ Dhall.Map.insert "name" (Dhall.toTextLit newName) kvs
-updateName _ _ = Nothing
+updateName _ other = other
 
-addRawDeps :: Spago m => Config -> [PackageName] -> Expr -> m (Maybe Expr)
-addRawDeps config newPackages (Dhall.RecordLit kvs)
+addRawDeps :: Spago m => Config -> [PackageName] -> Expr -> m Expr
+addRawDeps config newPackages r@(Dhall.RecordLit kvs)
   | Just (Dhall.ListLit Nothing dependencies) <- Dhall.Map.lookup "dependencies" kvs = do
       case notInPackageSet of
         -- If none of the newPackages are outside of the set, add them to existing dependencies
-        []-> do
+        [] -> do
           oldPackages <- traverse (throws . Dhall.fromTextLit) dependencies
           let newDepsExpr
                 = Dhall.ListLit Nothing $ fmap (Dhall.toTextLit . packageName)
                 $ Seq.sort $ nubSeq (Seq.fromList newPackages <> fmap PackageName oldPackages)
-          pure $ Just $ Dhall.RecordLit $ Dhall.Map.insert "dependencies" newDepsExpr kvs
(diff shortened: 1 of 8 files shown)


jmackie:
Looks good to me :) Thanks for your work on this, it's much appreciated!

f-f:
You're welcome 🙂

PR #175, opened 2019-04-27
Rename `spago bundle` and `spago make-module` (issue #147)

### LINKED ISSUE #147: Rename `spago bundle` and `spago make-module`
These two commands do basically the same thing but have different enough names to be confusing.

So we should rename them to something clearer. Proposals:
1. `spago bundle-app`, `spago bundle-module`
2. `spago bundle app` `spago bundle module`
3. `spago bundle` defaults to "app" mode, and the module functionality is behind a flag `spago bundle --module`

Dretch:
`spago bundle` => `spago bundle-app`
`spago make-module` => `spago bundle-module`

I used `bundle-app` instead of `bundle app` for consistency with the existing command structure - i.e. currently we have `psc-package-{local-setup,insdhall,clean}` instead of `psc-package {local-setup,insdhall,clean}`.

I also avoided changing `--main` to `--module` because I don't have a view on what the best solution is there.

PR checklist:
- [X] Updated the version number [in `package.yaml`](https://github.com/spacchetti/spago/blob/master/package.yaml#L2) according to SemVer (`0.x.y.z`, `x` changes on breaking changes, `y` on additions, `z` on patches)

Fix #147


app/Main.hs
```diff
@@ -56,13 +56,13 @@ data Command
   -- | Run the project with some module, default Main
   | Run (Maybe ModuleName) BuildOptions
 
-  -- | Bundle the project, with optional main and target path arguments
+  -- | Bundle the project into an executable
   --   Builds the project before bundling
-  | Bundle (Maybe ModuleName) (Maybe TargetPath) NoBuild BuildOptions
+  | BundleApp (Maybe ModuleName) (Maybe TargetPath) NoBuild BuildOptions
```
f-f: Let's not remove entirely the old `Bundle` and `MakeModule` constructors, but let's have them just call `die` with the suggestion to call the new version of the command instead

f-f:
@Dretch update on the tests: it looks like it was an issue with the caches on Travis, now everything seems to be fine

Dretch:
Thanks @f-f -- I had run the tests locally using the command you suggested (which I figured out by looking at the Travis build). As you say, It would be helpful if this was clearer/easier.

I will have a look at implementing the the other things you have mentioned. :)

f-f:
CI is not passing because the way we download dependencies is flaky. This should be fixed once and for all in #188, but this is good to merge

PR #188, opened 2019-05-07
Add global packages cache and allow commit hashes

### LINKED ISSUE #133: Local Global Cache of Packages
Low priority.

Coming from [this comment](https://github.com/JordanMartinez/purescript-jordans-reference/issues/194#issuecomment-471276994):
> I'll also note that there's no technical issue against having a local global cache of packages, it's just that no one opened an issue about it yet :smile: (so please open one if you think it's a good thing to have)

Similar to `bower` in that it has a caching mechanism, provide a cache of already-downloaded dependencies, so that they can be used across multiple projects on the same computer.
Or something like that.

This would help lower the bandwidth my learning repo uses.

I'm not sure how hard this is to implement, nor whether it's a good idea (when full considerations are considered). At the very least, this can be further discussed. I started looking into Nix, but still haven't given it the attention it deserves.

### LINKED ISSUE #200: Support commit hashes in Package Set
```
      ace =
          mkPackage
          [ "effect"
          , "web-html"
          , "web-uievents"
          , "arrays"
          , "foreign"
          , "nullable"
          , "prelude"
          ]
          "https://github.com/slamdata/purescript-ace.git"
          "1fcf4f85f8f6c0bb83f4ef9b6de2adad9af82ad3"
```
gives me
```
error: no such remote ref 1fcf4f85f8f6c0bb83f4ef9b6de2adad9af82ad3
```
If I use `"master"` then it works, however I am using this all within a nix project and pp2n creates derivations that cannot use branch names. Not sure what I'm supposed to do 👎

f-f:
### Description of the change

Here we add a "global cache" for packages downloads. (fixes #133)

It works so that we download [this file containing metadata about commits and tags of the packages in the package set](https://github.com/spacchetti/package-sets-metadata/blob/master/metadataV1.json), and see if the refs we want to download are there.
If they are we also copy them to this global cache once they download, so we can reuse them in other projects.
An additional optimization that this introduces is that since we are only downloading from GitHub we can point to repo archives, so avoiding to use the `git` commands for everything (which are really slow). This means that we can also fetch archives with commit refs.
Now, if a package is not "globally cacheable" the old git method will be used in the local cache, and here we make it slightly slower (because we `git fetch` all the refs instead of just one), but allow to use commit hashes as refs too. So this fixes #200.

Another thing that we fix here is the "temporary directory" problem: before this we were fetching things inside the local cache directory, possibly giving "broken downloads" if the command failed. Now we do everything in temp directories, which makes is slightly more complicated but makes sure the result is always correct.

### PR checklist:

- [x] Download the `metadataV1` file if older than a day or so
- [x] Share the `Metadata` code with spago
- [x] Read from there instead of Github
- [x] Add management commands
- [x] Actually checkout commit refs in git repos anyways
- [x] Update the version number [in `package.yaml`](https://github.com/spacchetti/spago/blob/master/package.yaml#L2) according to SemVer (`0.x.y.z`, `x` bumps on breaking changes, `y` on additions, `z` on patches)
- [x] Add the change to the "Unreleased" section of the changelog
- [x] Add paragraph on global cache under "Explanations" in readme
- [x] Add test for installing by commit hash


f-f:
(cc @Dretch the new tests are a pleasure to use 😄)

Issue #191, opened 2019-05-08
Document how to add dev-only dependencies


thurn:
Trying to use spago for the first time, it is not apparent what the equivalent of `bower install --save-dev` is, if I want to add something like `purescript-spec` as a dependency for my unit tests.

Looking at some random examples on Github, it seems like people mostly just include these along with their regular dependencies? Perhaps that is the correct approach, but either way this feels like it merits a mention in the documentation since it's a core Bower feature.

thurn:
Oh, I guess this is a Duplicate of #178

f-f:
Yes! This is also going to be documented in https://github.com/spacchetti/spago/pull/201

Issue #187, opened 2019-05-06
Yarn-installed Spago doesn't work on Windows


fsoikin:
If I install Spago via Yarn (doesn't matter whether locally or globally), it seems that the shell wrappers are out of order somehow:

```
> yarn add spago
...
> npx spago
/bin/bash: C:\my_project\node_modules\.bin\..\spago\spago.exe: No such file or directory
```

At the same time, I cannot use Stack-installed Spago, because then it doesn't pick up my local Node environment, and ends up using my globally installed `purs`, which is of a different version:

```
>  spago build
spago.exe: Oh noes! It looks like the PureScript version installed on your system is
outdated for the package-set you're using.

installed version:   0.12.0
package-set version: 0.12.5

Please upgrade your `purs` version.
```

So the bottom line is: Spago is pretty much unusable to me. :-/

f-f:
Hi @fsoikin, sorry to hear this and thanks for the report!

Unfortunately none of the maintainers here runs Windows, so we have to rely on your help to debug these issues.

Yours looks like the issue described by @GoNZooo in https://github.com/spacchetti/spago/issues/157#issuecomment-478535480
I don't have any idea myself why this might be happening. Did you try to install only with `yarn` or also `npm`? Is the behaviour the same?

Btw you might be able to use the Stack-installed version by wrapping it up in an npm script. 
Let's say you edit your `package.json` to contain
```
  "scripts": {
    ...
    "spago": "spago",
    ...
```

Then you might be able to run the global spago in the local node environment like so, e.g. to run `build`: `npm run spago -- build` (not at my computer right now so I didn't try this but I did this kind of wrapping sometimes)

btrepp:
Hiyo.

Did you try in cmd.exe?

`yarn global add spago` installed things for me in `C:\Users\Beau\AppData\Local\Yarn\bin`.

`
C:\Users\Beau>dir C:\Users\Beau\AppData\Local\Yarn\bin
 Volume in drive C has no label.
 Volume Serial Number is BC1B-9F33

 Directory of C:\Users\Beau\AppData\Local\Yarn\bin

09-May-19  05:59 PM    <DIR>          .
09-May-19  05:59 PM    <DIR>          ..
09-May-19  05:59 PM               350 spago
09-May-19  05:59 PM                58 spago.cmd
               2 File(s)            408 bytes
               2 Dir(s)  440,973,336,576 bytes free
`

[...]

PR #201, opened 2019-05-15
Update readme

### LINKED ISSUE #142: Publishing a library with Spago
Coming from @f-f's comment [here (very end of the comment)](https://github.com/JordanMartinez/purescript-jordans-reference/issues/194#issuecomment-471276994):
> You can use spago to develop libraries, the only part that require you to use pulp and bower are versioning and publishing on bower. The reasons why this is required are stated here.
In general you'll have a bower file, but it's only necessary that it contains the right list of dependencies, the version can be * (at least for adding it to the package set).
> Basically all of this can be automated, and this would mean having spago version and spago publish

Followed by my comment [here (very end of comment)](https://github.com/JordanMartinez/purescript-jordans-reference/issues/194#issuecomment-471810222):
> Perhaps I should clarify that statement. The point is that you can't publish docs to Pursuit unless you are using bower at some point. While it may be possible to use spago to create libraries, I'm not sure whether that would be a good idea because I could "patch" a dependency in my package set before doing that and potentially break expectations.
> So, perhaps that could work but only if the package set did not include any overrides or additions?

### LINKED ISSUE #150: Provide a way to get dependency licenses
This is something we need in order to comply with GDPR if we want to switch to Spago.

### LINKED ISSUE #178: Question about separating executables / tests from libraries
There's a lot of good documentation here, maybe this is in there but I didn't see it at first glance. One of the biggest pains for me with using `pulp` + `bower` versus something like `cabal` is that there's only two categories for dependencies -- `dependencies` and `devDependencies` -- which really sucks when you have multiple executables, a test suite, and a library in the same project.

My fallback is of course using a `Makefile` or an elaborate `package.json`, making sure I separate _anything_ not used by the library from whatever's in the `src` directory, and use a million pulp flags and args. I'm wondering what `spago` is able to do to help, or what's in the plan.

### LINKED ISSUE #192: spago installed via npm does not seem to fully install
```
brandon@c9460f72c9e4:/wd$ type -a spago
spago is /usr/local/bin/spago
brandon@c9460f72c9e4:/wd$ cat /usr/local/bin/spago
#!/usr/bin/env bash
set -e

echo "This is a stub that will be replaced on install. \
It is needed to workaround https://github.com/yarnpkg/yarn/issues/3421"
```

I have a container for this that reproduces the issue: 305e449d39b9a146f5bfca1f335a80cd2c2bebe0 at https://github.com/bbarker/purescript-hodgepodge, or just use master's HEAD for now.

To run the container, use `psc.sh` in the repo, or a modified version that has fewer mounts, though I'm not sure what the behavior will be (should be mostly fine though). To build the container, you can run `build.socker.sh`.

### LINKED ISSUE #193: Issue with calling purescript functions from javascript
This commit demonstrates how I tried and failed to go from using a main-based function to using function calls from within `<script>` elements: https://github.com/bbarker/purescript-concur-starter/commit/98a28a498b44c3dd7a359c6bd550c6811ce6d510

Not necessarily more concise, but for completeness, the index.html and index.js are now:

```html
<!doctype>
<html>
<head>
  <meta charset="utf-8"/>
  <title>Hello Concur!</title>
  <script type="application/javascript" src="index.js"></script>
</head>
<body>
  <div id="root"></div>
  <script type="application/javascript" src="index.js">
    NotMain.notMain();
  </script>
</body>
</html>

```

```javascript
require("./output/NotMain");                  // builds but runtime error
// require("NotMain");                        // doesn't build
// import NotMain from "./output/NotMain";    // builds but runtime error
```

Granted, this was with parcel bundler - the Spago Readme sort of implied that i might need to use something other than`spago bundle-module` if using js libs (this uses React) - please correct me if I'm wrong.

CC'ing @ajnsit, the author of the original code, as he might have some insight, particularly on the parcel-bundler choice.

## Error in browser

```
SyntaxError: An invalid or illegal string was specified hmr-runtime.js:29:2
    parcelRequire<["node_modules/parcel-bundler/src/builtins/hmr-runtime.js"]< hmr-runtime.js:29
    newRequire purescript-concur-starter.e31bb0bc.js:47
    parcelRequire purescript-concur-starter.e31bb0bc.js:81
    <anonymous> purescript-concur-starter.e31bb0bc.js:8
```

## To reproduce

```
git clone https://github.com/bbarker/purescript-concur-starter.git
cd purescript-concur-starter
git checkout function_call_from_js
npm run start
# now browse to the dist/index.html page in your browser
```

f-f:
### Description of the change

- heavily update the README following [some very good guidelines on writing technical documentation](https://www.divio.com/blog/documentation/): now documentation is split between parts with different needs/tone instead of being in a giant tutorial:
  - "Tutorial"
  - "Howto"
  - "Explanation"
  - "Troubleshooting"
- address documentation for several issues:
  - fix #193 
  - fix #192 
  - #187 (issue will be kept open for further investigation)
  - fix #178, #191 
  - fix #150 
  - fix #142


JordanMartinez:
@f-f Alex and I read through that very helpful post, but we also read through a number of other resources. I'm not sure whether you'd be interested in it, but we summarized our conclusions [here](https://github.com/chexxor/purescript-documentation-discussion/blob/context-narrative/02-Context-or-Narrative/Good-Documentation.md)

f-f:
@JordanMartinez thanks, that's a nice read too! 👏

PR #212, opened 2019-05-25
Port tests from Python to Haskell

### LINKED ISSUE #177: Port tests to Haskell
Right now the tests are hacked together in Python, and this is an issue because:
- new contributors need to know also Python instead of just Haskell
- the setup is really fiddly - e.g. if the tests fail halfway through one has to remove files by hand

Dretch:
- This is a straight port, with no test re-organisation, and the tests are dependent on each other and so must run in the order defined.
- I changed the `spago psc-package-local-setup` test so it checks for success on the second run instead of failure, because the original Python test said it was checking for failure, but the check never actually happened, and so the test passed despite the command succeeding.
- Travis (OS X & Linux) and Appveyor (Windows) builds are working. The Travis build is not running here for some reason but it passed at https://travis-ci.org/Dretch/spago/builds/537154380

### PR checklist:

- [ ] Updated the version number [in `package.yaml`](https://github.com/spacchetti/spago/blob/master/package.yaml#L2) according to SemVer (`0.x.y.z`, `x` bumps on breaking changes, `y` on additions, `z` on patches)

Fix #177


test/SpagoSpec.hs
```diff
@@ -0,0 +1,158 @@
...
+
+    it "Spago should have set up a project" $ do
+
+      spago ["init"] >>= shouldBeSuccess
+
+    it "Spago should refuse to overwrite an existing project without -f" $ do
+
+      spago ["init"] >>= shouldBeFailure
+
+    it "Cleaning of config files should succeed" $ do
```
f-f:
```suggestion
```

If we remove this line the cleanup will be done as part of the previous step, as it happens in other cases too

f-f:
```suggestion
```

If we remove this line the cleanup will be done as part of the previous step, as it happens in other cases too

test/SpagoSpec.hs
```diff
@@ -0,0 +1,158 @@
...
+
+    it "Spago should be able to add dependencies" $ do
+
+      spago ["install", "-j10", "simple-json", "foreign"] >>= shouldBeSuccess
+      mv "spago.dhall" "spago-install-success.dhall"
+      checkFixture "spago-install-success.dhall"
+
+    it "Spago should always succeed in doing init with force" $ do
+
+      spago ["init", "-f"] >>= shouldBeSuccess
```
f-f: I think we can move this init to the next case, since this is basically only "setting up the environment"

test/SpagoSpec.hs
```diff
@@ -0,0 +1,158 @@
...
+testDir :: FilePath
+testDir = "test/spago-test"
+
+clean :: IO ()
+clean = do
+  whenM (testdir testDir) (rmtree testDir)
+  mkdir testDir
+
+spec :: Spec
+spec = beforeAll clean $ around_ (withCwd testDir) $ do
```
f-f: In the psc-package spec we clean both before and after but here only before, is it accidental or there's a reason for it?

Dretch: It was intentional, but I don't mind changing it:
- `test/psc-package-local-test` is checked into git, so it gets cleaned afterwards to avoid unwanted changes showing up in the git diff.
- `test/spago-test` is not checked into git (the folder is in the `.gitignore`), so we can leave the crumbs from the tests there for possible debugging purposes.

f-f: Oh right, let's keep it like this 👍

test/SpagoSpec.hs
Dretch: Done

f-f:
Looks great, thanks!

I added you as a collaborator so you can merge once CI passes 🙂

PR #215, opened 2019-05-26
Fix Yarn & NPM shim issues on Windows

### LINKED ISSUE #187: Yarn-installed Spago doesn't work on Windows

stkb:
### Description of the change

(Hopefully) a fix for the shim issues on Windows described in https://github.com/spacchetti/spago/issues/187#issuecomment-494029142. Namely:
- Shims created by Yarn are broken because they try to run spago via bash.
- Shims created by NPM for bash don't work because they reference an undefined `$basedir` variable

To fix the NPM issue in particular, instead of having the package try to shim to the spago binary directly, a "2nd-level shim" nodejs script is used. This script then runs the spago binary. Since this shim doesn't
get changed during the package install process, Yarn doesn't get confused either. The script is only used on Windows.

So the result is this:
- The binary declared in the package's `package.json` is simply `spago`.
- A `spago` file is included with the package. This is a nodejs script that simply runs an expected sibling `spago.exe` file.
- On installing the package:
  - (On Windows) Yarn/NPM create their shims to call the `spago` script under nodejs. These have no issues.
  - (On Mac/Linux) A symlink is created to the `spago` file as normal.
- After the package is installed, in a `postinstall` script:
  - (On Windows) the `spago.exe` binary file is downloaded, so the nodejs shim will now work.
  - (On Mac/Linux) The binary file is downloaded, which since it's also named `spago`, simply replaces the javascript file, and the symlinks point to this file instead.

Have manually tested on all combinations of the following on Windows: Cmd, Git-Bash, "Linux" (WSL); NPM and Yarn; global and local package installations.

I guess if this patch proves successful then the note about Yarn in the readme can also be removed.

### PR checklist:

- [ ] (Not sure if this is needed?) Updated the version number [in `package.yaml`](https://github.com/spacchetti/spago/blob/master/package.yaml#L2) according to SemVer (`0.x.y.z`, `x` bumps on breaking changes, `y` on additions, `z` on patches)

Fix #187

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `spago-curator` tool to generate metadata from the package set (#202)
 
 Bugfixes:
-- Another attempt to fix NPM and Yarn installations on Windows (#187)
+- Another attempt to fix NPM and Yarn installations on Windows (#215, #187)
 
 ## [0.8.0] - 2019-05-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `spago-curator` tool to generate metadata from the package set (#202)
 
+Bugfixes:
+- Another attempt to fix NPM and Yarn installations on Windows (#187)
+
 ## [0.8.0] - 2019-05-16
 
 Breaking changes:
@@ -145,4 +148,3 @@ Main changes from the previous "spacchetti-cli" incarnation:
 - Freeze `spacchetti` package-set import in `packages.dhall`, so `dhall` caching works for subsequent executions
 - Move to v4.0.0 of `dhall`
 - Add integration tests for most of the commands (#31, #30)
-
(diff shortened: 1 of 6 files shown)


stkb:
Don't know why the MacOS test failed. It doesn't even use the npm package?

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `spago-curator` tool to generate metadata from the package set (#202)
 
+Bugfixes:
+- Another attempt to fix NPM and Yarn installations on Windows (#187)
```
f-f:
```suggestion
- Another attempt to fix NPM and Yarn installations on Windows (#215, #187)
```

stkb:
OK have added those changes

f-f:
Thank you @stkb!

PR #224, opened 2019-05-30
Fix psc-package-clean and test cleanup on Windows

stkb:
### Description of the change

Replaces use of `Turtle.rmtree` with `System.Directory.removePathForcibly` to fix the following issues on Windows.

`spago psc-package-clean` (and its test) fail because:
1. Git marks files under its /objects subdirectory as read-only
2. Turtle.rmtree fails to delete files marked as read-only

So `rmtree` fails to remove any folder that's a git repository.

As well as `psc-package-clean`, general test cleanup fails, meaning the test suite will fail completely when run for the second time.

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+Bugfixes:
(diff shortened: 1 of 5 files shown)


stkb:
Thanks!

f-f:
You're welcome 🙂

PR #234, opened 2019-05-31
Get PureScript binary for Travis from GitHub releases

justinwoo:
i don't know how you fix appveyor, maybe someone should get a new
chocolatey package.

### Description of the change

TODO

### Checklist:


appveyor.yml
```diff
@@ -40,7 +39,7 @@ build_script:
   - tar -zcvf windows.tar.gz spago.exe
 
 test_script:
-  - set PATH=C:\projects\spago\bin;%PATH%
+  - set PATH=C:\projects\spago\bin;C:\projects\spago\purescript;purescript;%PATH%
```
stkb: extra `purescript;`, is this a typo?

f-f: No, I didn't know which one of the two would work so I put both 😄

f-f: Fixed 🙂

PR #235, opened 2019-05-31
Repl: run with callCommand instead of callProcess (fixes #227)

### LINKED ISSUE #227: repl command fails on Windows
`spago repl -- --port 3600` command fails on Windows.

This error is given:
`spago.exe: purs: callProcess: does not exist (No such file or directory)`

When I add the -v option to print the `purs` command executed output does not change.

purs command itself also results in the following error message, but apparently runs:
`purs repl: No input files; try running 'pulp psci' instead.
For help getting started, visit https://github.com/purescript/documentation/blob/master/guides/PSCi.md
Usage: For basic information, try the '--help' option.

I have installed the package as per `spago README.md`.

Any suggestions?

stkb:
### Description of the change

This is just a quick fix so that `spago repl` will start the repl on Windows where PureScript is installed with NPM. However a rewrite of `Purs.hs` (as discussed in #227) that deals with all issues, is probably needed eventually.

I'm wondering if the AppVeyor script should install `purescript` with npm instead of Chocolatey to catch these types of issues?

### Checklist:


f-f:
@stkb sorry if I broke CI, I got too excited about upgrading to the 0.13 package set without waiting for CI to finish.

It looks like we cannot rely on the npm nor chocolatey packages being up to date, so e.g. #234 is switching to just downloading the binary from GitHub releases. Do you have some idea on how to do the same with AppVeyor?

f-f:
Nvm, I fixed it 🎉 

I'll rebase here once CI passes there 😄

stkb:
Tests have passed... I'm guessing this is ok to merge now

f-f:
Yep, thank you for doing this!

PR #246, opened 2019-06-03
Export ps

bbarker:
### Adding an --export option so that exported modules can be used from JavaScript easily.

This is a switch for `bundle-module`; if `-e` or `--export` are given then `export default PS;` is tacked to the end of the file.

### Checklist:

   - I tested this manually, but haven't looked to see if it is applicable - let me know.

**Note**: Suggest a squash and merge

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Bugfixes:
 - Fix `spago repl` starting on Windows where PureScript was installed with NPM (#235, #227)
 - Fix missing filenames when encountering parse errors in Dhall files (#241, #222)
 - Download packages in local repo instead of global tempdir (#243, #220)
+- Adds `--export` switch for `bundle-module` to allow use of PureScript modules from JS (#246)
 
 Other improvements:
 - Tests: test suite now works fully on Windows (#224)
--- a/README.md
+++ b/README.md
@@ -643,6 +643,28 @@ $ node -e "console.log(require('./index).main)"
 [Function]
 ```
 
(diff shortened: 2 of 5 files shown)


f-f:
Hi @bbarker thanks for the PR!

Looking at it, there are some things that I'm not sure of:
- I wouldn't like to add another flag in there (it's a slippery slope and I'd like to avoid the situation in which getting a build right shouldn't be a game of composing flags)
- `export default ..` is an ES6-only syntax, which would make spago's output less compatible with some tools

Would it work for your usecase to use the `import * as ..` syntax?

E.g. if you do `spago bundle-module --export --main Foo --to index.js`, then in your JS file you can do `import * as Foo from 'index.js'`, and then you should be able to call e.g. `Foo.bar()`

bbarker:
@f-f Just to check your last suggestion with `import * as Foo from 'index.js'`, what would `--export` be doing in this case (if anything)?

f-f:
@bbarker my suggestion is to be used with the current Spago version, so if that works we can skip adding the `--export` flag

bbarker:
Cool, well I thought I had tried that! Too many permutations to try. But your suggestion works. In my index.js I just have:

```javascript
import * as Metajelo from  "./index.opt.js";
window.Metajelo = Metajelo;
```

This seems to work for me, so maybe i'll just open a separate small PR that modifies the same section of the README I just did with this info?

f-f:
> This seems to work for me, so maybe i'll just open a separate small PR that modifies the same section of the README I just did with this info?

@bbarker yes that would be great thanks! 👏

PR #248, opened 2019-06-03
Enable easier usage of stack within nix

bbarker:
### Description of the change

I tested this in my nix setup using `stack --nix build` - I think this is fairly standard.

### Checklist:

 - I don't think it is really necessary? It should only affect new users building spago in a particular way.
 - I don't think it is really necessary.
 - N/A


f-f:
Thanks!

PR #249, opened 2019-06-03
adding notes for using exported modules from JavaScript

bbarker:
### Checklist:

 - N/A
 - Well, not really a new feature, more of a hidden feature!
 - N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -643,6 +643,27 @@ $ node -e "console.log(require('./index).main)"
 [Function]
 ```
 
+If you want to be able access modules and their functions from JavaScript land,
+(highly useful for use in `<script>` tags in the browser), then you can use
+a recipe like the following. In your index.js that is input the the bundler
+(e.g. Parcel):
+
+```javascript
+// index.js
+
+import * as Foo from  "./index.opt.js";
+window.Foo = Foo;
+```
+
+This references your exported module `Foo` and makes it globally accessible;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -643,6 +643,27 @@ $ node -e "console.log(require('./index).main)"
 [Function]
 ```
 
+If you want to be able access modules and their functions from JavaScript land,
+(highly useful for use in `<script>` tags in the browser), then you can use
+a recipe like the following. In your index.js that is input the the bundler
+(e.g. Parcel):
```
f-f: I'm a bit confused by this, because just in the next section we talk about using the bundle from JS, so maybe this content is better suited there?

README.md
```diff
@@ -643,6 +643,27 @@ $ node -e "console.log(require('./index).main)"
...
+
+```javascript
+// index.js
+
+import * as Foo from  "./index.opt.js";
```
f-f: ..but there we instruct to use something like `require('./output/Main')`, why recommend this instead?

bbarker:
All good points, I thought I tried some similar suggestions, but will try again soon and see what works and what doesn't in this regard.

f-f:
Since there's no activity since 3 months I'll close this as I'd like to keep the tracker tidy. Let's reopen when we want to get this in shape again 🙂

PR #250, opened 2019-06-04
When asserting directory permissions produce an error message

zudov:
Previously a gnarly `guard` would cause this (when `.spago` is not writable):

```
$ spago install
spago: user error (mzero)
```

Now it would produce:

```
$ spago install
spago: Directory ".spago" is not accessible. Permissions {readable = True, writable = False, executable = False, searchable = True}
```


f-f:
Thanks! ❤️

PR #253, opened 2019-06-06
Read purs version from inside the set instead of its GitHub tag

### LINKED ISSUE #225: Handle the case where the compiler version is newer than the package set compiler version
For example, trying to build a v0.12.5 package set with a v0.13.0 compiler:
```
spago: Oh noes! It looks like the PureScript version installed on your system is

installed version:   0.13.0
package-set version: 0.12.5

Please upgrade your `purs` version.
```
Also, is it possible to tell spago that I know what I'm doing and I'd like to disable this check?

f-f:
Currently `spago` tries to find out which `purs` version is supported by the set by reading the GitHub tag the upstream comes from. This brings a whole bunch of issues, e.g. we cannot ensure anything if the package set comes from somewhere else, this check is hard to override, etc.

Once https://github.com/purescript/package-sets/pull/343 lands we'll be able to read the "minimum supported purs version" just as the version of one of the packages in the set.

This fixes #225, because we:
- fix the error message to just say "SemVer mismatch" instead of newer/older
- make the "minimum purs version" easily overridable, as it's just the version of a package, so there's no need for a flag
- make this version check actually robust, as we're not trying to climb the Dhall import chain from the AST anymore

An unrelated change coming in this PR too is that I run the package through [weeder](http://hackage.haskell.org/package/weeder) and trimmed the dependencies we use

### DIFF
--- a/app/Curator.hs
+++ b/app/Curator.hs
@@ -28,11 +28,10 @@ import qualified Turtle
 
 import           Data.Aeson.Encode.Pretty       (encodePretty)
 import           Spago.GlobalCache
-import           Spago.PackageSet               (Package (..), PackageName (..), PackageSet,
-                                                 Repo (..))
+import           Spago.PackageSet               (Package (..), PackageName (..), Repo (..))
 
 type Expr = Dhall.DhallExpr Dhall.Import
-
+type PackageSetMap = Map PackageName Package
 
 
 data SpagoUpdaterMessage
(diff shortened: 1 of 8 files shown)


f-f:
Note: before CI can pass on this we'll have to wait for a new package set tag including purescript/package-sets#343

f-f:
cc @hdgarrood: is the error message acceptable or should I detail some more there?

hdgarrood:
Looks good :)

Issue #257, opened 2019-06-10
Fails to install branches with / in the branch name


codedmart:
We commonly use branch names like `kimi/update/table`. Spago fails to install branches with `/` in the branch name. Example output error that I am seeing.
```
spago: Installation failed.

Error:

.spago/__download-lumi-components-kimi/update/table-c8f2fba29333e838: createDirectory: does not exist (No such file or directory)
```

f-f:
This should be fixed anyways, but I'll note that branches are not really recommended with `spago`: the reason is that if you push more commits to the branch after you installed it once, spago won't pick up the new changes and you'll have to delete the `.spago` folder in order to make it pull the new stuff.

You could avoid all of this by just pointing at commits

PR #258, opened 2019-06-10
Replace `/`'s for cache version dir name

codedmart:
* Fixes GH-257

### Description of the change

TODO

### Checklist:


test/SpagoSpec.hs
```diff
@@ -92,6 +91,11 @@ spec = beforeAll clean $ around_ (withCwd testDir) $ do
       writeTextFile "packages.dhall" "let pkgs = ./packagesBase.dhall in pkgs // { simple-json = pkgs.simple-json // { version = \"d45590f493d68baae174b2d3062d502c0cc4c265\" } }"
       spago ["install", "simple-json"] >>= shouldBeSuccess
 
+    it "Spago should be able to install a package version by branch name with `/`'s in it" $ do
+
+      writeTextFile "packages.dhall" "let pkgs = ./packagesBase.dhall in pkgs // { simple-json = pkgs.simple-json // { version = \"compiler/0.12\" } }"
```
codedmart: Not sure if this is the best way to handle/test this since this branch could be removed at any time.

f-f: Agreed, let's not depend on some random branches. I created [this properly named branch](https://github.com/spacchetti/purescript-metadata/tree/spago-test/branch-with-slash) in this fake package, and I promise I won't remove it 😄 

Note that the `metadata` package is already in the set so you can override the version in the same way

src/Spago/FetchPackage.hs
```diff
@@ -185,3 +185,9 @@ getLocalCacheDir (packageName, Package{ repo = Remote _, ..}) = do
 getLocalCacheDir (_, Package{ repo = Local path }) =
   Text.unpack path
 
+
+-- | Returns the name of the cache dir based on version from `PackageName`
+--   If the branch version you are trying to install has `/`'s in the branch name replace
+--   the `/`'s and return the folder name
+getCacheVersionDir :: Text -> Text
+getCacheVersionDir = Text.replace "/" "-"
```
hdgarrood: I think ideally this function should be injective, i.e. different inputs map to different outputs. Admittedly a bit of an edge case, but happens with this code if you have a repo with two branches called `harry/fix-1` and `harry-fix-1`?

codedmart: :man_facepalming: Yes good point thanks!

codedmart:
@f-f OK updated now. As per @hdgarrood's suggestion also escaping a few other characters that could be in branch names as well.

hdgarrood:
I think we should include % in the list of characters to be escaped as currently I think “%2F” and “/“ will both map to “/“.

Also I’ve just remembered that case insensitive file systems are a thing...

f-f:
@hdgarrood agreed, this is currently still not injective but if we add `%` to the list it will. 

And yes, e.g. my mac is case insensitive, so two branches with different cases would still clash. I think though we can just add a `TODO` about this and hope no one tries to fetch two branches differing only by case 😄 

Btw @codedmart if you want to have some fun you could setup [Quickcheck](http://hackage.haskell.org/package/QuickCheck) with a property that requires that `getCacheVersionDir` cannot map two different names to the same output (I'll merge this PR in any case, so if you want to do this feel free to open another PR. Though it's probably more interesting if you write the test before fixing the implementation, so you can marvel at Quickcheck proving things)

codedmart:
Yup forgot `%` and the case sensitivity. Will fix. I will look at the test as well. Thanks!

hdgarrood:
I suppose one option is to say that any uppercase character needs escaping too; in my experience branch names rarely contain uppercase characters.

hdgarrood:
Also if you want to write a QuickCheck property for this, I think you'd need to be a bit careful if you want to end up with one that's actually useful; if you just write
```
quickCheck \x y -> (x /= y) ==> (getVersionCacheDir x /= getVersionCacheDir y)
```
then the chance of a collision is so astronomically tiny that that test is only really slowing down your test suite. I'd recommend writing your own generator which uses a much smaller set of characters to sample from, in particular including the ones we know to be problematic. Then, instead of running lots of individual tests, we should just run a single test where we generate a large _set_ of potential branch names, apply the `getVersionCacheDir` function to all of them, assemble the results into another set, and check that the result set is the same size as the input set. I think this is the only way you'll get the test to fail reliably if the function is not actually injective.

f-f:
@codedmart this got a bit out of sync after all the ZuriHac's contributions, so I'll merge it now so we can ship it in the upcoming release later today. Thank you! 🙂

PR #259, opened 2019-06-11
0.8.3 → 0.8.4

f-f:
Note that I'll wait for #258 before merging this


f-f:
Update: I originally wanted to wait for #258, but I'd also like to have a release before ZuriHac, so we can cut another release with #258 later

PR #264, opened 2019-06-14
Report all missing packages

### LINKED ISSUE #223: Report all missing packages together
If there are a number of packages which are listed as dependencies in my `spago.dhall` but which do not exist in my package set, `spago` reports these one by one. For example:

```
$ spago build
spago: Package `fixed-precision` does not exist in package set
# Add `fixed-precision` ...
$ spago build
spago: Package `reflection` does not exist in package set
# Add `reflection` ...
$ spago build
...
```

I think it would be better to report all missing packages together, if possible. Of course it won't always be possible to do so: suppose package A depends on package B, and package B depends on C, and I add A as an addition without adding B or C; in this case we might not know that C is missing until we add B.

For a minimal reproduction example, run `spago init` in an empty directory, and modify the `spago.dhall` file as follows:
```
{-
Welcome to a Spago project!
-}
{ name =
    "my-project"
, dependencies =
    [ "effect", "console", "psci-support", "blah", "what" ]
, packages =
    ./packages.dhall
}
```
and then run `spago build`. The error you receive is:
```
spago: Package `blah` does not exist in package set
```
However, it would be nicer to get something like
```
spago: The following packages do not exist in your package set:
  - blah
  - what
```

eahlberg:
### Description of the change
If there are multiple dependencies which do not exist in the package set, instead of just reporting the first missing package, all missing packages will be reported which fixes #223. In the same manner, all circular dependency errors will be reported. 

### Checklist:

### DIFF
--- a/src/Spago/Packages.hs
+++ b/src/Spago/Packages.hs
@@ -85,47 +85,24 @@ getProjectDeps :: Spago m => Config -> m [(PackageName, Package)]
 getProjectDeps Config{..} = getTransitiveDeps packageSet dependencies
 
 
-fst' :: (a, b, c) -> a
-fst' (x, _, _) = x
-
-snd' :: (a, b, c) -> b
-snd' (_, x, _) = x
-
-trd' :: (a, b, c) -> c
-trd' (_, _, x) = x
-
 -- | Return the transitive dependencies of a list of packages
 --   Code basically from here:
 --   https://github.com/purescript/psc-package/blob/648da70ae9b7ed48216ed03f930c1a6e8e902c0e/app/Main.hs#L227
 getTransitiveDeps :: Spago m => PackageSet -> [PackageName] -> m [(PackageName, Package)]
 getTransitiveDeps PackageSet{..} deps = do
   echoDebug "Getting transitive deps"
-  let xs  = map (go Set.empty Set.empty Set.empty) deps
+  let (packageMap, notFoundErrors, cycleErrors) = foldMap (go Set.empty Set.empty Set.empty) deps
 
-  let cycleErrors = foldMap trd' xs
-  let notFoundErrors = foldMap snd' xs
-  let packageMap = Map.toList $ foldMap fst' xs
-
-  handleErrors cycleErrors notFoundErrors packageMap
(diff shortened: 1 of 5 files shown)


src/Spago/Packages.hs
```diff
@@ -85,27 +85,48 @@ getProjectDeps :: Spago m => Config -> m [(PackageName, Package)]
...
 getTransitiveDeps :: Spago m => PackageSet -> [PackageName] -> m [(PackageName, Package)]
 getTransitiveDeps PackageSet{..} deps = do
   echoDebug "Getting transitive deps"
-  Map.toList . fold <$> traverse (go Set.empty) deps
+  let xs  = map (go Set.empty Set.empty Set.empty) deps
+
+  let cycleErrors = foldMap trd' xs
+  let notFoundErrors = foldMap snd' xs
+  let packageMap = Map.toList $ foldMap fst' xs
+
```
eahlberg: It'd be cool if we could avoid doing the fold-map-dance three times like above 🤔

src/Spago/Packages.hs
```diff
@@ -85,27 +85,48 @@ getProjectDeps :: Spago m => Config -> m [(PackageName, Package)]
...
 
+fst' :: (a, b, c) -> a
+fst' (x, _, _) = x
+
+snd' :: (a, b, c) -> b
+snd' (_, x, _) = x
+
+trd' :: (a, b, c) -> c
+trd' (_, _, x) = x
+
```
eahlberg: It'd be nice if these three functions could be avoided, or at least be moved to somewhere more appropriate.

eahlberg:
@f-f thanks for the intro to `spago` earlier today! If you (or anyone else) has feedback about the code in the PR, it's really appreciated! Regarding the correctness, I'm still confused why the tests fail when it works locally by running `spago build` in a folder as described in #223, but I'll try to dig into that some more.

eahlberg:
Apparently, the tests pass on CI 🤔

eahlberg:
Alright, so I got a bit further with this:
- the tests now pass locally without any change (that I'm aware of) from my side, so maybe it was some kind of environment issue 🤷‍♂️
- I also refactored some of the ugly parts
- I'm a bit unsure whether this warrants updating or adding some test case, so that'd be good if someone could take a look at

PR #268, opened 2019-06-15
Improve packages test

eahlberg:
### Description of the change

Improve packages test by:
- [x] using a fixture in the test which checks for missing packages
- [x]  using a fixture in the test which checks for circular dependencies

### Checklist:

- [ ] ~~Added the change to the "Unreleased" section of the changelog~~


eahlberg:
@f-f not sure if this is what you had in mind, but I gave it a go!

I also intend to add a similar, independent, test for circular dependencies since I couldn't find any in the test suite. Does that sound reasonable?

Issue #159, opened 2019-03-26
Migrate Bower projects when doing `init`


antislava:
It would probably be nice to be able to have the dependency list as an optional input parameter to `spago init`, especially when working with existing packages based on `bower`. I suspect this is somehow related to #24 but I wasn't sure about the main focus there (most links there refer to no longer existing pages, e.g. `from-bower` scripts). Alternatively, one could optionally, define dependencies list in separate file `spago-dependencies.json` (which, being a simple list of strings, would also be a valid `dhall` file imported by `spago.dhall`). This file could be generated or overwritten by a simple script (probably, as a temporary solution before deciding whether to integrate the dependency import functionality deeper into the `spago` tool or not - most likely, it should remain a separate command/script keeping in mind the principle of separation of concerns), e.g.
```
jq '(.dependencies? | keys? // []) + (.devDependencies? | keys? // []) | map(ltrimstr("purescript-"))' ./bower.json > spago-dependencies.json
```

```
cat spago.dhall

{ name =
    "package-name"
, dependencies =
    ./spago-dependencies.json
, packages =
    ./packages.dhall
}
```

f-f:
Hi @antislava! This is a good analysis, and I think it'd be worth adding your script to a FAQ section "Migrating from bower"

The next step after documenting this would be to automatically migrate a bower project when finding one. We do already migrate psc-package projects automatically in `spago init`, and I think it would be good to also support `bower.json` migration in the same way (that is, I don't think we should be able to provide options to init, as it should just detect the bowerfile)

I probably won't manage to do this myself as I'm concentrating on other things on the road to 1.0 (because your script does the job pretty well already), so I'll mark this as "help wanted" and I'm available to provide support, answer questions and give pointers if someone wants to implement this 🙂 (this is valid also for #24, which is about importing bower dependencies as `additions` in your local package set)

PR #272, opened 2019-06-16
Migration from Bower

jonathanlking:
### Description of the change

This (when completed) addresses #159.

### Checklist:


jonathanlking:
This was originally intended as a command to verify that package set satisfies bower dependencies, however after speaking to @f-f I was made aware of #159. I'm going to refactor it to run on `spago init` and add the dependencies to the `spago.dhall`.

f-f:
@jonathanlking I think we're done with the invasive changes to Bower-related code after #324, would you still like to work on this? If yes then we can rebase on #324, if not I can pick this up as it's needed for 1.0 anyways 🙂 

Also no worries about fixing #159 here, I'd be fine merging pretty much what you have here already (except we should remove the command line stuff) and tackling the rest in a later PR - the policy is that I'm fine even merging code that is not used anywhere yet if it's needed as a stepping stone for other things

jonathanlking:
@f-f Thanks for letting me know 😄
I think your suggestion about removing the command line interface an merging sounds good. I still think there is some value in this check being "continuously" run, rather than just at conversion - given that we need to keep a `bower.json` file around for Pursuit.
I intend to rebase and remove the command line interface. I had forgotten about local packages when implementing this, so they are currently unsupported.

f-f:
@jonathanlking thank you! A few notes:
- note that when using `spago bump-version` (implemented, shipping in next version) and `spago publish` (to be implemented soon) the `bower.json` is generated from the `spago.dhall`, and not viceversa. I.e. this conversion will need to happen only when porting a bower package to Spago, and after that Spago will take care of keeping bower file consistent over time
- no worries about local packages, we'll fix them in the process of taking this code into use (also there's #301 that shuffles the local packages a bit, so it's ok to not worry about them here)

f-f:
@jonathanlking I think if you merge https://github.com/jonathanlking/spago/pull/1 then it will pass CI

jonathanlking:
@f-f thanks 🙂 - I've merged it and fixed the warnings.

PR #280, opened 2019-06-17
Add support for global REPL

### LINKED ISSUE #168: Global spago repl
Something I often do with `stack` is just open a terminal and do `stack ghci` or `stack ghci --package text --package bifunctors` and play around with types and what not.

This doesn't seem possible with `spago` at the moment, would it be worth implementing this?

eahlberg:
### Description of the change

Adds support for starting a REPL within a folder which has not been setup as a spago project. Fixes #168.

### Checklist:


src/Spago/Build.hs
```diff
@@ -87,10 +89,26 @@ build BuildOptions{..} maybePostBuild = do
...
+    else Temp.withSystemTempDirectory "spago-repl-tmp" $ \dir -> do
+      Turtle.cd (Turtle.decodeString dir)
+
+      Packages.initProject False
+
+      config <- Config.parseConfig
+      deps <- Packages.getProjectDeps config
+      let globs = Packages.getGlobs deps <> Config.configSourcePaths config <> sourcePaths
+
+      Packages.install Nothing Nothing (fmap fst deps)
```
eahlberg: Should the flags to `install` come from the user?

eahlberg:
@f-f I decided to take a look at this and I tried to implement it according to https://github.com/spacchetti/spago/issues/168#issuecomment-478526643. However, I’m a bit unsure of a couple of things (in addition to the comment above):
- Any idea for how this could be tested? I couldn’t find a test for `spago repl`, is there a specific reason for that?
- I saw that `PackageSet.ensureFrozen` was called in `Config.ensureConfig`, which is not used for this, but I'm not sure what freezing a package set means. Should it be used here?

src/Spago/Build.hs
```diff
@@ -87,10 +89,26 @@ build BuildOptions{..} maybePostBuild = do
 repl :: Spago m => [Purs.SourcePath] -> [Purs.ExtraArg] -> m ()
 repl sourcePaths passthroughArgs = do
   echoDebug "Running `spago repl`"
-  config <- Config.ensureConfig
-  deps <- Packages.getProjectDeps config
-  let globs = Packages.getGlobs deps <> Config.configSourcePaths config <> sourcePaths
-  Purs.repl globs passthroughArgs
+
+  hasPackagesDhall <- testfile "packages.dhall"
```
f-f: I think here we could use the [`try`](https://www.stackage.org/haddock/lts-13.25/base-4.12.0.0/Control-Exception.html#v:try) pattern on the config to make this cleaner, something like:

```hs
try (Config.ensureConfig) >>= \case
  Right config -> do
    -- start the normal project repl with the `config`
    ...
  Left err -> do
    -- here we do the global repl stuff
    echoDebug err
    ...
```

The reason is that even if we get the boolean right in the `testfile` the `ensureConfig` might still except, but instead we can just catch the exception with the `try`

src/Spago/Build.hs
f-f: Yes! Also note that we don't need the full blown `install`, but we can call just the `fetchPackages` part here, e.g. see
https://github.com/spacchetti/spago/blob/92e9c64bdde0373a64514def96b7f78738601909/src/Spago/Packages.hs#L169

src/Spago/Build.hs
```diff
@@ -87,10 +89,26 @@ build BuildOptions{..} maybePostBuild = do
...
+  if hasPackagesDhall
+    then do
+      config <- Config.ensureConfig
+      deps <- Packages.getProjectDeps config
+      let globs = Packages.getGlobs deps <> Config.configSourcePaths config <> sourcePaths
+      Purs.repl globs passthroughArgs
+    else Temp.withSystemTempDirectory "spago-repl-tmp" $ \dir -> do
```
f-f: In the past we had permissions issues with `withSystemTempDirectory`, so maybe we could use `withTempDirectory` with the [global cache directory](https://github.com/spacchetti/spago/blob/92e9c64bdde0373a64514def96b7f78738601909/src/Spago/GlobalCache.hs#L150) as base path?

src/Spago/Config.hs
```diff
@@ -2,6 +2,7 @@
 module Spago.Config
   ( makeConfig
   , ensureConfig
+  , parseConfig
```
f-f: I'd like to not expose `parseConfig`, could we use `ensureConfig` instead since it takes care of some more things?

f-f:
@eahlberg looks good so far! I added some minor comments, and to answer your other points:
- let's skip tests for this one as I don't have ideas about how to test `spago repl`
- "freezing" a Dhall file means to add `sha256` hashes to the remote imports (the package set URL in our case). Since in this case we're initing a new project it should be a no-op as the template already has hashes, so it should be fine to use `ensureConfig`.
  We have [a small section about freezing](https://github.com/spacchetti/spago#package-set-caching-problems) in the README, but it's very small as it's a detail that's supposed to be automated away from the user. However in there you can find a link to a longer description of Dhall's freezing mechanism

f-f:
I just noticed that in addition to the install flags we'll also need to add the `--package` flag to add packages as specified in #168. It's going to look exactly as the `packageNames` argument:
https://github.com/spacchetti/spago/blob/92e9c64bdde0373a64514def96b7f78738601909/app/Spago.hs#L158
..except it will be an `opt` instead of an `arg`

src/Spago/Build.hs
eahlberg: Agree! Actually didn’t even consider the case where another exception might occur

src/Spago/Build.hs
eahlberg: Ok, good to know!

src/Spago/Config.hs
eahlberg: Sure

CHANGELOG.md
```diff
@@ -187,7 +191,7 @@ New features:
 
 Bugfixes:
 - Don't overwrite files when doing `init`, just skip the copy if some file exists (#56)
-- Print `git` output in case of failure when doing `install` (#54, #59) 
+- Print `git` output in case of failure when doing `install` (#54, #59)
```
eahlberg: Not too happy with the diff here, apparently my editor removed the whitespaces, other than that it should be the same. I'll remove if you'd like.

app/Spago.hs
```diff
@@ -156,6 +156,7 @@ parser = do
     sourcePaths = CLI.many (CLI.opt (Just . SourcePath) "path" 'p' "Source path to include")
     packageName = CLI.arg (Just . PackageName) "package" "Specify a package name. You can list them with `list-packages`"
     packageNames = CLI.many $ CLI.arg (Just . PackageName) "package" "Package name to add as dependency"
+    replPackageNames = CLI.many $ CLI.opt (Just . PackageName) "package" 'r' "Package name to add to the REPL as dependency"
     passthroughArgs = many $ CLI.arg (Just . ExtraArg) " ..any `purs compile` option" "Options passed through to `purs compile`; use -- to separate"
```
eahlberg: Unfortunately there’s a nameclash between `-p/—path` here and the shortname for `--package`. Do you know any conventions or an alternative short name for `--package?` (`’r’` is just there temporarily)

src/Spago/Build.hs
```diff
@@ -84,13 +86,33 @@ build BuildOptions{..} maybePostBuild = do
...
-  deps <- Packages.getProjectDeps config
-  let globs = Packages.getGlobs deps <> Config.configSourcePaths config <> sourcePaths
-  Purs.repl globs passthroughArgs
+
+  try Config.ensureConfig >>= \case
+    Right config -> do
+      deps <- Packages.getProjectDeps config
+      let globs = Packages.getGlobs deps <> Config.configSourcePaths config <> sourcePaths
+      Purs.repl globs passthroughArgs
+    Left (err :: SpagoError) -> do
```
eahlberg: I'm new to exceptions in Haskell. Is it correct to catch `SpagoError` here?

eahlberg:
@f-f thanks for the feedback! 🙇‍♂️ I think I’ve addressed all of it and left a couple of questions above, so feel free to take a look when time permits.

CHANGELOG.md
f-f: No worries, this is fine 🙂

app/Spago.hs
f-f: How about `d`/`--dependency`?

src/Spago/Build.hs
f-f: Good question. In Haskell the catchall exception is `SomeException`, and we usually just catch that around the project (and I'd also catch that here)

The reason why you don't want to catch it everywhere is that sometimes you're better off just failing on Async exceptions (which are caught by `SomeException` too). I find async exceptions to be a very tricky part of Haskell to get right. You can read more about them [here](https://simonmar.github.io/posts/2017-01-24-asynchronous-exceptions.html) or [here](https://www.fpcomplete.com/blog/2018/04/async-exception-handling-haskell)

eahlberg: I see, thanks for the pointers!

f-f: You're welcome 🙂

eahlberg:
@f-f cool, thanks!

PR #283, opened 2019-06-18
Add --purs-command flag to customize calls to the compiler

### LINKED ISSUE #252: Usage with `purescript-psa`
As mentioned on slack, we'd like to be able to use [`psa`][]. [`psa`][] tends to be a direct swap for `purs`, so that's a common workflow when writing commands by hand. There's no explicit requirement that we have to set the binary name `purs` to something else, but that might be the simpler/more robust way to go about it.

Is this a thing `spago` can support? The implementation of that support I have no strong feelings about at the moment.

[`psa`]: https://github.com/natefaubion/purescript-psa

f-f:
This allows to customize the command used by Spago to call the compiler (defaults to `purs`), and so it should fix #252 assuming that the command specified is CLI-compatible with `purs`

However, it looks like `psa` is not CLI-compatible, so I'm slightly dubious on the usefulness of this. E.g. this is what happens if I run some command using `psa` as command:
```
$ spago -v -p psa install
Running `spago install`
Ensuring that the package set is frozen
Getting transitive deps
Running `fetchPackages`
Checking if `purs` is up to date
Running command: `psa --version`
spago: Oh noes! It looks like the PureScript version installed on your system is not compatible with the package-set you're using.

installed `purs` version:    0.7.3
minimum package-set version: 0.13.0

...
```

That is, `psa --version` behaves differently from `purs --version`, and this confuses `spago` (that is expecting `0.13.0` instead of `0.7.3` there)

cc @joneshf


joneshf:
Hmm. That's a shame. I guess what I ultimately want is for `purs` to output errors. It's quite the pain that it chooses not to do that and instead of forces that downstream to tools like `psa` and causes all of this complexity. None of this is really something you need to solve here.

If it can't happen, it can't happen. Thanks for trying, though!

f-f:
@joneshf makes sense!

I'll close this and #252 then

f-f:
Reopening as requested on Slack

vlatkoB:
Is there some way to add censor items to `spago`?

PR #285, opened 2019-06-19
Handle branch names differing only by case

hdgarrood:
### Description of the change

Escape upper case characters in branch names to ensure that branches
which only differ in case don't cause issues on case-insensitive file
systems.

Also switch to using an allowlist for escaping characters: it seems a
little safer to specify the characters which don't need escaping (and to
escape everything else), than to specify characters which do need
escaping.

Finally, add a test that `getCacheVersionDir` is injective.

### Checklist:

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -134,3 +134,7 @@ tests:
     - directory
     - temporary
     - extra
+    - containers
+    - QuickCheck
+    - spago
+    - extra
--- a/package.yaml
+++ b/package.yaml
@@ -138,4 +138,3 @@ tests:
     - containers
     - QuickCheck
     - spago
-    - extra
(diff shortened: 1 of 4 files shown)


hdgarrood:
It turns out that using QuickCheck to check that a function is injective requires a bit more effort than I thought it would. This is the simplest and most sensible thing I could come up with which reliably fails if I modify `getVersionCacheDir` to no longer be injective.

package.yaml
```diff
@@ -134,3 +134,7 @@ tests:
     - directory
     - temporary
     - extra
+    - containers
+    - QuickCheck
+    - spago
+    - extra
```
f-f:
```suggestion
```

hdgarrood: I don't think this can be removed, it's being used for `nubOrd`

f-f: It's already being imported in the line preceding your additions

hdgarrood: Oh of course, haha :man_facepalming:

Issue #256, opened 2019-06-07
Print warnings to stderr, not stdout


hdgarrood:
At the moment I'm using `spago sources` to build a list of source globs which I then pass to `purs` directly. This mostly works great, except for the fact that some warnings get printed to stdout, which confuses things:

```
$ spago sources 2>/dev/null
WARNING: wasn't able to check if your package set file is frozen
.spago/aff/v5.1.1/src/**/*.purs
.spago/aff-coroutines/v7.0.0/src/**/*.purs
.spago/arrays/v5.3.0/src/**/*.purs
.spago/avar/v3.0.0/src/**/*.purs
.spago/bifunctors/v4.0.0/src/**/*.purs
```

This leads to a bunch of warnings such as:

```
> set -o noglob && purs compile $(npm run --silent sources:purs | tr '\n' ' ')

purs compile: No files found using pattern: WARNING:
purs compile: No files found using pattern: wasn't
purs compile: No files found using pattern: able
purs compile: No files found using pattern: to
purs compile: No files found using pattern: check
purs compile: No files found using pattern: if
purs compile: No files found using pattern: your
purs compile: No files found using pattern: package
purs compile: No files found using pattern: set
purs compile: No files found using pattern: file
purs compile: No files found using pattern: is
purs compile: No files found using pattern: frozen
```

Incidentally, once #173 is a thing I can probably just use that and not need to worry about this.

f-f:
Good point. I think that since in some cases the output on `stdout` should be meaningful for machines too then the program should log _everything_ on `stderr`.

I think this is also the easiest thing to do, since it's just about having [`echo`](https://github.com/spacchetti/spago/blob/d5b6d4620b56c2dd8b64fa635eb4b340815ddfda/src/Spago/Prelude.hs#L128) print to `stderr`

hdgarrood:
I've realised one potential issue with doing this, which is that the compiler currently prints json errors to stderr, so if you run `spago build -- --json-errors` then you'll get both the json errors and spago logging sent to the same stream.

I suppose one option is to also provide a `--silent` flag, like `npm run --silent`, which suppresses all output from spago itself so that you only get output from subcommands. I guess the compiler ought to print json errors to stdout really too, but that might be a bit awkward to change at this point.

hdgarrood:
See also https://github.com/purescript/purescript/issues/3672

f-f:
[...]

PR #299, opened 2019-07-01
Add `--format` flag to `spago docs`

### LINKED ISSUE #294: Allow pass-through flags for `spago docs`
`purs docs` accepts a `--format` flag that can be used to generate Markdown documentation and Vim/Emacs tags as well as HTML (the default). Currently `spago docs` doesn't accept any arguments, so you always get HTML. Could we allow e.g. `spago docs -- --format ctags` in the same way that `spago build` accepts pass-through arguments?

If we think it's a good idea, I'm sure I could PR this!

elliotdavies:
### Description of the change

Add a flag to `spago docs` to support different documentation formats. Fixes #294.

Running `stack test` on `master` seems to lead to about half the tests failing, so I can't verify that these changes haven't broken them further.

### Checklist:

### DIFF
--- a/src/Spago/Purs.hs
+++ b/src/Spago/Purs.hs
@@ -54,11 +54,34 @@ bundle withMain (ModuleName moduleName) (TargetPath targetPath) = do
     ("Bundle failed.")
 
 
-docs :: Spago m => [SourcePath] -> m ()
-docs sourcePaths = do
+data DocsFormat
+  = Html
+  | Markdown
+  | Ctags
+  | Etags
+
+parseDocsFormat :: Text -> Maybe DocsFormat
+parseDocsFormat = \case
+  "html"     -> Just Html
+  "markdown" -> Just Markdown
+  "ctags"    -> Just Ctags
+  "etags"    -> Just Etags
+  _          -> Nothing
+
+printDocsFormat :: DocsFormat -> Text
+printDocsFormat = \case
+  Html     -> "html"
+  Markdown -> "markdown"
+  Ctags    -> "ctags"
+  Etags    -> "etags"
+
+
+docs :: Spago m => Maybe DocsFormat -> [SourcePath] -> m ()
+docs format sourcePaths = do
   let
     paths = Text.intercalate " " $ Messages.surroundQuote <$> map unSourcePath sourcePaths
-    cmd = "purs docs " <> paths <> " --format html"
+    formatStr = printDocsFormat $ fromMaybe Html format
+    cmd = "purs docs " <> paths <> " --format " <> formatStr
   runWithOutput cmd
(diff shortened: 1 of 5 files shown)


elliotdavies:
@f-f Awesome, I just need to do a bit more work to specify an output file when you choose `markdown | ctags | etags` - otherwise by default the compiler just dumps to stdout. (You only get generated files when choosing `html`.)

elliotdavies:
@f-f Actually, a question - in PureScript 0.13 the compiler will output HTML and MD to a `generated-docs` directory and ctags/etags to `tags` or `TAGS` files by default. In 0.12 it outputs HTML to `generated-docs` but any other option gets dumped to stdout.

What do you think is best for Spago to support? The defaults in 0.13 are fine, I think, but for anyone using 0.12 the stdout dump isn't going to be very useful (you can't even do `spago docs --format ctags > tags` because the tags file will end up including Spago's own logging). I could implement an `--output` option but `purs` doesn't support that in 0.12 so we'd have to implement it ourselves. Or we can just leave all the behaviour as it is.

f-f:
@elliotdavies great point. Since we're still in 0.x of both Spago and Purs let's not worry too much about old Purs versions (i.e. it's fine unless we break them, which doesn't seem the case). And once we fix #256 we'll have a `--silent` flag to totally silence Spago's own output, so we should be good anyways.

So let's go with 0.13 defaults

elliotdavies:
@f-f OK! In that case I think this is ready to review :smile:

src/Spago/Purs.hs
```diff
@@ -54,13 +54,36 @@ bundle withMain (ModuleName moduleName) (TargetPath targetPath) = do
...
+docs :: Spago m => Maybe DocsFormat -> [SourcePath] -> m ()
+docs format sourcePaths = do
   let
     paths = Text.intercalate " " $ Messages.surroundQuote <$> map unSourcePath sourcePaths
-    cmd = "purs docs " <> paths <> " --format html"
+    formatStr = printDocsFormat $ fromMaybe Html format
+    cmd = "purs docs " <> paths <> " --format " <> formatStr
   runWithOutput cmd
-    ("Docs generated. Index is at " <> Messages.surroundQuote "./generated-docs/index.html")
+    "Docs generation succeeded."
```
f-f: I'd like to keep informing the user on where the files are generated, so maybe here we could print different things depending on the `DocsFormat` value?

elliotdavies: Ah, in 0.13 the compiler prints the location of the files for you. For example:

- `Documentation written to: generated-docs/html`
- `Documentation written to: tags`

Then we get our output underneath.

f-f: Oh that's great!

f-f:
Thank you @elliotdavies! 🙂

PR #301, opened 2019-07-04
Import local packages `as Location`

### LINKED ISSUE #244: Local dependency paths are incorrect when `packages.dhall` is in another directory
I'm rolling out Spago in a monorepo where we have multiple packages, each with their own `spago.dhall`, and a single `packages.dhall` so that all the dependencies stay in sync.

The folder structure looks like:

```
- packages.dhall
- package-a
  - spago.dhall
- package-b
  - spago.dhall
```

The `packages.dhall` contains something like:

```
let additions = {
  package-a =
    mkPackage
      ./package-a/spago.dhall
      "./package-a"
      "v1.0.0"

, package-b =
    mkPackage
      ./package-b/spago.dhall
      "./package-b"
      "v1.0.0"
}
```

And then `package-a` lists `package-b` as a dependency in its `spago.dhall`.

`package-b`'s path is listed as `"./package-b"` and this is read by Spago and passed straight to the compiler (with `src/**/*.purs` appended). However, because the `packages.dhall` has been moved up one level this doesn't resolve: the actual path to `b` from `a` should be `../package-b`.

I put together an SSCCE here that you can run: https://github.com/elliotdavies/spago-dependencies-example

My proposed solution would be to modify Spago so that it checks the location of `packages.dhall` and generates the correct relative path. I'm happy to do that work if you think it's a good idea!

f-f:
This PR fixes #244, by changing how local packages are specified in the `packages.dhall` (yes this is a breaking change, because the old local syntax will now error out)

Taking the README example, we go from this:
```dhall
let additions =
  { foobar =
      mkPackage
        (../foobar/spago.dhall).dependencies
        "../foobar"
        "local-fix-whatever"
  }
```

to this:
```dhall
let additions = 
  { foobar = ../foobar/spago.dhall as Location }
```

This means a couple of things:
- the issue from #244 is fixed because Dhall will adjust the relative path properly
- but since it won't resolve the file, it's now possible to have _a single shared_ `packages.dhall`, that will contain e.g. all the packages in your repo (this wasn't possible before as you'd have circular imports in your Dhall file)

Things missing:
- [x] handle all the various errors with nice error messages
- [x] wait for a new `dhall-lang` release, happening in https://github.com/dhall-lang/dhall-lang/pull/635
- [x] wait for a new `dhall-haskell` release, happening in https://github.com/dhall-lang/dhall-haskell/pull/1156
- [x] wait for #289 to be merged
- [x] add tests for building with local packages
- [x] add tests for the `list-packages` command
- [x] update the README
- [x] add a proper "monorepo" section
- [x] verify that #309 works properly
- [x] check for references of `mkPackage` (see #322)


elliotdavies:
@f-f Thanks for this, I'll try it with our monorepo. Excited that it also properly supports a single `packages.dhall` - we've been factoring out `spago-deps.dhall` files to avoid the circular imports and it's a pain!

elliotdavies:
@f-f Still hitting circular import errors on this branch, I'm afraid. Our setup is:

```
- lib
  - purescript
    - packages.dhall
    - lib1
      - spago.dhall
    - lib2
      - spago.dhall
    - lib3
      - spago.dhall
app1
  - spago.dhall
app2
  - spago.dhall
```

`packages.dhall` is our shared single packages file and imports the libs as

```
let additions =
  { lib1 = ./lib1/spago.dhall as Location
  , lib2 = ./lib2/spago.dhall as Location
  , lib3 = ./lib3/spago.dhall as Location
  }
```

Each lib has a pretty trivial `spago.dhall` that includes `packages = ../packages.dhall` and similarly `app1/spago.dhall` includes `packages = ../lib/purescript/packages.dhall`.

When I `spago install` inside `app1` I get

```
spago: The following packages have circular dependencies:
  - lib1
  - lib2
  - lib3
```

f-f:
@elliotdavies wow great catch! I can reproduce the issue locally, and it's not due to circular dependencies per se, but it's because the imports in `lib1/spago.dhall` are resolved relatively to the folder you're calling Spago from.
So e.g. if you are in `app1` and you have a `deps.dhall` both in `lib1` and in `app1`, you'll get the latter one, which most likely contains `lib1` as a dependency, causing the circular deps problem.

So the root cause is the kinda-hacky implementation (but I cannot think of many other possible ones, so we'll keep this one) of local packages resolution here:

https://github.com/spacchetti/spago/blob/cfa28d7142cc45a7415dde7f1faaa68b51e3c3c7/src/Spago/Config.hs#L76-86

What's happening here is that we:
- pattern match on a `as Location` import in line 76
- read the raw AST (i.e. not resolved yet, so it contains imports and everything) in line 81
- transform the AST so that it goes from a config like `{ packages = .., dependencies = .. }` to just whatever the content of `dependencies` is, in line 85
- then resolve the expression in line 86 (relative to our current directory)

So if the expression is pure no problem, if it contains relative imports then they are not going to be correct.

Two possible solutions here:
- the quick one is that we just forbid imports inside `dependencies` for local packages
- the right one (but the one with more potential corner cases) is to just set the relative path to the imported package

f-f:
Right, I recalled there was a nice way to plug this in, so it was an easier fix than I expected. 
b12ea08 fixes the issue for me locally, @elliotdavies could you try it?

elliotdavies:
@f-f Yep, that's done it! (After a brief moment of panic when Spago just said "aaa" - eventually I realised I'd hit [this TODO error case](https://github.com/spacchetti/spago/pull/301/files#diff-47b561d66d3fdc8da263081ea4048920R80) 😂) Glad you were able to find a nice fix.

I might have some time Sunday or Monday if you'd like a hand ticking off some of the TODO items for this?

f-f:
@elliotdavies thanks, some help would be lovely! There are in fact some independent changes that should be merged in `master` before this goes in, and that you could pick up:
- adding tests with output fixtures for `spago list-packages` and `spago list-packages --json` (since I got very close to breaking them in this PR, so I'd like to have tests merged before to ensure I'm not breaking their behaviour)
- adding tests for `spago build` involving local packages (we don't have any) and your current monorepo setup (it'd be a nice demonstration of this feature to go from having to split out the dependencies in separate files to just having them in the `spago.dhall`)

f-f:
cc @Dretch: since in this branch I'm shuffling code for the `Config` (mainly going from `Either` to `IO` as the base monad to make things less wrapped, since we error out immediately anyways) in bf71f813955dfe5435556f2f30c3baa992d7e35a I'm also absorbing the `PublishConfig` into the `Config` to have less parsing code

Dretch:
> cc @Dretch: since in this branch I'm shuffling code for the `Config` parsing (mainly going from `Either` to `IO` as the base monad to make things less wrapped, since we error out immediately anyways) in [bf71f81](https://github.com/spacchetti/spago/commit/bf71f813955dfe5435556f2f30c3baa992d7e35a) I'm also absorbing the `PublishConfig` into the `Config` to have less parsing code

It all looks sensible as far as I can see. :+1:

PR #302, opened 2019-07-05
Add `--no-config-format` flag. Fixes #300.

### LINKED ISSUE #300: Is it possible to turn off autoformatting?
It looks like every time `spago` runs, it reformats the Dhall file. Is there a way to tell `spago` not to format the files when it runs?

aniketd:
### Description of the change

- Add `dontFormatConfig` to `GlobalOptions`
- Add condition to `Spago.Dhall.writeRawExpr` with `Bool` parameter

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,7 @@ Bugfixes:
 New features:
 - Add support for starting a repl within a folder which has not been setup as a spago project (#168)
 - Add `--format` flag to `spago docs` (#294)
-- Add `--no-config-format` global flag. Fixes #300. Skips formatting
-  `spago.dhall` during operations.
+- Add `--no-config-format` global flag to skip formatting Dhall files during operations. (#300, #302)
 
 ## [0.8.5] - 2019-06-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ Bugfixes:
 New features:
 - Add support for starting a repl within a folder which has not been setup as a spago project (#168)
 - Add `--format` flag to `spago docs` (#294)
+- Add `--no-config-format` global flag. Fixes #300. Skips formatting
+  `spago.dhall` during operations.
 
 ## [0.8.5] - 2019-06-18
 
--- a/src/Spago/Dhall.hs
+++ b/src/Spago/Dhall.hs
@@ -50,14 +50,15 @@ readRawExpr pathText = do
     else (pure Nothing)
 
 
-writeRawExpr :: Text -> (Text, DhallExpr Dhall.Import) -> IO ()
-writeRawExpr pathText (header, expr) = do
+writeRawExpr :: Bool -> Text -> (Text, DhallExpr Dhall.Import) -> IO ()
+writeRawExpr hasDontFormatConfig pathText (header, expr) = do
   -- After modifying the expression, we have to check if it still typechecks
   -- if it doesn't we don't write to file.
   resolvedExpr <- Dhall.Import.load expr
-  throws (Dhall.TypeCheck.typeOf resolvedExpr)
-  writeTextFile (pathFromText pathText) $ prettyWithHeader header expr <> "\n"
-  format pathText
+  _ <- throws (Dhall.TypeCheck.typeOf resolvedExpr)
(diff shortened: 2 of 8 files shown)


src/Spago/Dhall.hs
```diff
@@ -50,14 +50,15 @@ readRawExpr pathText = do
...
+writeRawExpr :: Bool -> Text -> (Text, DhallExpr Dhall.Import) -> IO ()
+writeRawExpr hasDontFormatConfig pathText (header, expr) = do
   -- After modifying the expression, we have to check if it still typechecks
   -- if it doesn't we don't write to file.
   resolvedExpr <- Dhall.Import.load expr
-  throws (Dhall.TypeCheck.typeOf resolvedExpr)
-  writeTextFile (pathFromText pathText) $ prettyWithHeader header expr <> "\n"
-  format pathText
+  _ <- throws (Dhall.TypeCheck.typeOf resolvedExpr)
+  unless hasDontFormatConfig $ do
```
f-f: Since we only want to disable the formatting (note that we're skipping also the write here) we should move this condition inside the `format` function

src/Spago/Dhall.hs
```diff
@@ -50,14 +50,15 @@ readRawExpr pathText = do
     else (pure Nothing)
 
 
-writeRawExpr :: Text -> (Text, DhallExpr Dhall.Import) -> IO ()
-writeRawExpr pathText (header, expr) = do
+writeRawExpr :: Bool -> Text -> (Text, DhallExpr Dhall.Import) -> IO ()
```
f-f: Could we have a small datatype like `data DoFormat = DoFormat | NoFormat` instead of a `Bool`? (for reasons why, see [boolean blindness](https://github.com/quchen/articles/blob/master/algebraic-blindness.md))

You can see an example of how to read it from CLI [here](https://github.com/spacchetti/spago/blob/a77b869edcfddd592f45726400b816de8c1f070e/app/Spago.hs#L122)

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ Bugfixes:
 New features:
 - Add support for starting a repl within a folder which has not been setup as a spago project (#168)
 - Add `--format` flag to `spago docs` (#294)
+- Add `--no-config-format` global flag. Fixes #300. Skips formatting
```
f-f:
```suggestion
- Add `--no-config-format` global flag to skip formatting of the Dhall files during operations (#300, #302)
```

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ Bugfixes:
...
+  `spago.dhall` during operations.
```
f-f:
```suggestion
```

src/Spago/Dhall.hs
aniketd: I tried this, but `spago build -F` still ends up formatting `spago.dhall`.

src/Spago/Dhall.hs
aniketd: Thanks for the pointers. The article was an important read. :smile: . I have changed the implementation. :+1:

src/Spago/Dhall.hs
aniketd: Is there some other way to avoid this? Should I instead check if we can avoid writing the file as part of a `build`?

aniketd:
Thanks for the insightful review. :)  I have made another commit that addresses all the changes requested :+1:

src/Spago/Dhall.hs
f-f: Oh I see now! The formatting still happens because we automatically migrate old versions of the config to the newest versions [here](https://github.com/spacchetti/spago/blob/75a3407275aa46f73b4f0e5d90049eb5dd42289b/src/Spago/Config.hs#L111): we read in the AST, apply a transformation (`addSourcePaths` in this case) and write the file back. I guess we could check if the config is at the latest version and only migrate if it's not.
Though there are going to be some more changes around that piece of code (with #301) so let's merge this PR and I'll fix this detail while rebasing #301

aniketd: Oh :thinking: alright! Thanks @f-f . I will keep an eye on #301 to know how you did it.

PR #304, opened 2019-07-08
Add tests for the list-packages command

f-f:
To avoid regressions as I risked in https://github.com/spacchetti/spago/pull/301 (see https://github.com/spacchetti/spago/pull/301#issuecomment-508927141 for more context)


test/SpagoSpec.hs
```diff
@@ -214,3 +214,19 @@ spec = around_ setup $ do
...
       spago ["bundle-module", "--to", "bundle-module.js", "--no-build"] >>= shouldBeSuccess
       checkFixture "bundle-module.js"
+
+  describe "spago list-packages" $ do
+
+    it "Spago should list-packages successfully" $ do
+
+      spago ["init"] >>= shouldBeSuccess
+      mv "packages.dhall" "packages-old.dhall"
+      writeTextFile "packages.dhall" "https://raw.githubusercontent.com/purescript/package-sets/psc-0.13.0-20190626/src/packages.dhall sha256:9905f07c9c3bd62fb3205e2108515811a89d55cff24f4341652f61ddacfcf148"
```
Dretch: Ah, this makes sense. I did some over-complicated stuff instead in `BumpVersionSpec.hs` https://github.com/spacchetti/spago/blob/3be573a0a44c1b23ef245cd48a33caf0fc24b978/test/BumpVersionSpec.hs#L27 ... I'll send a pull request unless someone gets there first.

f-f:
Interesting, the `spago list-packages` command seems to hang when run inside the tests on both macOS and Windows (I can reproduce on my mac)

PR #305, opened 2019-07-09
Use `psa` for compiling if installed

f-f:
### Description of the change

This detects if `psa` is installed in the system and if so automatically uses it to compile.

Also adds a global `--no-psa` flag to disable this.

Closes #283 (which this supersedes)

### Checklist:


i-am-the-slime:
Does it work with a local psa in `node_modules`?

stkb:
@i-am-the-slime No, nor does it for `purs`, and it's something that probably should be added.

PR #309, opened 2019-07-10
Watch all sources, including dependencies, when building

### LINKED ISSUE #172: Watch directory dependencies?
Back at it again with the "workspace" problems...

If I have spago project `A` that depends on spago project `B` (in a sibling directory), it would be nice if `spago build --watch` run from project `A` responded to changes in project `B`. This may not be possible, but just a suggestion :)

See https://github.com/jmackie/spago-issues for a reproducible example of what I'd like to do (that link will probably rot, apologies)

f-f:
Fix #172


f-f:
While debugging this with @jmackie we figured that it works for most paths, except for the ones that have this shape:
```
"/some/folder/../lib/purescript-stuff/src/**/*.purs"
```

And I suspect the Glob library does not properly match files in this case, so we should canonicalize the path. So let's hold this until #301 goes in, and then see if this gets fixed, otherwise we'll have to `canonicalizePath` when calling `getGlobs`

f-f:
I'll merge this now since we need to get this in anyways, and I added a checkbox in #301

PR #311, opened 2019-07-10
Fix local dependency tests

elliotdavies:
Change local dependency tests to use new `as Location` syntax.

### DIFF
--- a/test/SpagoSpec.hs
+++ b/test/SpagoSpec.hs
@@ -122,12 +122,12 @@ spec = around_ setup $ do
 
     it "Spago should install successfully when there are local dependencies sharing the same packages.dhall" $ do
 
-      -- Create local 'lib-a' package
+      -- Create local 'lib-a' package that depends on lib-c
       mkdir "lib-a"
       cd "lib-a"
       spago ["init"] >>= shouldBeSuccess
       rm "spago.dhall"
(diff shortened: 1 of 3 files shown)


test/SpagoSpec.hs
```diff
@@ -123,22 +122,29 @@ spec = around_ setup $ do
...
-      -- Create local 'lib-a' and 'lib-b' packages
-      for_ ["lib-a", "lib-b"] $ \name -> do
-        mkdir $ fromText name
-        cd $ fromText name
-        spago ["init"] >>= shouldBeSuccess
-        rm "spago.dhall"
-        writeTextFile "spago.dhall" $ "{ name = \"" <> name <> "\", dependencies = ./spago-deps.dhall, packages = ../packages.dhall }"
-        writeTextFile "spago-deps.dhall" "[\"console\", \"effect\", \"prelude\"]"
-        cd ".."
+      -- Create local 'lib-a' package
```
f-f: Could we add a `lib-c` that only `lib-a` depends on? (i.e. the `app` wouldn't depend on it, this is to test that local transitive dependencies are still included in the build)

elliotdavies: @f-f Absolutely!

f-f:
Wonderful, thank you! 👏

PR #324, opened 2019-07-15
Generate GIT-REPO#version versions in bower.json for packages that are not found in Bower.

Dretch:
This is a follow-on from #303.

### Checklist:


src/Spago/Bower.hs
```diff
@@ -93,7 +93,7 @@ mkPackageName spagoName = do
 mkBowerVersion :: Spago m => Bower.PackageName -> Text -> Text -> m Bower.VersionRange
 mkBowerVersion packageName version repo = do
 
-  let cmd = "bower info --json '" <> Bower.runPackageName packageName <> "#" <> version <> "'"
+  let cmd = "bower info --json " <> Bower.runPackageName packageName <> "#" <> version
```
Dretch: Thanks @f-f 

The ticks were to try and reduce the chances of versions (particular branch/tag names) with spaces in causing problems.

However, just adding ticks is not a very secure solution. I will see if `procStrictWithErr` will work instead.

f-f: Would it work to use double quotes?

Dretch: It would work I guess and probably by fine... but I think constructing shell strings like this tends to result in bugs (e.g. what if a branch name has a quote in it) and security holes.

I guess I could finish this issue using `shell` and double quotes and then look for a better solution afterwards.

src/Spago/Bower.hs
```diff
@@ -31,6 +33,12 @@ bowerPath :: IsString t => t
 bowerPath = "bower.json"
 
 
+bowerCmd :: Text
+bowerCmd = case System.buildOS of
+  Windows -> "bower." -- workaround windows issue: https://github.com/haskell/process/issues/140
```
f-f: Wow this is cool 😄

Dretch: But still broke :face_with_head_bandage: I'll try and find a Windows VM.

src/Spago/Bower.hs
```diff
@@ -87,23 +100,34 @@ mkPackageName spagoName = do
...
+    else pure $ Bower.VersionRange $ repo <> "#" <> version
+
+
 mkDependencies :: Spago m => Config -> m [(Bower.PackageName, Bower.VersionRange)]
 mkDependencies config = do
-
-  reposMeta <- GlobalCache.getMetadata Nothing
   deps <- Packages.getDirectDeps config
-
   for deps $ \(PackageName{..}, Package{..}) -> do
```
f-f: I was wondering if we should do this in parallel, as from what I've seen `bower info` can be quite slow. We could use [an async pool like we do here](https://github.com/spacchetti/spago/blob/3eaba105a0a9ff59a35709e5d6336c1d81d73215/src/Spago/FetchPackage.hs#L67), or just [`mapTasks`](http://hackage.haskell.org/package/async-pool-0.9.0.2/docs/Control-Concurrent-Async-Pool.html#v:mapTasks) as we don't need the fancy error handling

f-f: (this is not a requirement for merging, just noting that it might a bottleneck on larger packages)

Dretch: Yeah - done. Bower info is pretty slow (it takes a couple of seconds per call here, but I do have very high latency internet).

PR #325, opened 2019-07-15
Catch exceptions whilst trying to fetch metadata

carlosdagos:
### Description of the change

The previous way was only handling the case where the metadata fetched
could not be correctly parsed. Now it will also disable the global
cache if the HTTP request fails for any reason.

_NOTE_

I don't know if I'm on to the right thing here. In my build environment
the metadata URL isn't reachable due to firewalls (and I can't change that
in the short term, unfortunately). Maybe this isn't the right change and
I should open an issue to track all issues related to my use case 😄 	

Thanks for `spago`! I'm really enjoying using it 👍  

### Checklist:

### DIFF
--- a/src/Spago/GlobalCache.hs
+++ b/src/Spago/GlobalCache.hs
@@ -101,7 +101,8 @@ getMetadata cacheFlag = do
         Just a  -> a
 
       downloadMeta = handleAny
-        (const $ do
+        (\err -> do
+            echoDebug $ "Metadata fetch failed with exception: " <> tshow err
             echo "WARNING: Unable to download GitHub metadata, global cache will be disabled"
             pure mempty)
         (do
--- a/src/Spago/GlobalCache.hs
+++ b/src/Spago/GlobalCache.hs
@@ -100,16 +100,20 @@ getMetadata cacheFlag = do
         Nothing -> mempty
         Just a  -> a
 
-      downloadMeta = do
-        metaBS <- (Http.httpBS metaURL >>= pure . Http.getResponseBody)
-        case decodeStrict' metaBS of
-          Nothing -> do
+      downloadMeta = handleAny
+        (const $ do
             echo "WARNING: Unable to download GitHub metadata, global cache will be disabled"
-            pure mempty
-          Just meta -> do
-            assertDirectory globalCacheDir
-            liftIO $ BS.writeFile globalPathToMeta metaBS
-            pure meta
+            pure mempty)
+        (do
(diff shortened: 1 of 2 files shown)


src/Spago/GlobalCache.hs
```diff
@@ -99,16 +99,20 @@ getMetadata cacheFlag = do
...
+        (const $ do
             echo "WARNING: Unable to download GitHub metadata, global cache will be disabled"
-            pure mempty
-          Just meta -> do
-            assertDirectory globalCacheDir
-            liftIO $ BS.writeFile globalPathToMeta metaBS
-            pure meta
+            pure mempty)
+        (do
+            metaBS <- Http.getResponseBody `fmap` Http.httpBS metaURL
```
f-f: If you'd like to be able to customize the location of the metadata (because e.g. you have a proxy to GitHub in your network, or you could download it locally, etc) we could try to read it from environment, e.g. something like:

```suggestion
            metaBS <- Dhall.input Dhall.lazyText ("env:SPAGO_METADATA_JSON ? " <> metaURL <> " as Text")
```

This won't compile because types mismatch but just to give you an idea. This Dhall string:
- tries to read the `SPAGO_METADATA_JSON` variable from environment
- if that's missing, it tries to import the default URL `as Text`
- imports the result as a lazy `Text`, so then we can parse it

When setting the `SPAGO_METADATA_JSON` variable, you could put there something like `https://some.proxy/metadata.json as Text` or `./local/metadata.json as Text` or even `"{ \"some\": \"escaped json\" }"`

src/Spago/GlobalCache.hs
```diff
@@ -100,16 +100,20 @@ getMetadata cacheFlag = do
         Nothing -> mempty
         Just a  -> a
 
-      downloadMeta = do
-        metaBS <- (Http.httpBS metaURL >>= pure . Http.getResponseBody)
-        case decodeStrict' metaBS of
-          Nothing -> do
+      downloadMeta = handleAny
+        (const $ do
```
f-f: Could we `echoDebug` the exception here? In this way it would only show if one uses `--verbose`

carlosdagos: Done 😄

src/Spago/GlobalCache.hs
carlosdagos: I really like this idea. Will experiment with it a bit 👍

f-f:
@carlosdagos I'll merge this before it gets too out of sync with `master`, feel free to open another PR to add support for overriding the metadata URL 🙂 

Thank you!

PR #331, opened 2019-07-24
Add `--deps-only` flag to build only with deps and no project sources

justinwoo:
### Description of the change

Fix https://github.com/spacchetti/spago/issues/330

Had to fix this because I got tired of git add .; rm -rf src test; sp build; git co .

### DIFF
--- a/src/Spago/Build.hs
+++ b/src/Spago/Build.hs
@@ -9,8 +9,8 @@ module Spago.Build
   , Watch (..)
   , NoBuild (..)
   , NoInstall (..)
-  , DepsOnly (..)
   , BuildOptions (..)
+  , Packages.DepsOnly (..)
   , Purs.ExtraArg (..)
   , Purs.ModuleName (..)
   , Purs.SourcePath (..)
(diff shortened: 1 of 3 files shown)


src/Spago/Build.hs
```diff
@@ -43,6 +43,14 @@ data NoBuild = NoBuild | DoBuild
 -- | Flag to skip the automatic installation of libraries on build
 data NoInstall = NoInstall | DoInstall
 
+-- | Only build deps and ignore input paths
+data DepsOnly = DepsOnly | WithAll
+
+applyDepsOnly :: DepsOnly -> [Purs.SourcePath] -> [Purs.SourcePath]
+applyDepsOnly depsOnly sources = case depsOnly of
```
f-f: Since in every place we call `Packages.getGlobs` we also attach the project sources, lets:
- take `depsOnly` and the `Config` as parameters in `getGlobs`
- do this check there instead of doing this check and concatenating the project paths all over the place

justinwoo:
Someone else can take this branch and work on this

justinwoo:
we need to invest 40 hours into an automatic merge bot

f-f:
Oh no, we just added Mergify to `dhall-haskell` and it works great. I'll copy the config over here

f-f:
@justinwoo so don't merge this one as I'll use it to test mergify

PR #333, opened 2019-07-24
Add docs search

### LINKED ISSUE #89: Add searchbox to generated documentation
We should be able to generate documentation for all the libraries in the package set (and/or just the libraries used in the project), in a similar way to what `stack haddock` does for Haskell projects.

The use-cases for this are multiple and useful:
- get specialized documentation for your project without having to rely on Pursuit (because you might have dependencies not from the package set and/or on Pursuit)
- more generally, allow an easy way to have an offline Pursuit
- generate a "static Pursuit" for every package set in Spacchetti and host that alongside the package set, which is effectively the [Stackage](https://www.stackage.org/) approach.  
  Note that this would obsolete #27, as it would entirely remove the "where do I upload my library" problem → you would just add it to a Spacchetti package-set, and have it "published" in the next Spacchetti release

How to achieve this:
- it is possible to generate html docs for the whole package-set by running a `purs docs` command.  
  Justin had a good start with [acme-spago](https://github.com/justinwoo/acme-spago). He included a spago project in there, but if this is integrated in spago itself, we wouldn't need it (as we would just generate that info on the fly)
- then we'd need a way to search in the existing docs. Since the whole thing needs to be static files (i.e. no backend), we should have a frontend-side search). A simple way to do this would be to generate markdown docs and then index them with something like [Lunr](https://lunrjs.com/) (@justinwoo suggested scraping the externs JSON files, but I'm not sure what that would entail, so probably worth detailing more)

klntsky:
The binary size is now 48MB.

Some more features to consider:

- Add a CLI flag that disables docs patching when running `spago docs`.
- Add ability to skip compilation when running `spago search`.
- Pin JS files by hash

Fix #89

### DIFF
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -103,10 +103,11 @@ data Command
   -- | Deletes the .psc-package folder
   | PscPackageClean
 
+  -- | Runs `purescript-docs-search search`.
+  | Search
 
   -- | Show version
   | Version
-
 
   -- | Bundle the project into an executable (replaced by BundleApp)
   | Bundle
@@ -179,6 +180,7 @@ parser = do
       , bundleApp
       , bundleModule
       , docs
+      , search
       ]
 
     initProject =
@@ -229,6 +231,11 @@ parser = do
       , Docs <$> docsFormat <*> sourcePaths <*> depsOnly
       )
 
+    search =
+      ( "search"
+      , "Run the search engine."
+      , pure Search
+      )
 
     packageSetCommands = CLI.subcommandGroup "Package set commands:"
       [ install
@@ -377,5 +384,6 @@ main = do
       PscPackageLocalSetup force            -> liftIO $ PscPackage.localSetup force
       PscPackageInsDhall                    -> liftIO $ PscPackage.insDhall
       PscPackageClean                       -> liftIO $ PscPackage.clean
+      Search                                -> Spago.Build.search
       Bundle                                -> die Messages.bundleCommandRenamed
       MakeModule                            -> die Messages.makeModuleCommandRenamed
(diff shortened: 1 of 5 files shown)


app/Spago.hs
```diff
@@ -229,6 +231,11 @@ parser = do
       , Docs <$> docsFormat <*> sourcePaths <*> depsOnly
       )
 
+    search =
+      ( "search"
+      , "Run the search engine."
```
f-f:
```suggestion
      , "Start a search REPL to find definitions matching names and types"
```

src/Spago/Templates.hs
```diff
@@ -25,3 +25,15 @@ gitignore = $(embedFileUtf8 "templates/gitignore")
 
 bowerJson :: B.ByteString
 bowerJson = $(embedFile "templates/bower.json")
+
+docsSearchApp :: T.Text
+docsSearchApp =
+  $(embedURLWithFallback
+    "https://github.com/spacchetti/purescript-docs-search/releases/download/v0.0.4/docs-search-app.js"
```
f-f: Let's factor out the tag in a top level definition, as:
- we might want to use it (e.g. in `spago version`)
- it's easier to upgrade the templates

f-f:
@klntsky I'll merge now so we can include this in today's release. Thank you for doing this! 🙂

src/Spago/Templates.hs
klntsky: > ‘docsSearchVersion’ is used in a top-level splice, quasi-quote, or annotation,
>        and must be imported, not defined locally

To which module should I put it?

f-f: @klntsky how about `Prelude`?

googleson78:
This is awesome! Thanks!

PR #338, opened 2019-07-26
Refactor dry-run machinery so we print a clear list of dry actions

f-f:
Also make the `--jobs` flag global instead of passing it around everywhere

New output:
```
$ spago bump-version minor
Generating bower.json using the package set versions..
Generated bower.json using the package set
No git version tags found, so assuming current version is v0.0.0
Running `bower install` so `pulp publish` can read resolved versions from it

WARNING: this is a dry run, so these side effects were not performed:
- create new git tag "v0.1.0"

Use the `--no-dry-run` flag to run them
```

### DIFF
--- a/src/Spago/Bower.hs
+++ b/src/Spago/Bower.hs
@@ -22,7 +22,6 @@ import qualified Web.Bower.PackageMeta      as Bower
 
 import           Spago.Config               (Config (..), PublishConfig (..))
 import qualified Spago.Config               as Config
-import           Spago.DryRun               (DryRun (..))
 import qualified Spago.Git                  as Git
 import qualified Spago.Packages             as Packages
 import           Spago.PackageSet           (PackageName (..), Package (..), Repo (..))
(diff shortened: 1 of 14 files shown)


src/Spago/Bower.hs
```diff
@@ -65,23 +65,18 @@ writeBowerJson limitJobs dryRun = do
...
   when ignored $ do
-    die $ bowerPath <> " is being ignored by git - change this before continuing."
+    die $ bowerPath <> " is being ignored by git - change this before continuing"
 
-  case dryRun of
-    DryRun -> echo $ "Skipped writing " <> bowerPath <> " because this is a dry run."
-    NoDryRun -> do
-      liftIO $ ByteString.writeFile bowerPath bowerJson
-      echo $ "Generated " <> bowerPath <> " using the package set."
+  liftIO $ ByteString.writeFile bowerPath bowerJson
```
Dretch: No strong feelings, but maybe writing `bower.json` should also be a `DryAction` ?

Maybe not because it is easy to undo, but then so is creating a Git tag (unless it has been pushed elsewhere).

src/Spago/Version.hs
```diff
@@ -70,37 +70,32 @@ getNextVersion spec version@SemVer{..} =
...
+  Git.commitAndTag newVersionTag $ oldVersionTag <> " → " <> newVersionTag
+  echo $ "Git tag created for new version: " <> newVersionTag
 
 
 -- | Bump and tag a new version in preparation for release.
-bumpVersion :: Spago m => Maybe Int -> DryRun -> VersionBump -> m ()
-bumpVersion limitJobs dryRun spec = do
-  DryRun.showHelp dryRun
-
-  Git.requireCleanWorkingTree
```
Dretch: I'm not sure about removing `Git.requireCleanWorkingTree`, since (at least without other changes to compensate), it results in misleading output.

If I run (on this branch) `spago bump-version patch`, and I have some uncommitted changes, then Spago says `spago: A new bower.json has been generated. Please commit this and run bump-version again.` ... but actually no new `bower.json` was generated :stuck_out_tongue:

src/Spago/Bower.hs
f-f: Yeah I agree. My rationaly for doing this was: what if one wants to take a look at the generated file before creating a tag?

On the other hand also creating a tag is easy to undo, so I'll make everything dry and clarify the log message for the tag creation

PR #339, opened 2019-07-29
Stop reformatting the config file at every run

### LINKED ISSUE #300: Is it possible to turn off autoformatting?

f-f:
Fix #300 

This implements the fix described in https://github.com/spacchetti/spago/issues/300#issuecomment-515676453: we remove the `--no-config-format` flag and Spago will not reformat the config file unless it's needed, removing the need for the flag

cc @joneshf @liamgoodacre @aniketd: since there is some noise in the PR, the actual fix for this is basically [these lines](https://github.com/spacchetti/spago/pull/339/files#diff-47b561d66d3fdc8da263081ea4048920R250-R253)

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,17 +324,6 @@ $ spago run --node-args "arg1 arg2"
 ```
 
 
-### Avoid re-formatting the `spago.dhall` and `packages.dhall` with each command
-
-You can pass the `--no-config-format` or `-F` global flag:
-
-``` bash
-$ spago build -F
-Installation complete.
-Build succeeded.
-```
-
-
 ### Test my project
 
 You can also test your project with `spago`:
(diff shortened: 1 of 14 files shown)


f-f:
Also cc @aniketd: the fix for this is basically [these lines](https://github.com/spacchetti/spago/pull/339/files#diff-47b561d66d3fdc8da263081ea4048920R250-R253)

joneshf:
Thanks for keeping me in the loop!

f-f:
You're welcome 🙂

aniketd:
Thanks @f-f !

PR #345, opened 2019-07-29
Improve error message for overriding acceptable compiler version (#343)

### LINKED ISSUE #343: Improve error message for overriding acceptable compiler version
> - if you know what you're doing and you want to void this check, you can override the `version` of the `metadata` package in the packages.dhall

It is not clear that the user should put the actual compiler version to the version field of the package. I think the UX can be improved by giving a concrete example in the error message.

Consider the following:

> - if you know what you're doing and you want to void this check, you can override the `version` of the `metadata` package in the packages.dhall:
> let overrides = { metadata = upstream.metadata ⫽ { version = "..." } }

Where "..." stands for the actual version. Or maybe describing the process in plain English would be better.

Also, I think that some special version value like "any" could be used to silence the compiler version check completely (useful to simplify the process of checking the code against multiple compiler versions).

klntsky:
Fix #343


f-f:
Thanks!

Issue #355, opened 2019-08-02
First class support for alternate backends


andyarvanitis:
Has there been any discussion of being able to specify the options passed to `purs` in the config file? My particular use case (alternate backends) currently requires `spago build -- -g corefn`, which admittedly isn't bad, but it would be nice to just set it once for a project.

f-f:
Hi @andyarvanitis! There has been no discussion about this yet, but so far the trend is that if there are commonly used compiler flags we try to take care of the more general use-case at the spago level rather than just passing them through (because there are issues with the passthrough, e.g. see #353 or #216)

In this case I would be interested in investigating if we can have first-class support for alternate backends in spago, so e.g. in order to get the whole alternative build running you'd only have to add something like `backend = "go"` to your `spago.dhall`
I don't know what's the build workflow for the various backends, what are your thoughts on this?

andyarvanitis:
That sounds fine to me, and would make it even easier for people to use the alternate backends. The Go backend would probably be a better candidate (than C++) to start with, since its tooling is simpler. I'm still making some adjustments, but I can let you know how it works and what the requirements would be.

f-f:
@andyarvanitis thanks, looking forward to it! I'll rename this issue to track support for alternate backends then 🙂

andyarvanitis:
Following up on my previous comment, I think I have the golang tooling tweaking done (for now, at least). In the end, the only thing you currently need to do a go-backend build (typical case) involves one more step than what I showed in this issue earlier, namely:
```sh
$ spago build -- -g corefn
$ psgo
```
If `spago` (via the project config file) can provide those `purs` flags and then execute `psgo`, the user would get a native build/executable. 

FYI, `psgo` has a command switch of `--run` which doesn't build an executable, but instead runs the results of the compilation immediately. Maybe something to think about supporting for backends in general. Same goes for `--no-build`, which only generates intermediate language (e.g. Go) source files.

Let me know what you think.

f-f:
Sounds great! A possible UX for all of this could be:
- adding an optional `backend` key to the `spago.dhall`, e.g. `backend = "go"`
- if that's set, `spago build` would call `spago build -- -g corefn && psgo`
- since `psgo` supports `--run` then we could make it so that `spago run` calls `spago build -- -g corefn && psgo --run`
[...]

Issue #216, opened 2019-05-27
bundle-* commands don't support the purs --output option


stkb:
If you want your output dir as anything other than `/output`, then using this is fine:
```
spago build -- --output outputDir
```

However this doesn't work...
```
spago bundle-app --to index.js -- --output outputDir
```
...because although it compiles the .purs files into `outputDir`, It then runs the `purs bundle` command with "output" hardcoded:
```
Running command: `purs bundle "output/*/*.js" -m Main --main Main -o index.js`
purs bundle: No input files.
```

I guess the solution would be either to detect the --output/-o args passed to the purs compiler, or possibly to add an --output option to spago itself? An entry in `spago.dhall` would also be nice, if you're going that way.

f-f:
@stkb great catch! 

And yeah I agree with your analysis, we could either detect the `--output` in the args being passed to purs, or if this is too tricky we can add the `--output` flag to all the relevant commands, so we could just use that for everything. (I guess the question in this case would be "but what happens when we pass the `--output` to `purs` but the user also adds it as an extra argument for purs?")

I think adding adding an entry to `spago.dhall` is also a good idea, but a proper handling of the flags is necessary either way (because e.g. you might want to override the config value with a flag)

stkb:
Good points. This kinda opened up a can of worms 😃, but after having thought through all the options, I think it's fairly simple.

Currently, args specified after `-- ` don't get added to the `purs bundle` command, only to `purs compile` I don't know if this was a conscious decision, but I think it should stay this way for simplicity (at least for now). 

There are however two `purs bundle` options not currently supported by spago (`--namespace` and `--source-maps`), but I think that's ok; you can say: `spago bundle-*` is just a convenience command. If you want all the options, you can do `spago build && purs bundle ...` instead.

If there's config option for the output dir in `spago.dhall`, I guess I don't really also need an `--output` arg for spago? (Since all it does is save a few characters over `-- --output`, which I can use if I really need to override the config).

So it's then just detecting use of `-- --output` in the bundle- commands too.

I guess I could have a go at this.

coot:
[zephyr](https://github.com/coot/zephyr/) reads from `output` writes to `dce-output`; it would be nice if:
```
spago -s --output dce-output
```
would read from the output directory to create the bundle.

f-f:
[...]

PR #366, opened 2019-08-10
Add an explicit purs args flag

### LINKED ISSUE #353: Idea: Separate `purs compile` flags somehow
What do you think about separating flags passed to `purs compile` explicitly in some way, for example with `--purs-compile-flags=`?

Right now `--help` says you should use `--` if you want to specify that something should go to `purs compile`, but what really confused me is that it is not **mandatory** to separate them with `--`.
Running `spago build src/X.purs src/Y.purs` works fine - which is not immediately obvious when looking at `spago build --help` (it has no mentions of files as arguments).

This was especially problematic when I tried `spago build --watch --path src/**/*.purs`, only to see how not all of my files were "being watched". In reality only the first file from the expanded glob is being watched, because `--path` accepts only one file - the rest are passed onto `purs compile`, allowing compilation to proceed without errors (or at least that's what I think is going on here).

googleson78:
### Description of the change

Add an explicit flag for arguments passed over to `purs` as per #353. Fixes #353 

I am not sure about the short flag name, but everything that immediately comes to mind is already taken.
### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+Breaking changes (!!!):
+- **Flags and arguments that you want to give to `purs` are now passed with `--purs-args`**
+
+  The previous behaviour in which all arguments that could not parse as `spago` arguments
+  were passed along to `purs` was sometimes confusing (e.g. when using `--path` and multiple
+  arguments).
 
 ## [0.9.0] - 2019-07-30
 
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ $ spago build --path 'another_source/**/*.purs'
 E.g. if you wish to output your files in some other place than `output/`, you can run
 
 ```bash
-$ spago build -- -o myOutput/
+$ spago build --purs-args '-o myOutput/'
 ```
 
 If you wish to automatically have your project rebuilt when making changes to source files
@@ -315,7 +315,7 @@ $ spago run
 $ spago run --main ModulePath.To.Main
 
 # And pass arguments through to `purs compile`
-$ spago run --main ModulePath.To.Main -- --verbose-errors
+$ spago run --main ModulePath.To.Main --purs-args '--verbose-errors'
 
 # Or pass arguments to node
 $ spago run --node-args "arg1 arg2"
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -165,9 +165,9 @@ parser = do
 
     packageName     = CLI.arg (Just . PackageName) "package" "Specify a package name. You can list them with `list-packages`"
     packageNames    = many $ CLI.arg (Just . PackageName) "package" "Package name to add as dependency"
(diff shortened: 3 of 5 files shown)


thomashoneyman:
Quick question -- does this work with an arbitrary string to pass to `purs`, like this?

```sh
spago build --purs-args '--verbose-errors -o outputDirectory'
```

f-f:
Thinking more about this maybe it'd be better if we:
- just remove the passthrough arguments
- absorb into spago the most used `purs compile` flags

This is because adding this flag wouldn't make it any easier for us to control what's being passed to purs, while controlling it would help us greatly in fixing #216 and #355 (because duplicate flags make `purs` fail with a bad error message)

So I'd propose that instead of adding the `--purs-args` flag we'd close this escape hatch and add the following flags to the `BuildOptions`:
- `-o,--output`
- `-v,--verbose-errors`
- `-g,--codegen`
- `--json-errors`

However, if we do the change I propose above then we'd totally cut away the `psa` flags, so we cannot really go for it..

Also this last item on the list made me realize that all of this is potentially a big breaking change for editor tooling, because instead of `spago build -- --json-errors` you'd have to instead call `spago build --json-errors` or `spago build --purs-args="--json-errors"` (cc @nwolverson, does this make sense?)

So in the end I think we should not change anything at all here, as there are more things relying on all of this than what we initially thought

nwolverson:
Editor tooling - from my POV that's just a config change, so it's a change anyone using that setup would have to make. Personal opinion, wouldn't use that as a deciding factor.

Would say that if you "close the hatch" it's kind of committing to supporting all compiler flags, though I guess that's not too bad if you have a higher release cadence

googleson78:
tl;dr - I think disabling passthrough is too restrictive and passing all non-parsed `spago` arguments to `purs` makes it too easy to shoot yourself and not get feedback on what you did wrong.

Long -
I think of passthrough arguments as a "patch" that eventually disappears when you can do at least 100% of what the compiler. But it is also highly idealistic to think you will ever get to that point, because of the ever-evolving nature of the compiler. Even if you were to support all the compiler flags you would end up having to invest a lot of time into implementing dummy option passing, that your build system is not interested in at all. So :-1: for removing it from me.

Adding those flags to `spago` will definitely help with issues like #216, because people will have less (no) reason to use `purs` directly for their functionality. And to me the passthrough is "have this gun, point it wherever you want (including your own feet)" - people can always pass something incompatible with `spago`. And that's why I think it should be explicit, and not an implicit thing.

(sorry for this reply being all over the place)

justinwoo:
I hate the `--` style args in any tool I use, so I really like passing the args explicitly :+1: But yes, I think there's no way all passthroughs should be removed, because this makes it so that using the tool in any non-cookie-cutter way becomes a total nuisance, to the point of needing to abandon the tool.

I'd rather support both modes of operation between `--purs-args` and `--` for a while though, and if it's not a problem, just keep them. Maybe some see this as the worst option though. I would also think that adding those aliases for the most common commands would be nice enough, since any additional usages should be supported via `--purs-args` or `--`.

googleson78:
Uh.. [not sure if this is my fault?](https://travis-ci.com/spacchetti/spago/jobs/225738683)

f-f:
@googleson78 no it's not, see #378

PR #369, opened 2019-08-11
Fix syntax error in template comment

jhrcek:
### Description of the change

The code example in packages.dhall template has syntax error (duplicated comma), so it can't be just copy-pasted.

### Checklist:


f-f:
Thanks @jhrcek, great catch!

CI is currently broken, I'll merge this as soon as I sort that out 🙂

Issue #346, opened 2019-07-30
Spago run --watch runs(/compiles) twice with vim


googleson78:
I'm not sure if this is a bug with `spago` or something stemming from my setup (I don't have the same issue with `stack --file-watch` though), so sorry in advance if this is a mistake on my part!

Repro steps:

0. `spago init`
1. `spago run --watch`
2. After the initial run, save `src/Main.purs` (even without modifying it). I use `vim` to do this step.

Expected behaviour:
Files are compiled and run once.

Observed behaviour:
Files are run (and possibly compiled?) twice.

Sample output after a save:
```
File changed, rebuilding: "/home/googleson78/git/spago-watch-double-compile-repro/src/Main.purs"
Compiling Main
Build succeeded.
🍝
Success! Waiting for next file change.
Type help for available commands. Press enter to force a rebuild.
Build succeeded.
🍝
Success! Waiting for next file change.
Type help for available commands. Press enter to force a rebuild.
```
`spago` version is `0.8.5.0`

klntsky:
My neovim does this too. But not `nano`.
I believe this is related: https://vi.stackexchange.com/questions/11629/set-backupcopy-yes-doesnt-still-writes-the-file-twice

googleson78:
Thanks!

Setting `backupcopy=yes` resolves this problem, however I am still left wondering what the difference between `spago` and `stack`s file change detection is.

f-f:
I literally copypasted the implementation from Stack's one, so I'm not sure. I'll take a look

f-f:
@googleson78 which OS are you on?

googleson78:
Linux (Debian testing)

f-f:
@googleson78 could you try the branch from #362?

If that doesn't work you could try tweaking the debounce time [here](https://github.com/spacchetti/spago/pull/362/files#diff-dc77e7cd40666b5c8208568a77cfb050R28)

klntsky:
May it be so that debounce settings do not apply if a file was deleted and re-created, just like Vim does?

klntsky:
I tried building from #362 and I still get all `File changed, triggering a build` messages duplicated when using nvim.

f-f:
> May it be so that debounce settings do not apply if a file was deleted and re-created, just like Vim does?

Oh right, now it makes sense. Then I cannot figure out why it's different from Stack's behaviour

klntsky:
Because the code is different (it diverged)

f-f:
[...]

PR #371, opened 2019-08-12
Implemented custom event debouncing for file watcher.

klntsky:


elliotdavies:
I also see the issue in #346 when using Vim, and this solves the problem for me :smile:

src/Spago/Watch.hs
```diff
@@ -47,24 +53,43 @@ fileWatchConf
...
-          liftIO $ atomically $ do
-            globs <- readTVar allGlobs
-            when (shouldRebuild globs)
+          timeNow <- liftIO getCurrentTime
+
+          rebuilding <- liftIO $ atomically $ do
+            globs                <- readTVar allGlobs
+            (lastTime, lastPath) <- readTVar lastEvent
+
+            let shouldRebuild =
```
f-f: I find these 4 lines of logic a bit convoluted to parse, but I feel we cannot get them any simpler. Could you add some comments here on top with the rules to decide if we should rebuild? (basically what the logic implements but in natural language)

src/Spago/Watch.hs
```diff
@@ -18,14 +18,15 @@ import qualified System.FSNotify        as Watch
...
+import           Data.Time.Clock        (NominalDiffTime, getCurrentTime, diffUTCTime)
 
 -- Should we clear the screen on rebuild?
 data ClearScreen = DoClear | NoClear
   deriving Eq
 
 watch :: Spago m => Set.Set Glob.Pattern -> ClearScreen -> m () -> m ()
 watch globs shouldClear action = do
-  let config = Watch.defaultConfig { Watch.confDebounce = Watch.Debounce 0.1 } -- in seconds
+  let config = Watch.defaultConfig
```
f-f: If we're skipping the `fsnotify` debouncing should we use `Watch.NoDebounce` in the config? Or are we adding our own debouncing on top of that?

PR #384, opened 2019-08-21
Simplify getGlobalCacheDir, change Windows default

### LINKED ISSUE #380: spago not working on windows
`spago init` worked fine but `spago build` and `spago install psci-support` are giving the following error:  
`spago.exe: Directory "\\" is not accessible. Permissions {readable = True, writable = False, executable = False, searchable = True}`  
I also tried to run cmd.exe as System Administrator but didn't help. 

purescirpt version: 0.13.2
spago version : 0.9.0.0
windows 8

stkb:
Simplifies greatly the getGlobalCacheDir function by using `Directory.getXdgDirectory`, which does most of the work for us.

On Linux/MacOS, behavior isn't changed: will search for `$XDG_CACHE_HOME`, then `$HOME`, and will error if neither are found.*

On Windows, the default (if no `xdg_cache_home` exists) is changed to `%LOCALAPPDATA%`. This is really the "correct" place for it on Windows (rather than `~/.cache`) to be a good citizen.

It no longer checks for/uses `$HOME` on Windows, but I'd argue that isn't necessary: if you want it the *nixy way, you can just set `$xdg_cache_home` instead (as I do). Should also fix #380. 

----
*Now something I didn't quite understand was the alternative3 case in the original code.

https://github.com/spacchetti/spago/blob/13cf90cddb1e3e2b7338169d98799b6bc6a54ffe/src/Spago/GlobalCache.hs#L149-L178

At the end of the first line you have `alternative₃ <|> err`, but alternative3 can't fail, so the error message will never be shown?

So I don't know which outcome is intended if neither `$xdg_cache_home` nor `$home` are found: use a `.spago-global-cache` folder (in the project folder?), or show an error message; so I just went with the error message for now, which I've also modified, since it no longer applies to Windows.

Will have to see if this works on the CI...

----

### Checklist:
(Will do after review)

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -20,5 +20,6 @@ extra-deps:
 - th-lift-0.8.0.1@sha256:cceb81b12c0580e02a7a3898b6d60cca5e1be080741f69ddde4f12210d8ba7ca,1960
 - th-lift-instances-0.1.13@sha256:2852e468511805cb25d9e3923c9e91647d008ab4a764ec0921e5e40ff8a8e874,2625
 - semver-range-0.2.8
+- directory-1.3.4.0@sha256:500019f04494324d1df16cf83eefeb3f809b2b20b32a32ccd755ee0439c18bfd
 nix:
   packages: [zlib]
--- a/stack.yaml
+++ b/stack.yaml
@@ -21,5 +21,6 @@ extra-deps:
 - th-lift-instances-0.1.13@sha256:2852e468511805cb25d9e3923c9e91647d008ab4a764ec0921e5e40ff8a8e874,2625
 - semver-range-0.2.8
 - directory-1.3.4.0@sha256:500019f04494324d1df16cf83eefeb3f809b2b20b32a32ccd755ee0439c18bfd
+- process-1.6.5.1@sha256:77a9afeb676357f67fe5cf1ad79aca0745fb6f7fb96b786d510af08f622643f6,2468
 nix:
   packages: [zlib]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Bugfixes:
 - Do not compile files twice when using `--watch` and Vim (#346)
 - fix Dhall syntax error in packages.dhall template
 - Use `git clone` instead of `git fetch` when fetching a package (#373)
+- Fixes Windows global cache location; now uses `LocalAppData` as default (#384, #380)
 
 ## [0.9.0] - 2019-07-30
 
--- a/package.yaml
+++ b/package.yaml
@@ -67,7 +67,7 @@ library:
   - fsnotify
   - Glob
   - stm
-  - directory
+  - directory >= 1.3.4.0
   - mtl
   - exceptions
   - unliftio
--- a/src/Spago/GlobalCache.hs
+++ b/src/Spago/GlobalCache.hs
@@ -9,11 +9,10 @@ import qualified Data.ByteString        as BS
 import qualified Data.Map.Strict        as Map
 import qualified Data.Text              as Text
 import qualified Network.HTTP.Simple    as Http
-import qualified System.Environment
 import qualified System.FilePath        as FilePath
 import qualified Turtle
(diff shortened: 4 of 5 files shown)


stkb:
Ok CI has failed; I will look into it...

f-f:
Thanks @stkb! Some context and a few comments on this:
- CI is failing on Linux and Windows because you removed the `echoDebug` line that is used in some fixtures (I'd say we should keep it as it helps with debugging), and on macOS because it fails to find a build plan (and this is puzzling)
- the existence of the error predates the existence of alternative3. Nowadays we have a decent error message in there, but the reasoning for adding alternative3 was that the error was bad and here we can instead just provide a "safe" fallback and never fail.
  The reason why a "project local global cache" is considered safe is that Spago already has another "project local cache", and if for some reason we cannot access it then we cannot really perform any work (so it's ok to fail)
  So here I'd like to just get rid of the error message and use the safe alternative.
- The reason why we're not using `Directory.getXdgDirectory` is that I copied this code from `dhall-haskell`.
  In there we had this same change in https://github.com/dhall-lang/dhall-haskell/pull/550, and then added back this code in https://github.com/dhall-lang/dhall-haskell/pull/621, and then just got rid of the `directory` dependency altogether in https://github.com/dhall-lang/dhall-haskell/pull/789 as it was causing issues.
  Here I'm fine with this change since we don't have to support old GHC versions

f-f:
Addendum: it's also interesting to see [how Pulp does this](https://github.com/purescript-contrib/pulp/blob/e09129b32a42e1275ac19f1225cf4d3d61084967/src/Pulp/Login.purs#L114): they search for `USERPROFILE` on Windows and `HOME` elsewhere

stkb:
Thanks @f-f for all the explanation. Have:
- Put back the echoDebug statement
- Made '.spago-global-cache` the fallback option
- Removed the now-unused error message.
- Added an entry to the changelog
- Rebased on latest commit

Linux & Windows is now passing, but MacOS still failing on the build plan. BTW, I'm not sure if the `stack.yaml.lock` file should have been updated given the new explicit dependency on `directory >= 1.3.4.0` ? (don't really know fully how Stack works 🤷‍♂)

Re: how Pulp does it: yeah `%UserProfile%` is the Windows `$HOME` equivalent and it will get you that, but the way we'll be doing it now (with LocalAppData) really is the correct way.

f-f:
@stkb I'll try your branch on my Mac, one sec

stack.yaml
```diff
@@ -20,5 +20,6 @@ extra-deps:
 - th-lift-0.8.0.1@sha256:cceb81b12c0580e02a7a3898b6d60cca5e1be080741f69ddde4f12210d8ba7ca,1960
 - th-lift-instances-0.1.13@sha256:2852e468511805cb25d9e3923c9e91647d008ab4a764ec0921e5e40ff8a8e874,2625
 - semver-range-0.2.8
+- directory-1.3.4.0@sha256:500019f04494324d1df16cf83eefeb3f809b2b20b32a32ccd755ee0439c18bfd
```
f-f: If you add this CI should pass on macOS too

```suggestion
- directory-1.3.4.0@sha256:500019f04494324d1df16cf83eefeb3f809b2b20b32a32ccd755ee0439c18bfd
- process-1.6.5.1@sha256:77a9afeb676357f67fe5cf1ad79aca0745fb6f7fb96b786d510af08f622643f6,2468
```

f-f: Update: I realised that this might not be enough, as Stack will update the `stack.yaml.lock` and we should commit that too

stkb:
@f-f The build succeeded with that change so I have merged 😃. I still don't know about `stack.yaml.lock`: Stack isn't recreating it for me from either just a `stack build` or `stack build --stack-yaml stack.yaml`

f-f:
@stkb thanks! And no worries, if CI says it's good then it's good 😄 (I'll update the lockfile in the next commits)

Stack started creating/updating the lockfiles from version 2 on, so I suspect you're using a pre-2.0 version

PR #387, opened 2019-08-26
Fix watch paths on Windows

stkb:
Fixes invalid path issue described in https://github.com/spacchetti/spago/issues/380#issuecomment-524583489

[`Watch.globToParent`](https://github.com/spacchetti/spago/blob/ba97e38bd5377b8ca316b4dd3dfe598c70708681/src/Spago/Watch.hs#L177-L192) had some issues on Windows:
- Use of `(== pathSeparator)` can cause problems since there are 2 different path separators.
- Splitting and re-combining the glob paths resulted in paths missing a slash after the drive letter, ie: `C:Path\To\Folder`. These are treated as relative paths (on the given drive), eventually resulting in invalid paths.

This commit does bring a slight change in behavior in that relative paths no longer have "./" prepended to them, but I don't think that's a problem? (a relative path is still a relative path, with or without the "./").

There aren't currently any tests for watching (that's kinda difficult I think?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Bugfixes:
 - Use `git clone` instead of `git fetch` when fetching a package (#373)
 - Fixes Windows global cache location; now uses `LocalAppData` as default (#384, #380)
 - Fix failure to copy to global cache on a different filesystem (#385)
+- Fix watch function on Windows (issue with paths) (#387, #380)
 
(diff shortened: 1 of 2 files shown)


stkb:
Agreed and I just noticed the `UnitSpec.hs` file so I might add some there. But I think this is good for now (if someone can manually check it still works on a non-Windows platform)

PR #396, opened 2019-08-29
Implement 'spago publish'

### LINKED ISSUE #204: Add `spago publish`
This is Step 2 in the publishing flow described in https://github.com/spacchetti/spago/issues/142#issuecomment-473684622

We'd add a `spago publish` to push a new version to both the Bower registry and `package-sets`. It would:
- make sure that some fields necessary for publishing are specified in the `spago.dhall`: `repo`, `license`, `version` (not sure about this last one, should we get it from the git tag?)
  This means that the people not publishing a library will not have to have these fields defined (i.e. we should have a `PublishingConfig` type on the Haskell side, and try to conform the config to that instead of the normal `Config` type)
- generate an ephemeral `bower.json` in a temp directory (maybe we need to copy the whole sources for it to work, as I don't seem to find an option to tell bower to use a config file from somewhere else than the current folder).
  The idea here is that generating a Bowerfile is reproducible from a `spago.dhall`, so you shouldn't need to include a `bower.json` in your repo.
  The versions of packages specified as dependencies can come from the package set or can probably be `*` (I think there are drawbacks to either choice: one is too restrictive, the other too permissive)
- call `git` and `bower` commands in the background to publish a git tag and publish on the bower registry (note: this would require `bower` to be in the path, if not we should tell people to install bower)
- print out a link to open an issue in package-sets about adding the package there (basically #72)

f-f:
Fix #204

This implements the `spago publish` command, finishing up the "publish pipeline" and making Spago available for libraries as well.

The implementation very much follows [Pulp's one](https://github.com/purescript-contrib/pulp/blob/master/src/Pulp/Publish.purs), first pushing the git tag to the remote, then calling `purs publish`, then uploading the docs to Pursuit.

The command runs in "dry run mode" by default, sample output:
```
Generating a new Bower config using the package set versions..
Generated a valid Bower config using the package set
Found current version from git tag: v1.2.3
Creating resolutions file from Bower configurations..

WARNING: this is a dry run, so these side effects were not performed:
* push the new tag to the remote repository
* register the package on the Bower registry if necessary
* upload the docs for this version on Pursuit

Use the `--no-dry-run` flag to run them
```

cc @Dretch @hdgarrood 

TODO:
- [x] tests


f-f:
While writing tests I stumbled on this error from `purs publish`:
```
spago: Error while running `purs publish`. Log:
There is a problem with your package, which meant that it could not be
published.
Details:
  Compile error:
    Error found:
    in module Type.Row
    at bower_components/purescript-typelevel-prelude/src/Type/Row.purs:4:5 - 4:24 (line 4, column 5 - line 4, column 24)

      Export for type Prim.RowList.Cons conflicts with type class Prim.Row.Cons


    See https://github.com/purescript/documentation/blob/master/errors/ExportConflict.md for more information,
    or to contribute content related to this error.
```

I'm using the latest `purs` (`0.13.3`) and package-sets (`psc-0.13.3-20190827`), so this should be all fine since the package set includes `typelevel-prelude@5.0.0`

However, in the test project I'm installing `tortellini`, that has `typelevel-prelude@4.0.0` in its Bower config, so the result of `bower install` (and the generated resolutions fed to `purs publish` use that one instead of the correct version (note: in this case `tortellini` works perfectly well with `5.0.0` but it's just pinned to `^4.0.0`)

I think we have a couple of ways here:
1. police all the packages in the set and enforce using right boundaries instead of pinning a single version. This is similar to what Hackage is doing, and it's a daunting amount of work I'd prefer to really not do
2. or skip the `bower install` here and instead use all the version info from the package set (since that gives us a working combination of versions with the current compiler, while the bower resolver doesn't)

I'll most likely go with option (2), but just putting this out here to see if it makes sense (cc @hdgarrood @justinwoo)

hdgarrood:
Have you considered leaving it as is and letting it fail? That seems to me like the right thing to do here. I don't think it's a good idea to ignore version bounds on other published packages. Ideally a package shouldn't make it into the package set at all if its version bounds aren't compatible with the rest of the set.

hdgarrood:
I think this is slightly different to your option 1 (unless I've misunderstood) - I'm suggesting that package sets should continue to pin a single version of each package, but also each package's Bower dependencies must have version ranges which allow whichever versions of all of their dependencies are in the set, if that isn't already the case.

f-f:
@hdgarrood 

My option 1 was quite generic ("police version bounds") so I think it fits what you describe here. I thought about letting this fail, but putting myself in the shoes of the publisher of the package I feel this is quite bad UX: the package works with the current package set I have versioned, I am able to cut a version of it, but I cannot publish it _now_ because some dependency upstream has wrong bounds, and instead I have to open a PR and wait for a week 😕 (this for N packages that have wrong bounds)

On the other hand having a package published with a set of dependencies while its upstreams use other versions in their bower config would prevent the Bower solver from actually building a plan, so option 2 is also not good.

An alternative would be to police package bounds before they enter `package-sets`,  so that CI would check if the package you're trying to add there includes the versions in the set in its bounds. This is the most viable, but as I mentioned above it's a lot of work (both for `package-sets` maintainers and for package publishers/submitters)

In the end I think Hackage's situation is the better one here, where they can edit package metadata (including bounds) without cutting new releases - but we cannot do it here in a git-based situation

hdgarrood:
It is indeed bad UX for the maintainer at that moment in time, but I think publishing a known-broken package is worse for the ecosystem overall; it's bad manners to put a package up on the bower registry and on Pursuit if you know that it's not installable with Bower. I don't think it's that bad to have to wait for a little while for upstream to accept changes, because you can always add the package to your overrides in the meantime.

> An alternative would be to police package bounds before they enter package-sets, so that CI would check if the package you're trying to add there includes the versions in the set in its bounds. This is the most viable, but as I mentioned above it's a lot of work (both for package-sets maintainers and for package publishers/submitters)

Yes, I definitely think this is something that should be happening, unless we want to give up with version bounds entirely (which as I've written elsewhere, I feel quite strongly that we should not do).

f-f:
@hdgarrood 

> It is indeed bad UX for the maintainer at that moment in time, but I think publishing a known-broken package is worse for the ecosystem overall; it's bad manners to put a package up on the bower registry and on Pursuit if you know that it's not installable with Bower

Agreed, that's why I said option 2 is also bad

> Yes, I definitely think this is something that should be happening, unless we want to give up with version bounds entirely (which as I've written elsewhere, I feel quite strongly that we should not do)

I think this would also be bad UX, so I would not like to have it in place because the threshold of contributing to package-sets is already quite high when compared with just publishing on Bower, and we certainly don't want to raise it more.

If we want to keep version bounds I think we should start thinking about a registry that instead of leaving package versions "inside" the package artifacts (as it is currently) instead keeps them in a centralized place (e.g. a git repo) where bounds can be easily adjusted, etc.

As I picture it "adding a package" would be:
- pushing a git tag as we do now
- opening a PR to the registry (this can be automated) where we can eventually adjust bounds for all the packages we need to adjust - and this reminds me of `nixpkgs`

This would require moving away from Bower entirely - though we could have a transition period where we mirror all packages from the Bower registry

f-f:
In any case, I think `spago publish` as it is right now is kind of unusable so I'm not sure if we should actually merge/release it 😕

klntsky:
> If we want to keep version bounds I think we should start thinking about a registry that instead of leaving package versions "inside" the package artifacts (as it is currently) instead keeps them in a centralized place (e.g. a git repo) where bounds can be easily adjusted, etc.

This is very similar to Hackage's "trustees".

https://hackage.haskell.org/packages/trustees/

f-f:
> This is very similar to Hackage's "trustees"

@klntsky yes! Except in Hackage the situation is more complicated since the actual bounds information is "out of band" and inside the Hackage registry, and the git repo there is just for openin issue for people to request changes.
Here I'm proposing we just use a git repo to contain all the bounds information for packages (i.e. we'd only care about the packages listed in packages manifests but not their versions. Note that this means that this registry can be compatible with `psc-package`, `bower` and `spago` already and with no changes), so that people could actually just PR changes to it

hdgarrood:
> This is fine for apps but not fine for libraries, because you cannot publish a library with overrides?

Yes, in this case you have to choose to either wait for your upstream library to update, fork it, or find a way to drop the dependency. The way I see it, this is just a fact of life for FOSS ecosystems really - part of being a package author is having to care about these things occasionally.

> I think this would also be bad UX, so I would not like to have it in place because the threshold of contributing to package-sets is already quite high when compared with just publishing on Bower, and we certainly don't want to raise it more.

We should expect the threshold to be higher here, surely, because of what a package set is?

I actually think that checking version bounds in package-sets would be a UX improvement, especially when you consider the ecosystem as a whole. As a package author, I want to know earlier if my version bounds are wrong. As a package consumer, I want my dependencies' bounds to be correct so that I'm not hosed if I need to make some tweaks to my upstream package set.

> If we want to keep version bounds I think we should start thinking about a registry that instead of leaving package versions "inside" the package artifacts (as it is currently) instead keeps them in a centralized place (e.g. a git repo) where bounds can be easily adjusted, etc.

I don't think there's much point in keeping a registry of version bounds in the absence of a tool which actually uses them to perform installations; what are the version bounds being used for in this scenario? Just for accepting packages into package sets?

f-f:
@hdgarrood I'll sum up where we are and try to place these ideas in the broader context of "package registries" that we often touch on

The problem we have here is that when you use Bower to manage dependencies then you always find a build plan, so publishing is fine.
But when you manage dependencies with a package set there's no checking that the versions are in the Bower bounds for the whole lifetime of the project (because that's kind of the point, otherwise you'd use a solver and find a plan, etc); but then when trying to publish on the Bower registry we actually have to find a plan and this is not a given (unless you adjust all the bounds, which means opening a certain amount of PRs and waiting for people to cut releases, which to me is unacceptably slow and cumbersome)

So the options seem to be:
1. do nothing, accept that publishing packages with Spago is painful
2. or set up another registry with centralised bounds handling and use eventually another tool to find build plans
3. or setup a check for Bower bounds on `package-sets` CI
4. or make package-sets disjunct from Bower (so that one could at least "publish" a package there)

...and I think they are all bad

I believe that one of the long-term goals is to migrate away from Bower though - or at least "decouple enough from it" - so "using another registry/tool" is not entirely off the table if we can find a decent transition story (i.e. the usual "breaking changes are totally fine if one can automate the migration")

Assuming we go with the "centralised bounds registry" idea (because it's desirable as described above), there is a way we can set it up and use it to minimise the pain involved in the process even if there's no tool using it directly. In this scenario we would:
- publish packages on Bower as usual with bounds and everything
- mirror all the info from the bower registry to this centralised-bounds-thing once a day
- have "bound overrides" in this central registry that we edit by hand or automatically based on the compilation results we find out in `package-sets` (note: these last two points are also what Stackage does)
- have @spacchettibotti lookup in this overrides file and proactively open PRs to the relevant packages to adjust the Bower bounds 

This would effectively be a mix of options (1) and (2) from above, and would allow us to:
- avoid adding more burden on people using `spago publish` (you'd still have to fix bounds for stuff outside of `package-sets`, but for that you're on your own anyways)
- avoid adding more burden on people adding packages to `package-sets`
- keep the Bower bounds up to date automatically, which is cool
- eventually write a tool that uses bounds information from this new registry, eventually use this info as part of an actual registry, etc

Dretch:
Thanks for doing this @f-f :+1: 

I think I understand your summary. All I can think to add is...

> 4. or make package-sets disjunct from Bower (so that one could at least "publish" a package there)

If I understand correctly, then this is currently more or less the status quo: if people use `pulp version` and `pulp publish` to publish their packages to Bower, and then separately send a pull request to the package set, then the dependency versions for that package in Bower will likely not be same as in the package set (and often the Bower bounds will be incorrect - as we have found here).

Given that this is already the status quo, then perhaps a short-term solution would be for `spago bump-version` and `spago publish` to behave more like Pulp, and give up on generating `bower.json` files (and also give up on calling `bower install`). Instead, Spago could just check that `bower.json` exists and that the version ranges for each dependency in it are compatible with the versions in the package set (but require the user to create/edit `bower.json`).

Doing this would mean at least that people could completely migrate away from Pulp and just use Spago for everything. People would still need to maintain `bower.json` files, though.

I was curious about how widespread the version number incompatibilites were, so I hacked a version of spago to show me (https://github.com/Dretch/spago/tree/verify-set-bower-hack). It is a pretty widespread phenomenon (on `psc-0.13.3-20190831`) - search for `INCOMPATIBLE` in 
[spago-verify-set-bower.txt](https://github.com/spacchetti/spago/files/3617902/spago-verify-set-bower.txt)

f-f:
@Dretch 

> If I understand correctly, then this is currently more or less the status quo: if people use pulp version and pulp publish to publish their packages to Bower, and then separately send a pull request to the package set, then the dependency versions for that package in Bower will likely not be same as in the package set (and often the Bower bounds will be incorrect - as we have found here)

What I described is not the status quo because what I meant above by "disjunct" is "allow packages in package-sets that are not on bower", which is currently a strict requirement to get a package in `package-sets`
The solution for the status quo that you describe is listed in point (3), and would place additional burden to people adding a package to the set

> Given that this is already the status quo, then perhaps a short-term solution would be for `spago bump-version` and `spago publish` to behave more like Pulp, and give up on generating `bower.json` files (and also give up on calling bower install). Instead, Spago could just check that `bower.json` exists and that the version ranges for each dependency in it are compatible with the versions in the package set (but require the user to create/edit bower.json).

I'm not too keen on this because it would be strictly more work for people to publish a package compared to what's the current implementation: they have to respect the boundaries in both cases, but at least now we generate the bowerfile for them even though it does not contain the correct versions

> Doing this would mean at least that people could completely migrate away from Pulp and just use Spago for everything. People would still need to maintain bower.json files, though.

Once we merge this people would be able to only use Spago for everything anyways. Though if I understand correctly your proposal would avoid depending on `bower` entirely

> I was curious about how widespread the version number incompatibilites were, so I hacked a version of spago to show me

Yeah in the end I'd say that we could ship what we have here plus a nice printout of which version boundaries of the dependencies would need to be updated in order to be able to publish a package.
Note that this is solution (1) from my above comment, which is "accept that publishing with Spago is going to be painful for now".
I'm not happy about this but it might be better than nothing because we can iterate on it.
The command is already marked as "experimental", and we could even hide it from the command list in the help message, so people don't actually use it en masse

f-f:
I'll close this one for now as it looks like there's no clear path forward to get a nice publishing flow with this strategy

flip111:
@f-f could you consider supporting `spago publish` for the cases when there are no version conflicts? Then version conflict handling can be added later, no?

f-f:
@flip111 the publishing situation is a bit weird right now since we are transitioning to a different package registry. One of my goals with the new registry is to make this version conflict less awkward and things easy for Spago - but things are still in flux so I cannot say more right now.

However, I'll note that basically all the code in this PR can be thrown away since it has diverged a lot and the publishing flow is going to change anyways

PR #400, opened 2019-09-02
Add --no-search flag to `spago docs`

klntsky:
Perhaps we'll want to be able to use a newer version of docs-search in `starsuit` without cutting a new spago release. Other users may find it useful too, if they don't want to pay their build time for a feature they may not need.


f-f:
@klntsky CI is failing because of my `spago login` patch from last week, I'll take a look..

PR #402, opened 2019-09-06
Look up remote imports dynamically when doing frozen check

### LINKED ISSUE #349: Locate packages.dhall based on spago.dhall
In `src/Spago/PackageSet.hs` the `ensureFrozen` function looks up the package set and logs a warning if it can't find it: https://github.com/spacchetti/spago/blob/2d349cc7c1ce02769688e136bdf9a3939c04ed35/src/Spago/PackageSet.hs#L302

At the moment the path for this is hard-coded to `packages.dhall` in the same directory: https://github.com/spacchetti/spago/blob/2d349cc7c1ce02769688e136bdf9a3939c04ed35/src/Spago/PackageSet.hs#L93

In a monorepo setup, where there is likely to be a single root `packages.dhall` but many distributed `spago.dhall` files, this means Spago commands almost always trigger this warning.

Given that the `spago.dhall` files must define a path to the `packages.dhall` anyway, would it make more sense to read and evaluate that path instead? This would give monorepo users the benefit of the check actually running, as well as reducing unnecessary warnings.

elliotdavies:
### Description of the change

To fix #349 we no longer look for `packages.dhall` directly when checking that the remote imports are frozen. Instead we look at `spago.dhall` and check all remote imports in its dependency graph.

### Checklist:


elliotdavies:
@f-f Submitting this as a draft to check I'm working along the same lines as you were thinking. If so I'll do the tests, changelog, etc.

The `_stack_` part of the `Status` didn't seem to contain much, but I was able to retrieve what we need from `_graph` instead.

package.yaml
```diff
@@ -85,6 +85,8 @@ library:
   - unordered-containers
   - retry
   - github
+  - transformers
+  - lens-family-core
```
f-f:
```suggestion
```

We are already importing `lens-family-core` in line 65

src/Spago/Dhall.hs
```diff
@@ -34,7 +36,7 @@ format pathText = liftIO $
 
 
 -- | Prettyprint a Dhall expression
-pretty :: Pretty.Pretty a => DhallExpr a -> Dhall.Text
+pretty :: Pretty.Pretty a => a -> Dhall.Text
```
f-f: Since this is now generic enough it could be moved to `Spago.Prelude`

src/Spago/Dhall.hs
```diff
@@ -46,6 +48,24 @@ prettyWithHeader header expr = do
   PrettyText.renderStrict $ Pretty.layoutSmart Pretty.defaultLayoutOptions doc
 
 
+-- | Return a list of all imports starting from a particular file
```
f-f: Brilliant! Yes this is what I was searching for, happy it can be done 🙂

src/Spago/PackageSet.hs
```diff
@@ -226,10 +238,9 @@ freeze = do
...
 ensureFrozen = do
   echoDebug "Ensuring that the package set is frozen"
-  rawPackageSet <- liftIO $ Dhall.readRawExpr path
-  case rawPackageSet of
-    Nothing -> echo "WARNING: wasn't able to check if your package set file is frozen"
-    Just (_header, expr) -> do
-      let areRemotesFrozen = foldMap isRemoteFrozen expr
-      unless (and areRemotesFrozen) $ do
-        freeze
+  imports <- liftIO $ Dhall.readImports "spago.dhall"
```
f-f: CI is failing because in `Spago.PscPackage` we also call `ensureFrozen`, and in there we only have a `packages.dhall` and not a `spago.dhall`

I suggest we add a `FilePath` parameter to `ensureFrozen`, so that we can pass `configPath` or `packagesPath` where needed

elliotdavies:
@f-f Ace - I'll try to get this done today in the hopes of making the v10 release :smile:

package.yaml
elliotdavies: Oops. I might alphabetise this list to make it a bit easier to scan...

elliotdavies:
@f-f Ready to re-review I think!

src/Spago/PackageSet.hs
```diff
@@ -202,16 +204,26 @@ isRemoteFrozen :: Dhall.Import -> [Bool]
...
   })             = [isJust hash]
 isRemoteFrozen _ = []
 
 
+localImportPath :: Dhall.Import -> Maybe System.IO.FilePath
+localImportPath (Dhall.Import
+  { importHashed = Dhall.ImportHashed
+    { importType = Dhall.Local Dhall.Here file
+    }
+  })              = Just $ Text.unpack $ "." <> pretty file
```
f-f: Did you mean this?

```suggestion
  })              = Just $ Text.unpack $ "." </> pretty file
```

If not, I'm not sure why we need to concatenate a `.` 🤔

src/Spago/Config.hs
```diff
@@ -151,7 +151,7 @@ ensureConfig = do
     die $ Messages.cannotFindConfig
   try parseConfig >>= \case
     Right config -> do
-      PackageSet.ensureFrozen
+      PackageSet.ensureFrozen defaultPath
```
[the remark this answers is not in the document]
elliotdavies: Oh quite right, that was silly of me

src/Spago/PackageSet.hs
elliotdavies: `pretty file` returns a filepath like `/../my-package/spago.dhall`, with a leading `/`. I guess we could drop that instead of prepending the `.`?

f-f: Right. How about calling `pretty` on the `importType`? [Its `Pretty` instance already does what we're doing here](https://github.com/dhall-lang/dhall-haskell/blob/640913f8d2cff6d80dfa836efa45ba3aa4220545/dhall/src/Dhall/Core.hs#L276)

elliotdavies: That'll do nicely!

elliotdavies:
@f-f Awesome! One other thing - at the moment we call `freeze` once per local import, so in a large monorepo setup you might see `Generating new hashes...` quite a few times. It's not a huge issue but the UX is a bit odd. Is it overkill to add a `shouldEcho` flag to `freeze`, or something?

f-f:
@elliotdavies great point, and I'm not sure how to solve it. We could try moving the print around (e.g. in `ensureFrozen` instead of `freeze`), but let's take that in another PR so we can see the changes separately from this, which I'll merge now. Thank you! 🙂

PR #420, opened 2019-09-21
Warn (but don't error) when trying to watch missing directories

### LINKED ISSUE #406: `-watch` commands error when source directories are missing
Say `spago.dhall` includes a `sources` configuration like the following, which is the default:

```
sources = [ "src/**/*.purs", "test/**/*.purs" ]
```

but you delete the `test` directory. Running `spago build` (for example) will log this warning but otherwise proceed to compilation:

```
```

However `spago build -w` will throw this error and go no further (the file watching still works but it never gets round to compiling):

```
getDirectoryContents:openDirStream: does not exist (No such file or directory)
Type help for available commands. Press enter to force a rebuild.
```

I'd guess the differing behaviour might come from here https://github.com/spacchetti/spago/blob/add6ec7ac9429d251c80f518052ebdcc9ca69759/src/Spago/Build.hs#L88-L90 but I haven't investigated properly yet.

elliotdavies:
### Description of the change

Fix #406 by checking that a glob matches some files before trying to watch them.

The varying combinations of `String`, `Text`, `FilePath` and `SourcePath` make the code a little hairy, but it seems to work 😅 

If you like we could actually do the `filterMatchingGlobsAndWarn` bit before the `BuildOnce`/`Watch` case statement, rather than relying on the compiler to warn the user.

### Checklist:

### DIFF
--- a/src/Spago/Build.hs
+++ b/src/Spago/Build.hs
@@ -87,7 +87,19 @@ build BuildOptions{..} maybePostBuild = do
   absoluteGlobs <- traverse makeAbsolute $ Text.unpack . Purs.unSourcePath <$> allGlobs
   case shouldWatch of
     BuildOnce -> buildAction
-    Watch     -> Watch.watch (Set.fromAscList $ fmap Glob.compile absoluteGlobs) shouldClear buildAction
+    Watch -> do
+      matches <- liftIO $ checkGlobsExist absoluteGlobs
+      Watch.watch (Set.fromAscList $ fmap Glob.compile matches) shouldClear buildAction
+
+  where
+    checkGlobsExist :: [String] -> IO ([String], [String])
+    checkGlobsExist globs = do
+    
+    checkGlobExists :: String -> IO Bool
+    checkGlobExists pattern = do
+      paths <- Glob.glob pattern
+      pure $ null paths
+
 
 -- | Start a repl
 repl
--- a/src/Spago/Build.hs
+++ b/src/Spago/Build.hs
@@ -25,7 +25,6 @@ import           Spago.Prelude
 import qualified Data.Set             as Set
 import qualified Data.Text            as Text
 import qualified System.FilePath.Glob as Glob
-import qualified System.IO            as System.IO
 import qualified System.IO.Temp       as Temp
 import qualified Turtle               as Turtle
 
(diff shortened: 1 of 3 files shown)


src/Spago/Build.hs
```diff
@@ -78,21 +80,36 @@ build BuildOptions{..} maybePostBuild = do
...
+  absoluteJSGlobs <- makeGlobsAbsolute allJsGlobs
 
   case shouldWatch of
-    BuildOnce -> buildAction
-    Watch     -> Watch.watch
-      (Set.fromAscList $ fmap Glob.compile $ absoluteGlobs <> absoluteJSGlobs)
-      shouldClear buildAction
+    BuildOnce -> buildAction allPsGlobs
+    Watch -> do
+      matches <- filterMatchingGlobsAndWarn allPsGlobs
```
f-f: Why do we match only on PS globs but not the JS ones?

f-f: Oh I see, I guess we do it only on the PS ones because we care about the directory being there, which we would detect anyways. I'd like to have a comment on top of this line explaining what I just figured out though 😄

elliotdavies: @f-f I only just pulled in the JS glob changes from `master` and was pondering this behaviour too 😅 I think it might be better to filter the JS globs too in case they end up differing for some reason?

f-f: It might make sense, though you'd end up with "duplicated" errors, right?

Alternatively we could group all the "non matching" globs and spit out a giant warning with all the non-matching paths after all the filtering

elliotdavies: @f-f So, while playing with warning for JS files I noticed some interesting behaviour.

`spago.dhall`:

```
sources = [ "src/**/*.purs", "test/**/*.purs", "fake/**/*.js" ]
```

Partial output from `spago build -w`:

```
WARNING: No matching files found when trying to watch pattern: test/**/*.purs
WARNING: No matching files found when trying to watch pattern: fake/**/*.js
WARNING: No matching files found when trying to watch pattern: ./../package-b/src/**/*.js
WARNING: No matching files found when trying to watch pattern: .spago/psci-support/v4.0.0/src/**/*.js
WARNING: No matching files found when trying to watch pattern: src/**/*.js
WARNING: No matching files found when trying to watch pattern: test/**/*.js
WARNING: No matching files found when trying to watch pattern: fake/**/*.js
```

A few observations:

1. There are separate warnings for `test/**/*.purs` and `test/**/*.js`; as you say, we could aggregate these
2. There's a warning for `src/**/*.js` even though there are `.purs` files in that directory, which feels like bad UX
3. `fake/**/*.js` shows up twice because we currently assume all the globs in `spago.dhall` are for `.purs` files
4. You get warnings for `package-b` (a local dependency from the monorepo) and `.spago/psci-support`. We probably shouldn't be watching anything in `.spago`, but either way I don't think it would be expected behaviour for these warnings to show up here

1 and 2 can be addressed by checking that the directory exists rather than the specific types of files inside, but 3 and 4 might need some further consideration.

f-f: @elliotdavies excellent analysis! 👏 

Interesting corner cases in here. For (1) and (2) we should indeed just convert globs to their parent folder and check that they exist. I think we can reuse [`globToParent`](https://github.com/spacchetti/spago/blob/46f0f7c3f0d3c630b5d03e9f6ef073caa847402a/src/Spago/Watch.hs#L178) for this (we do use it already in the watcher, which means that we might be able to consolidate the check from this PR inside the `setWatched` in there

To solve (3) I think we should also get the parent directory of the glob, and generate both the `.js` and `.purs` globs from there (we can fix this here or just log it in an issue and refactor the glob handling later)

Then (4) is a thing because of https://github.com/spacchetti/spago/issues/172, but you're right that we could avoid watching stuff in `.spago` and only watch local dependencies

elliotdavies: @f-f OK great, I've raised these as separate issues because they're not really related to this PR's intention:

- #431 for (3) and to think about globs in `spago.dhall` more generally
- #430 to avoid watching dependencies in `.spago` for efficiency and to mitigate (4)

For (1) and (2) I'd been thinking along the same lines so I'll go ahead and do that 😄

elliotdavies:
@f-f How's this looking now?

PR #426, opened 2019-09-25
enable support of alternate backend in building

csicar:
### Description of the change

- extend `Config` type with `Maybe Text` for the optional backend entry
- execute the `alternateBackend` in build, if it exists
- add Dhall utility for optional entry in a Map

### Checklist:

### DIFF
--- a/src/Spago/Build.hs
+++ b/src/Spago/Build.hs
@@ -25,6 +25,7 @@ import           Spago.Prelude
 
 import qualified Data.Set             as Set
 import qualified Data.Text            as Text
+import qualified Data.List            as List
 import qualified System.FilePath.Glob as Glob
 import qualified System.IO.Temp       as Temp
 import           System.Directory (getCurrentDirectory)
@@ -82,16 +83,16 @@ build BuildOptions{..} maybePostBuild = do
     DoInstall -> Fetch.fetchPackages cacheConfig deps packagesMinPursVersion
     NoInstall -> pure ()
   let allGlobs = Packages.getGlobs deps depsOnly configSourcePaths <> sourcePaths
+      genCoreFnOpts = [ Purs.ExtraArg "--codegen", Purs.ExtraArg "corefn" ]
       buildAction = do
         case alternateBackend of 
           Nothing ->
               Purs.compile allGlobs pursArgs
           Just backend -> do
-              Purs.compile allGlobs $
-                pursArgs ++
-                  [ Purs.ExtraArg "--codegen"
-                  , Purs.ExtraArg "corefn"
-                  ]
+              let normalizedArgs = concatMap (fmap Purs.ExtraArg . Text.words . Purs.unExtraArg) pursArgs
+              when (genCoreFnOpts `List.isInfixOf` normalizedArgs) $ 
+                die "No need to pass `--codegen corefn` explicitly when using the `backend` option. Remove the argument to solve the error."
+              Purs.compile allGlobs $ pursArgs ++ genCoreFnOpts
 
               shell backend empty >>= \case
                 ExitSuccess   -> pure ()
--- a/src/Spago/Build.hs
+++ b/src/Spago/Build.hs
@@ -83,16 +83,14 @@ build BuildOptions{..} maybePostBuild = do
     DoInstall -> Fetch.fetchPackages cacheConfig deps packagesMinPursVersion
     NoInstall -> pure ()
   let allGlobs = Packages.getGlobs deps depsOnly configSourcePaths <> sourcePaths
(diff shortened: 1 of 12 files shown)


src/Spago/Build.hs
```diff
@@ -83,10 +83,22 @@ build BuildOptions{..} maybePostBuild = do
...
-        Purs.compile allGlobs pursArgs
-        case maybePostBuild of
-          Just action -> action
-          Nothing     -> pure ()
+        case alternateBackend of 
+          Nothing ->
+              Purs.compile allGlobs pursArgs
+          Just backend -> do
+              Purs.compile allGlobs $
+                pursArgs ++
```
f-f: `purs` will just panic if you pass it the same argument twice, so here we should check if the user is already passing `--codegen`, and then do something about it (options: either override what the user is doing or just fail and explain the situation)

src/Spago/Build.hs
```diff
@@ -83,10 +83,22 @@ build BuildOptions{..} maybePostBuild = do
...
+                  [ Purs.ExtraArg "--codegen"
+                  , Purs.ExtraArg "corefn"
+                  ]
+
+              shell backend empty >>= \case
+                ExitSuccess   -> pure ()
+                ExitFailure n -> die $ "Backend exited with error:" <> repr n
```
f-f:
```suggestion
                ExitFailure n -> die $ "Backend " <> surroundQuote backend <> " exited with error:" <> repr n
```

test/fixtures/spago-configWithBackend.dhall
```diff
@@ -0,0 +1,15 @@
+{-
+Welcome to a Spago project!
+You can edit this file as you like.
+-}
+{ backend =
+    "echo hi from backend> alternate-backend-output.txt"
```
f-f: Oh this is clever 😄

csicar: I would have like to also check, that no arguments are passed to the backend command, but `$@` is not supported on windows and causes the ci to fail

src/Spago/Build.hs
csicar: I think failing is preferable, that way the user knows, that he can save himself the effort of passing the flag

src/Spago/Build.hs
```diff
@@ -82,11 +83,23 @@ build BuildOptions{..} maybePostBuild = do
...
-        Purs.compile allGlobs pursArgs
-        case maybePostBuild of
-          Just action -> action
-          Nothing     -> pure ()
+        case alternateBackend of 
+          Nothing ->
+              Purs.compile allGlobs pursArgs
+          Just backend -> do
+              let normalizedArgs = concatMap (fmap Purs.ExtraArg . Text.words . Purs.unExtraArg) pursArgs
+              when (genCoreFnOpts `List.isInfixOf` normalizedArgs) $
```
f-f: `purs` would still fail if the user passes e.g. `--codegen docs`, so let's check just for `--codegen` being in the array of `pursArgs`

csicar: good call

f-f:
@csicar thanks for doing this! 💯

csicar:
Thanks for merging :)

I really like the code style of the spago code-base. Are you following a specific style-guide? 

---

I've started work on the `--run` option (seems easy enough). But I think it does not make sense to implement it completely before the CLI spec is defined. 

E.g. for running the `test` command you would need to choose the `Test.Main` module as the entry point, which should probably be supported by the CLI spec

f-f:
@csicar 

> But I think it does not make sense to implement it completely before the CLI spec is defined

Agreed. I think we should close #355 (as the initial request was implemented here), and open two more issues: one for discussing/tracking the CLI spec for alternate backends, and the other one for the `--run` option

> I really like the code style of the spago code-base. Are you following a specific style-guide?

Thanks 🙂 I don't follow a specific styleguide (consciously at least), and by now I forgot all the things I read and incorporated in my style, but I can say that I am heavily influenced by [Gabriel Gonzales'](https://github.com/Gabriel439) style, which I find very pleasant to read.
So a good place to look could be the [`dhall-haskell`](https://github.com/dhall-lang/dhall-haskell) codebase.

In general I just try to write "super boring code", so that it's accessible to people that are not familiar with certain operators, so it's easy to refactor when coming back to it after months, etc.

Some articles I find useful and usually recommend:
- http://www.haskellforall.com/2015/09/how-to-make-your-haskell-code-more.html
- https://chrisdone.com/posts/german-naming-convention/

I also run [stylish-haskell](https://github.com/spacchetti/spago/blob/master/.stylish-haskell.yaml) on every save, to keep imports tidy

csicar:
thanks for the insight :)

f-f:
You're welcome! 🙂

PR #428, opened 2019-09-25
Add support for --no-comments to spago init

jhrcek:
### Description of the change

Here's the initial attempt to implement support for `--no-comments` switch (fix https://github.com/spacchetti/spago/issues/417)

The most future-proof approach seems to be to parse the dhall template file, remove comments from AST and pretty-print it before writing the file. This however is currently not possible due to dhall stripping most comments while parsing (see https://github.com/dhall-lang/dhall-haskell/issues/145).

This PR could be considered a hack taking advantage of this Dhall behavior and simply does parse + pretty print of dhall templates, which "accidentally" strips comments. This impl. will potentially break in the future when Dhall parser preserves comments, but I added tests to catch this.

I thought of couple of alternative approaches, but I find them even more hacky/brittle than the current PR:
1. Do something like `Text.dropWhile (/="-}")` on the dhall template files (would break once more than 1 comment is present in the template or comment is somewhere else than in the beginning)
2. Have another "no comment" templates and copy those instead (would require maintenance of duplicate templates)

### Checklist:

### DIFF
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -34,8 +34,8 @@ data Command
 
   -- | ### Commands for working with Spago projects
   --
-  -- | Initialize a new project
-  = Init Bool
+  -- | Initialize a new project (switches to 1. force overwrite and 2. strip comments)
+  = Init Bool Bool
 
   -- | Install (download) dependencies defined in spago.dhall
   | Install (Maybe CacheFlag) [PackageName]
@@ -128,6 +128,7 @@ parser = do
 
     force   = CLI.switch "force" 'f' "Overwrite any project found in the current directory"
     verbose = CLI.switch "verbose" 'v' "Enable additional debug logging, e.g. printing `purs` commands"
+    noComments = CLI.switch "no-comments" 'C' "Strip comments in the generated package.dhall and spago.dhall files"
 
     -- Note: the first constructor is the default when the flag is not provided
     watch       = bool BuildOnce Watch <$> CLI.switch "watch" 'w' "Watch for changes in local files and automatically rebuild"
@@ -174,7 +175,7 @@ parser = do
     initProject =
       ( "init"
       , "Initialize a new sample project, or migrate a psc-package one"
-      , Init <$> force
+      , Init <$> force <*> noComments
       )
 
     build =
@@ -332,7 +333,7 @@ main = do
   (command, globalOptions) <- CLI.options "Spago - manage your PureScript projects" parser
   (flip runReaderT) globalOptions $
     case command of
-      Init force                            -> Spago.Packages.initProject force
+      Init force noComments                 -> Spago.Packages.initProject force noComments
       Install cacheConfig packageNames      -> Spago.Packages.install cacheConfig packageNames
       ListPackages packagesFilter jsonFlag  -> Spago.Packages.listPackages packagesFilter jsonFlag
       Sources                               -> Spago.Packages.sources
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -12,10 +12,11 @@ import qualified Turtle              as CLI
 
 import           Spago.Build         (BuildOptions (..), DepsOnly (..), ExtraArg (..),
                                       ModuleName (..), NoBuild (..), NoInstall (..), NoSearch (..),
-                                      SourcePath (..), TargetPath (..), Watch (..), WithMain (..),
-                                      OpenDocs (..))
+                                      OpenDocs (..), SourcePath (..), TargetPath (..), Watch (..),
+                                      WithMain (..))
 import qualified Spago.Build
 import qualified Spago.Config        as Config
+import           Spago.Dhall         (TemplateComments (..))
 import           Spago.DryRun        (DryRun (..))
 import qualified Spago.GitHub
 import           Spago.GlobalCache   (CacheFlag (..))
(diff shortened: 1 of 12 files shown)


src/Spago/Build.hs
```diff
@@ -131,7 +131,7 @@ repl cacheFlag newPackages sourcePaths pursArgs depsOnly = do
       Temp.withTempDirectory cacheDir "spago-repl-tmp" $ \dir -> do
         Turtle.cd (Turtle.decodeString dir)
 
-        Packages.initProject False
+        Packages.initProject False False
```
jhrcek: Not sure if this default is ok here..

src/Spago/Dhall.hs
```diff
@@ -41,6 +41,17 @@ prettyWithHeader header expr = do
   let doc = Pretty.pretty header <> Pretty.pretty expr
   PrettyText.renderStrict $ Pretty.layoutSmart Pretty.defaultLayoutOptions doc
 
+stripComments :: String -> Text -> Text
+stripComments file dhallSrc =
+  case Parser.exprFromText file dhallSrc of
+    -- TODO parsing of template could potentially fail (e.g. with update of dhall version)
+    -- what to do in that case? Print a warning and return not-stripped template?
+    Left _ -> dhallSrc
```
jhrcek: Is this approach acceptable, or should I print some warning in Spago monad?

app/Spago.hs
```diff
@@ -149,6 +149,7 @@ parser = do
 
     force   = CLI.switch "force" 'f' "Overwrite any project found in the current directory"
     verbose = CLI.switch "verbose" 'v' "Enable additional debug logging, e.g. printing `purs` commands"
+    noComments = CLI.switch "no-comments" 'C' "Strip comments in the generated package.dhall and spago.dhall files"
```
jhrcek: It's kinda hard to come up with single letter switch name when most of the letters are taken :-) Do you have better suggestions than 'C'?

test/fixtures/packages-no-comments.dhall
```diff
@@ -0,0 +1,8 @@
+let upstream =
+      https://github.com/purescript/package-sets/releases/download/psc-0.13.3-20190920/packages.dhall sha256:53873cf2fc4a343a41f335ee47c1706ecf755ac7c5a336e8eb03ad23165dfd28
```
jhrcek: Having this fixture will add maintenance burden when bumping package set versions. Suggestions how to avoid it are welcome :-)

src/Spago/Build.hs
```diff
@@ -84,11 +84,11 @@ build BuildOptions{..} maybePostBuild = do
     NoInstall -> pure ()
   let allGlobs = Packages.getGlobs deps depsOnly configSourcePaths <> sourcePaths
       buildAction = do
-        case alternateBackend of 
+        case alternateBackend of
```
jhrcek: Hope you don't mind these whitespace changes, they're done by stylish-haskell. But I can undo them if you prefer.

test/fixtures/packages-no-comments.dhall
jhrcek: A simple alternative that could be used instead of having this fixture could be something like this code directly in the test:
```haskell
config <- readTextFile "spago.dhall"
config `shouldNotSatisfy` (Text.isInfixOf "{-") -- comments are not present
config `shouldSatisfy` (Text.isInfixOf "dependencies =") -- but some dhall stuff is present
```

src/Spago/Build.hs
f-f: No worries, whitespace changes are fine!

src/Spago/Build.hs
f-f: It is, but since we're starting to add booleans in here let's make some datatypes instead to avoid confusion (it's ok if you don't refactor the rest and only add a `data TemplateComments = WithComments | NoComments`)

test/fixtures/packages-no-comments.dhall
f-f: Yep, this last idea sounds great!

app/Spago.hs
f-f: I think `C` is great!

src/Spago/Dhall.hs
f-f: This approach is what I would implement myself, great by me 🙂

src/Spago/Build.hs
jhrcek: Absolutely, I agree. I didn't like another Boolean flag from the beginning (boolean blindness and stuff y'know), but wasn't sure where to put that data type. I added it to `Spago.Dhall` in my next commit. This is I think the most logical place for it (together with comment stripping functionality).

test/fixtures/packages-no-comments.dhall
jhrcek: Ok, implemented.

jhrcek:
Thanks for the feedback! spago works great and is really useful. Also the source code is nicely readable. I'd like to contribute some more to practice my Haskell :-) Let me know if there are some issues that you'd consider helpful to fix, otherwise I'll pick something with "good first issue" label

src/Spago/Build.hs
f-f: This works well, thanks! I think we'll eventually want to move all types to the `Spago.Types` module so that they are not scattered around the codebase, but on the other hand it's also OK to have them where they provide context

f-f:
@jhrcek thanks for the kind words, I'm happy to hear this 😊

I just went through all the issues marked with "good first issue", and they should all be fine to pick up! I also tried to provide context and/or an idea for implementation in there, but if there are any details missing or confusing I'd be happy to detail more in the relevant threads 🙂

PR #437, opened 2019-10-03
Build static binaries on Linux

### LINKED ISSUE #427: glibc2.27 introduced as dependency in 0.10.0
you can choose to do nothing about this, but it might be good to do something about it.

## What

glibc2.27 is now a runtime dependency of 0.10.0, whereas the previous versions did not do anything requiring this.

## Why

good question. some infection happened somewhere.

## How

Try using `xenial`, e.g. on Travis, and install `spago` via npm.

```
> spago install
spago: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.27' not found (required by spago)
```

Seems expected: https://packages.ubuntu.com/source/xenial/glibc

## Non-solution

* "static compilation"
* "use easy-purescript-nix"

## Potential solutions

* find culprit and purge
* build matrix, release for multiple glibc versions instead of only on bionic (https://github.com/spacchetti/spago/blob/760faf6d428e431e377d74ec77cb5dceef37eb81/.travis.yml#L4)
* advisory about non-usability on xenial (16.04LTS) via npm (upgrade to 18.04LTS, "patch ELF", "use easy-purescript-nix")

f-f:
Fix #427 

TODO:

- [x] Add `static` flag to `package.yaml`
- [x] Use the flag to also build statically on Travis
- [x] Update changelog
- [x] Remove from README mentions of bugs related to dynamic libs


f-f:
cc @justinwoo

justinwoo:
Might as well try it

hdgarrood:
Was just thinking about this problem (the purescript compiler binaries have also suffered from this). Have you encountered any problems with this approach yet? We might want to use the same approach in the compiler. I have read that glibc doesn't like to be statically linked, but I don't really understand why yet.

hdgarrood:
Ah - I tried using the static linux binary for 0.11.1.0 and I get segmentation faults when I try to do a clean install on `purescript-halogen-realworld`. I'm on Ubuntu 18.04.3 if that helps.

```
$ spago build
[info] Installing 112 dependencies.
[info] Searching for packages cache metadata..
[info] Recent packages cache metadata found, using it..
Segmentation fault (core dumped)
```

However, the 0.11.1.0 dynamic binary completes successfully. I wonder if this is the "glibc doesn't like to be statically linked" thing...

f-f:
@hdgarrood yeah, I cut `0.11.0` the other day but I had to unpublish it from NPM since the static binary is broken (I wishfully switched over the NPM installer to use the static binary already), see #497 

So the problem is indeed that linking to `glibc` doesn't work. In CI we get this kind of notices when compiling the static build:
```
...
/home/ubuntu/.stack/snapshots/x86_64-linux/e375c66e4edfc5b6b3c1a145921eaf6866da84869d4b7830400eba4eff536958/8.4.4/lib/x86_64-linux-ghc-8.4.4/network-2.6.3.6-I7vksraVbZpFsulkC1XEeF/libHSnetwork-2.6.3.6-I7vksraVbZpFsulkC1XEeF.a(HsNet.o):function hsnet_getaddrinfo: warning: Using 'getaddrinfo' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
...
```

The widely-deployed solution seems to be to link against `musl` when compiling statically. One could use some `musl`-powered image in CI for that, but I was thinking about just using the [static-haskell-nix](https://github.com/nh2/static-haskell-nix/tree/master/static-stack2nix-builder-example) machinery to do that (this is the same machinery that Dhall uses to compile the static binaries, and it works wonderfully)

If there is interest for `purs` to do this too we could setup a Hydra server (using [Dhall's infra scripts](https://github.com/dhall-lang/dhall-lang/tree/master/nixops)) to run all the nix-based CIs on shared infrastructure

hdgarrood:
For `purs`, I think that switching to `musl` and statically linking everything just for the sake of `libtinfo` might be a bit excessive; I'd rather look for a way to statically link just the `libtinfo` library and leave everything else as-is first.

hdgarrood:
It seems like statically linking terminfo is not really doable without some extra work anyway: https://github.com/judah/terminfo/issues/36

PR #438, opened 2019-10-03
Make `spago verify-set` fail if there are duplicate module names.

Dretch:
This is needed for https://github.com/purescript/package-sets/issues/490

Note that `spago verify <package>` does _not_ do the check, only `spago verify-set` does. This is because the check requires downloading and compiling all packages, which would make `spago verify <package>` almost as slow as `spago verify-set`.

### Checklist:


justinwoo:
looks like it does what is desired. i don't think this is necessarily what someone who manages their own package set will want to do, but in those cases, someone might use `dhall-to-json <<< packages.dhall | jq 'keys[]' -r | xargs -I% spago verify %` or something as needed.

f-f:
@justinwoo great point. We might want to add a `--skip-duplicated-module-check` flag then? (it's not an actual suggestion for the flag name, but just to give the idea)

PR #439, opened 2019-10-04
Improve confusing warning when installing already added deps

jhrcek:
### Description of the change

Fix for https://github.com/spacchetti/spago/issues/436

### Checklist:

### DIFF
--- a/src/Spago/Config.hs
+++ b/src/Spago/Config.hs
@@ -284,8 +284,8 @@ updateName newName (Dhall.RecordLit kvs)
 updateName _ other = other
 
 addRawDeps :: Spago m => Config -> [PackageName] -> Expr -> m Expr
-addRawDeps config newPackages r@(Dhall.RecordLit kvs)
-  | Just (Dhall.ListLit _ dependencies) <- Dhall.Map.lookup "dependencies" kvs = do
+addRawDeps config newPackages r@(Dhall.RecordLit kvs) = case Dhall.Map.lookup "dependencies" kvs of
+  Just (Dhall.ListLit _ dependencies) -> do
       case notInPackageSet of
         -- If none of the newPackages are outside of the set, add them to existing dependencies
         [] -> do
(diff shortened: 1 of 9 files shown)


src/Spago/Config.hs
```diff
@@ -373,5 +376,5 @@ transformMExpr rules =
 addDependencies :: Spago m => Config -> [PackageName] -> m ()
 addDependencies config newPackages = do
   configHasChanged <- withConfigAST $ addRawDeps config newPackages
-  when (not configHasChanged) $ do
-    echo "WARNING: configuration file was not updated. You should have a record with the `dependencies` key for this to work."
+  unless configHasChanged $
```
jhrcek: The key change: `configHasChanged == False` can happen for multiple reasons:
1. Misconfiguration of spago.dhall (e.g. missing `dependencies` field, or `dependencies` field not having type `List String`)
2. spago.dhall already contains the dep which user tried to install
3. adding empty package list

Fix - make this warning more generic and give more granular warnings in `addRawDeps`

templates/spago.dhall
```diff
@@ -5,7 +5,7 @@ You can edit this file as you like.
 { name =
     "my-project"
 , dependencies =
-    [ "effect", "console", "psci-support" ]
+    [ "console", "effect", "psci-support" ]
```
jhrcek: Why am I changing the order in this template?

Simply to make `spago install <SOMETHING_ALREADY_IN_DEPENDENCIES>` operation idempotent.
The order is now alphabetical, so the above come (which sorts the dependencies ins `addRawDeps`) doesn't change this config if it's not necessary.

test/Utils.hs
```diff
@@ -55,13 +55,15 @@ shouldBeSuccess result@(_code, _stdout, _stderr) = do
   result `shouldSatisfy` (\(code, _, _) -> code == ExitSuccess)
 
 shouldBeSuccessOutput :: HasCallStack => FilePath -> (ExitCode, Text, Text) -> IO ()
-shouldBeSuccessOutput expected result = do
-  expectedContent <- readFixture expected
-  result `shouldSatisfy` (\(code, stdout, _stderr) -> code == ExitSuccess && stdout == expectedContent)
+shouldBeSuccessOutput expected (code, stdout, _stderr) = do
+  expectedStdout <- readFixture expected
+  code `shouldBe` ExitSuccess
```
jhrcek: I was struggling when fixing the tests, because the assertion error wasn't specific enough.
It was saying something like "Predicate failed blabla" from which it wasn't clear why it failed.
With this change you get more precise string diff about how exactly the stdout differed.

jhrcek:
Tests failed on AppVeyor due to github rate limit. Not sure if I should rerun the tests (rate limit will probably be still in place at the moment)..

test/Utils.hs
f-f: Looks great! 👍

templates/spago.dhall
f-f: Yep, great call 💯

src/Spago/Config.hs
f-f: Yeah this is the best we cann do for now. I'd like to figure something out so that we can give specific error messages instead of this generic one, but this change is indeed much better than giving the wrong message out 🙂

PR #443, opened 2019-10-05
Make `spago install purescript-*` install the package without prefix instead of just give warning

jhrcek:
### Description of the change

This is my second attempt at improving the situation around https://github.com/spacchetti/spago/issues/367

In my first attempt (not published) I tried the approach you suggested in https://github.com/spacchetti/spago/issues/367#issuecomment-520241101:
I extended `getTransitiveDependencies` so that it can handle also `purescript-` prefix (by retrying failed lookup without the prefix). But this wasn't enough to fix the problem, because there were couple more places before that function is called which also lookup package in package set, which required further modifications.

I took a step back and looked at all the code which depends on `getTransitiveDependencies`.
Here's a diagram showing which top-level commands depend on it (arrow X->Y means function X is calling Y):
getTransitiveDependencies

As you can see quite a lot of commands depend on getting transitive dependencies.
This made me think that perhaps we should fix this in some better place. Like we should sanitize user input before we even allow it into package set.

So the fix in this PR only makes `spago install purescript-XYZ` behave like you describe in the description.

I guess my question now is: **are there other commands / scenarios apart from `spago install`
 that exhibit some problems caused by not stripping `purescript-` prefix?**

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,16 +14,17 @@ Breaking changes (😱!!!):
   `psc-package-local-setup`, `psc-package-insdhall` and `psc-package-clean` commands.
 
 New features:
-- Display a link to the generated docs' `index.html` (#379)
-- Add `--open` flag to `spago docs` which opens generated docs in browser (#379)
-- Support building for alternate backends (#355). E.g: Use `backend = "psgo"` entry in `spago.dhall` to compile with `psgo`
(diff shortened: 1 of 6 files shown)


jhrcek:
@f-f thank you for the kind words and for your trust in inviting me as a collaborator. I'll do my best to be a good collaborator :-)

I didn't add the change to the changelog because I was expecting more changes to be needed. I did one more commit fixing this and I also unified the latest changes to be a bit clearer and more unified.

PR #445, opened 2019-10-06
Speedup version bump tests

jhrcek:
### Description of the change

Attempt to speed up test suite by focusing on the slowest tests as described in https://github.com/spacchetti/spago/issues/440

Locally this change seems to have shaved off ~20s of test running time.

### Checklist:

### DIFF
--- a/src/Spago/Version.hs
+++ b/src/Spago/Version.hs
@@ -66,19 +66,18 @@ getCurrentVersion = do
       pure maxVersion
 
 
--- | Get the next version to use, or die if this would result in the version number going down/not changing.
 getNextVersion :: VersionBump -> SemVer -> Either Text SemVer
-getNextVersion spec version@SemVer{..} =
+getNextVersion spec currentV@SemVer{..} =
   case spec of
     Major -> Right $ SemVer (_svMajor + 1) 0 0 [] []
     Minor -> Right $ SemVer _svMajor (_svMinor + 1) 0 [] []
     Patch -> Right $ SemVer _svMajor _svMinor (_svPatch + 1) [] []
-    Exact v
-      | v > version -> Right v
+    Exact newV
+      | currentV < newV -> Right newV
       | otherwise -> do
-        let new = unparseVersion v
-            old = unparseVersion version
-        Left $ "The new version (" <> new <> ") must be higher than the current version (" <> old <> ")"
+        let new = unparseVersion newV
+            current = unparseVersion currentV
+        Left $ "The new version (" <> new <> ") must be higher than the current version (" <> current <> ")"
 
 
 -- | Make a tag for the new version.
--- a/src/Spago/Version.hs
+++ b/src/Spago/Version.hs
@@ -2,7 +2,8 @@ module Spago.Version
   ( VersionBump(..)
   , DryRun(..)
   , bumpVersion
-  , parseVersion
+  , getNextVersion
+  , parseVersionBump
   , unparseVersion
   ) where
 
@@ -24,6 +25,15 @@ data VersionBump
   | Minor
   | Patch
   | Exact SemVer
+
+
+parseVersionBump :: Text -> Maybe VersionBump
+parseVersionBump = \case
+  "major" -> Just Major
+  "minor" -> Just Minor
+  "patch" -> Just Patch
+  v | Right v' <- parseVersion v -> Just $ Exact v'
+  _ -> Nothing
 
 
 -- | Parses a version, ignoring an optional leading 'v', or returns an error message.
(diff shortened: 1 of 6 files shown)


src/Spago/Version.hs
```diff
@@ -55,18 +67,18 @@ getCurrentVersion = do
 
 
 -- | Get the next version to use, or die if this would result in the version number going down/not changing.
-getNextVersion :: Spago m => VersionBump -> SemVer -> m SemVer
+getNextVersion :: VersionBump -> SemVer -> Either Text SemVer
```
jhrcek: Made this pure so it's easier to unit test..

test/BumpVersionSpec.hs
```diff
@@ -53,70 +59,118 @@ setOverrides overrides = do
...
-    before_ (initGitTag "v1.3.4") $ it "Spago should bump patch version" $ do
+      it "Spago should complain when no git repo exists" $ do
 
-      spago ["bump-version", "--no-dry-run", "patch"] >>= shouldBeSuccess
-      getHighestTag >>= (`shouldBe` Just "v1.3.5")
+        spago ["bump-version", "minor"] >>= shouldBeFailureInfix
+          "Your git working tree is dirty. Please commit or stash your changes first"
 
-    before_ (initGitTag "v1.3.4") $ it "Spago should bump minor version" $ do
+      before_ (initGitTag "v1.2.3") $ it "Spago should only make a tag with `--no-dry-run`" $ do
```
jhrcek: The basic idea was to replace [these tests](https://github.com/spacchetti/spago/blob/52ddd50747bd139474008780cc612396a4b5477b/test/BumpVersionSpec.hs#L73-L94) with just one - they all test the same path, just differ in git tag version, which I think is adequately tested in the unit/property tests above.

src/Spago/Version.hs
Dretch: Seems reasonable. It is possibly worth re-wording the comment so it doesn't mention `die`.

test/BumpVersionSpec.hs
```diff
@@ -53,70 +59,118 @@ setOverrides overrides = do
...
+        parseVersionBump "1.2.3" `shouldBe` Just (Exact (SemVer 1 2 3 [] []))
 
-    before_ (initGitTag "v1.2.3") $ it "Spago should not make a tag when not passing --no-dry-run" $ do
+      -- TODO is this desired behavior, or should we just drop ONE 'v'? I'd agree it's edge case, but still :-)
+      it "should drop multiple 'v's from the beginning" $
+        parseVersionBump "vvvvvvvv1.2.3" `shouldBe` Just (Exact (SemVer 1 2 3 [] []))
 
-      spago ["bump-version", "minor"] >>= shouldBeSuccess
-      getHighestTag >>= (`shouldBe` Just "v1.2.3")
+      -- TODO should this work or should we strip these in parser implementation?
```
Dretch: PREREL / META should probably be allowed - for one thing because there are existing packages that have such versions - e.g. `halogen` `v5.0.0-rc.7`.

test/BumpVersionSpec.hs
```diff
@@ -53,70 +59,118 @@ setOverrides overrides = do
...
+      it "should parse version starting with 'v'" $
+        parseVersionBump "v1.2.3" `shouldBe` Just (Exact (SemVer 1 2 3 [] []))
 
-      spago ["bump-version", "minor"] >>= shouldBeFailureInfix
-        "Your git working tree is dirty. Please commit or stash your changes first"
+      it "should parse version not starting with 'v'" $
+        parseVersionBump "1.2.3" `shouldBe` Just (Exact (SemVer 1 2 3 [] []))
 
-    before_ (initGitTag "v1.2.3") $ it "Spago should not make a tag when not passing --no-dry-run" $ do
+      -- TODO is this desired behavior, or should we just drop ONE 'v'? I'd agree it's edge case, but still :-)
```
Dretch: I don't know. :man_shrugging:

test/BumpVersionSpec.hs
Dretch: :+1:

src/Spago/Version.hs
jhrcek: Good point. I actually went for deleting this comment, as I find the code self-explanatory enough.

jhrcek:
@f-f just wanted to check with you what's the general strategy around PR reviews. Should I wait for your review/approval before I merge, or is just one review enough, like in the case of this PR?

f-f:
@jhrcek great point, thanks for asking. This has not come up before so that's why we don't have a written policy about it yet

I think a review from a contributor is enough to merge, no need to wait for my review :)

f-f:
@jhrcek thanks for doing this 🙂

PR #446, opened 2019-10-07
Filter out from watch list files in `.spago` folder

Benjmhart:
### Description of the change

TODO

### Checklist:

### DIFF
--- a/src/Spago/Build.hs
+++ b/src/Spago/Build.hs
@@ -139,7 +139,7 @@ build buildOpts@BuildOptions{..} maybePostBuild = do
 
     wrap   = Purs.SourcePath . Text.pack
     unwrap = Text.unpack . Purs.unSourcePath
-    removeDotSpago = filter $ not . (".spago" `List.isInfixOf`)
+    removeDotSpago = filter (\glob -> ".spago" `notElem` splitDirectories glob)
 
 -- | Start a repl
 repl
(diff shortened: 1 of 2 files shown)


Benjmhart:
original issue https://github.com/spacchetti/spago/issues/430

src/Spago/Build.hs
```diff
@@ -137,6 +137,7 @@ build buildOpts@BuildOptions{..} maybePostBuild = do
 
     wrap   = Purs.SourcePath . Text.pack
     unwrap = Text.unpack . Purs.unSourcePath
+    removeDotSpago = filter $ not . (".spago" `List.isInfixOf`)
```
jhrcek: This approach seems reasonable to me. There's however one edge case that comes to my mind:
For example path like this `/home/John/my.spago.playground/my.project/` will be removed by this filter too. This particular  case could be prevented by treating `.spago` as directory name (see [splitDirectories](https://www.stackage.org/haddock/lts-14.7/filepath-1.4.2.1/System-FilePath-Posix.html#v:splitDirectories)) as opposed to infix of a string (`my.spago.playground`). But maybe I'm just too cautious :laughing: 

```suggestion
    removeDotSpago = filter (\glob -> ".spago" `notElem` splitDirectories glob)
```

jhrcek: Sorry, I just did one-line suggestion as that's what github comments UI allows. You'll need to add import to make the CI work.

Benjmhart: Thanks for the heads up,   I'm running a local build to verify and then I'll update the PR with code that actually works.

PR #450, opened 2019-10-08
Fix hlint warnings

jhrcek:
### Description of the change

I fixed relatively conservative subset of hlint warnings which IMO make the code more lightweight/readable by removing some redundancies (unnecessary parens, import aliases, using better functions where available etc.)

### Checklist:

### DIFF
--- a/src/Spago/PackageSet.hs
+++ b/src/Spago/PackageSet.hs
@@ -10,18 +10,20 @@ module Spago.PackageSet
 
 import           Spago.Prelude
 
-import qualified Data.Text           as Text
-import qualified Data.Versions       as Version
+import           Data.Ord        (comparing)
+import qualified Data.Text       as Text
+import qualified Data.Versions   as Version
 import qualified Dhall.Freeze
 import qualified Dhall.Pretty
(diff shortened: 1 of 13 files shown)


src/Spago/PackageSet.hs
```diff
@@ -231,31 +233,18 @@ rootPackagePath (Dhall.Import
...
 findRootOutputPath path = do
   echoDebug "Locating root path of packages.dhall"
   imports <- liftIO $ Dhall.readImports $ Text.pack path
   let localImports = mapMaybe rootPackagePath imports
-  pure $ (flip System.FilePath.replaceFileName) "output" <$> (findRootPath localImports)
+  pure $ flip System.FilePath.replaceFileName "output" <$> findRootPath localImports
 
 -- | Given a list of filepaths, find the one with the least folders
 findRootPath :: [System.IO.FilePath] -> Maybe System.IO.FilePath
-findRootPath paths
```
jhrcek: The only change not suggested by hlint, but I couldn't resist to simplify this to more readable one-liner..

Issue #434, opened 2019-10-03
CLI spec for alternate backends


f-f:
In #426 we added basic support for alternate backends. 

As discussed in https://github.com/spacchetti/spago/issues/355 we should write down a "CLI spec" that alternate backends can conform to so that they'd work with Spago out of the box.

csicar:
First idea to get the conversation started:

Option | Default | Description
-------- | --------- | --------------
no option | - | generates files in the target language and saves them in the directory ???
`--src` | `src/` | sets the directories where the purescript source files are located (used for finding FFI files)
`-o` | `output/[compilerName]`  | sets the directory, where the target files will be written to
`--run` | `Main.main` | builds & runs the supplied function. The command expects the function to have type `:: Effect Unit`
`--build` | - | compiles the program in the target language and writes the executable to the directory ???

Notes:
- it might be preferable to have `--build` be the default and `--no-build` be optional
- I'm not sure about the necessity of `--src` and `-o`, but as far as I know, spago supports alternative paths for `src` and `output` (is that correct?). Therefore it would make sense, that the backends also respect that.
- `pskt` writes generated executables in the `output/pskt/program.jar` file; AFAIK `psgo` writes the executable to `./Main`
- The default value for `--output` could also be just `output/`. (That's what `psgo` does)
- Hooking up to the `bundle-app` and `bundle-module` commands would be nice: For the alternative backends `bundle-app` would mean compiling to a executable (native exe or executable `.jar`) and `bundle-module` would mean compiling to a library (shared lib or non-executable jar)
    - The `--to` option would allow specifying where the compilation result should be saved.

f-f:
cc @nwolverson @andyarvanitis

andyarvanitis:
Looks good to me so far.
  1. I would vote for `--build` being the default (`--no-build` being optional).
  1. I'll look into `--src` and `-o` for both of my backends (I can at least provide a warning if one of them needs to be ignored)
  1. "The default value for `--output` could also be just `output/`. (That's what psgo does)" – Please!
  1. I can look into `bundle-app` and `bundle-module`

csicar:
I would say either having the `--build` or the `--no-build` flag, but not both as that would be confusing.
> I would vote for --build being the default (--no-build being optional).

So you mean, that without any options provided to the compiler it will create the files in the target language *and* build a binary?

andyarvanitis:
Right. I'm open to discussing it, though.

csicar:
[...]

PR #452, opened 2019-10-08
Add support for `spago run` with alternative backends

tenniscp25:
### Description of the change

- Made `spago run` call the backend with `--run` argument.
- Updated `spago run` specs for node backend.
- `spago test` still uses node backend.

As for the test, if creating a shell script and a batch file to echo $1 to check backend arg is the way to go, I'll be happy to implement it.

### Checklist:

### DIFF
--- a/src/Spago/Build.hs
+++ b/src/Spago/Build.hs
@@ -181,42 +181,51 @@ repl cacheFlag newPackages sourcePaths pursArgs depsOnly = do
 -- | Test the project: compile and run "Test.Main"
 --   (or the provided module name) with node
 test :: Spago m => Maybe Purs.ModuleName -> BuildOptions -> [Purs.ExtraArg] -> m ()
-test = runWithNode (Purs.ModuleName "Test.Main") (Just "Tests succeeded.") "Tests failed: "
+test maybeModuleName buildOpts extraArgs = do
+  Config.Config { alternateBackend } <- Config.ensureConfig
+  runBackend alternateBackend (Purs.ModuleName "Test.Main") (Just "Tests succeeded.") "Tests failed: " maybeModuleName buildOpts extraArgs
 
 -- | Run the project: compile and run "Main"
 --   (or the provided module name) with node
 run :: Spago m => Maybe Purs.ModuleName -> BuildOptions -> [Purs.ExtraArg] -> m ()
-run = runWithNode (Purs.ModuleName "Main") Nothing "Running failed, exit code: "
+run maybeModuleName buildOpts extraArgs = do
+  Config.Config { alternateBackend } <- Config.ensureConfig
+  runBackend alternateBackend (Purs.ModuleName "Main") Nothing "Running failed; " maybeModuleName buildOpts extraArgs
 
 -- | Run the project with node: compile and run with the provided ModuleName
 --   (or the default one if that's missing)
-runWithNode
+runBackend
   :: Spago m
-  => Purs.ModuleName
+  => Maybe Text
+  -> Purs.ModuleName
   -> Maybe Text
   -> Text
   -> Maybe Purs.ModuleName
   -> BuildOptions
   -> [Purs.ExtraArg]
   -> m ()
-runWithNode defaultModuleName maybeSuccessMessage failureMessage maybeModuleName buildOpts nodeArgs = do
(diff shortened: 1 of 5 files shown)


src/Spago/Build.hs
```diff
@@ -179,42 +179,49 @@ repl cacheFlag newPackages sourcePaths pursArgs depsOnly = do
...
+  echoDebug $ "Running with backend: " <> fromMaybe "nodejs" maybeBackend
+  let postBuild = maybe (nodeAction =<< getOutputPath buildOpts) backendAction maybeBackend
+  build buildOpts (Just postBuild)
   where
     moduleName = fromMaybe defaultModuleName maybeModuleName
-    args = Text.intercalate " " $ map Purs.unExtraArg nodeArgs
-    contents outputPath' =
-         let path = fromMaybe "output" outputPath'
+    nodeArgs = Text.intercalate " " $ map Purs.unExtraArg extraArgs
+    nodeContents = \outputPath'
```
jhrcek: It seems to me you didn't start off the most recent master. Yesterday I merged a R fixing bunch of hlint warnings. Could you please modify your PR so that these recent changes are respected? See https://github.com/spacchetti/spago/commit/9737c63b38ca6fbf65455edf2c344c9eec3ab8b9#diff-0b924580c83919e4251597a2dcf79b51L207-L217

E.g. on your PR branch you could do something like:
`git pull --rebase origin master` and then fix the merge conflicts..

f-f:
cc @csicar @andyarvanitis

csicar:
First of all, thank you for the contribution. It's awesome to see further work being done in that area.

This works properly with `pskt`. It does start the `pskt --run`, but I would be awesome if the `--run` options was passed the module to run. That way, we could also support `spago test`.

One problem, that remains is, that (at least `pskt`) does not know where it should look for the foreign files for the standard libraries. That problem is kinda orthogonal to this, but it means, that `spago run` does will fail for `pskt` while compiling the jar file, as the foreign files are not found.

src/Spago/Build.hs
```diff
@@ -179,42 +179,49 @@ repl cacheFlag newPackages sourcePaths pursArgs depsOnly = do
...
+      echoDebug $ "Writing .spago/run.js"
+      writeTextFile ".spago/run.js" (nodeContents outputPath')
       chmod executable ".spago/run.js"
-      shell cmd empty >>= \case
-        ExitSuccess   -> maybe (pure ()) echo maybeSuccessMessage
+      shell nodeCmd empty >>= \case
+        ExitSuccess   -> fromMaybe (pure ()) (echo <$> maybeSuccessMessage)
         ExitFailure n -> die $ failureMessage <> repr n
+    backendAction backend = do
+      Turtle.proc backend ["--run"] empty >>= \case
```
csicar: passing `["--run", moduleName]`  (or similar) as an argument to be awesome

src/Spago/Build.hs
tenniscp25: Fixed. Two hints remaining. Do you want me to fix them too?

Screen Shot 2019-10-10 at 2 37 58 AM

src/Spago/Build.hs
tenniscp25: I can do that but I find passing "Main" to psgo currently doesn't work. Should we first agree on module names for run and test actions?

csicar: At the moment `pskt` does not support that either. That addition can probably wait for #434 to be resolved. But it's good to know, that the change would not be too hard :)

src/Spago/Build.hs
jhrcek: Thank you! I leave it up to you to decide. Just pick whichever you find more readable :-)

tenniscp25:
Fixed.

tenniscp25:
Oops.. wrong button 🙈

f-f:
@tenniscp25 thank you for doing this! 🙂

andyarvanitis:
Happy to try this, but it will have to be this weekend. Looking forward to it!

andyarvanitis:
Runs fine with `psgo`; I tested both `spago build` and `spago run`. It looks like `spago test` behaves the same as `spago run` right now (I assume by design), which is fine as `psgo` doesn't have anything for test yet.

PR #453, opened 2019-10-10
Improve cache skipping warning

jhrcek:
### Description of the change

Fix 3 problems with this warning :smile: 
- is not actionable (not clear how user can fix this)
- it's not clear what problems the wrongly formatted causes to spago (the package won't be cached locally) 
- it's written as `"Warning"` where all the other warnings have `"WARNING"`

(sidenote: also fixed the url in the package-set: https://github.com/purescript/package-sets/pull/514)

### Checklist:

### DIFF
--- a/src/Spago/GlobalCache.hs
+++ b/src/Spago/GlobalCache.hs
@@ -65,7 +65,7 @@ globallyCache (packageName, Repo url, ref) downloadDir metadata cacheableCallbac
           cacheableCallback $ Turtle.encodeString resultDir
       where
     _ -> do
-      echo $ "Warning: was not able to match url with GitHub ones: " <> url
+      echo $ "WARNING: Not caching repo, because URL doesn't have the form of 'https://github.com/<ORG>/<REPO>.git': " <> url
       notCacheableCallback -- TODO: error?
   where
     isTag = do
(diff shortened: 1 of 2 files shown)


src/Spago/GlobalCache.hs
```diff
@@ -65,7 +65,7 @@ globallyCache (packageName, Repo url, ref) downloadDir metadata cacheableCallbac
           cacheableCallback $ Turtle.encodeString resultDir
       where
     _ -> do
-      echo $ "Warning: was not able to match url with GitHub ones: " <> url
+      echo $ "WARNING: Not caching repo, because URL doesn't have the form of 'https://github.com/<ORG>/<REPO>.git': " <> url
```
jhrcek: I was also thinking about making the parsing logic more robust (i.e. not to always expect `.git` at the end of url) but I rejected it because I think it's best to have unified format for all packages in the package-set.

f-f:
I have the feeling we could also try to accept URLs not containing `.git` (and I think the Curator implements this already)

src/Spago/Watch.hs
```diff
@@ -80,7 +80,7 @@ fileWatchConf watchConfig shouldClear inner = withManagerConf watchConfig $ \man
             -- and the last event either has different path, or has happened
             -- more than `debounceTime` seconds ago.
             let shouldRebuild =
-                   any matches (Set.toList globs)
+                   any matches globs
```
jhrcek: Unrelated minor fix: I realized that `Set` of globs is also foldable, so `any :: Foldable t => (a -> Bool) -> t a -> Bool` will work with it and we can avoid this conversion to List for every filesystem change event.

src/Spago/GlobalCache.hs
f-f: Yeah the unified format is good, though accepting URLs not containing `.git` should be pretty harmless (and I think the Curator implements this already)

However we don't have to change this now and we can do it in a followup PR

PR #463, opened 2019-10-20
Add  `spago output-path` command to show output folder

danieljharvey:
### Description of the change

Following #422 - this adds an `output-path` command that returns the output path we are using. Useful in a monorepo so that that Webpack etc can share the same folder and make builds faster.

### Checklist:


jhrcek:
The implementation seems fine to me. I just have some forward-looking musings.
Consider the current behavior of Haskell's `stack` tool. It has a similar command 
`stack path` which prints all the paths that user might be interested in

```bash
$ stack path
snapshot-doc-root: /home/myuser/something...
local-doc-root: /home/myuser/something/else...
local-hoogle-root: /home/myuser/other/thing ...
...
```
You can then add specific name of the path and it will just output one path:
```bash
$ stack path local-doc-root
/home/myuser/something/else...
```

I'm just thinking if users could potentially request other commands similar to this? If yes, it might be more future proof to adopt approach similar to stack and have all the "path printing commands" available via one command instead of introducing separate command for each.

danieljharvey:
Aye, good point. Stack’s design decisions are generally pretty sensible IMHO. I’m not going to be able to work on this until next week, but if at that point there is a vague consensus here for the above I’ll happily implement.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ New features:
 - `spago install purescript-XYZ` will now strip `purescript-` prefix and install XYZ (if it exists in package set) instead of just failing with a warning (#367)
 - `spago run` now recognizes backend specified in the configuration file and calls the backend with `--run` argument.
 - documentation now includes a step-by-step guide on setting up a Spago/Parcel project (#456)
+- `spago output-path` returns the output path so that it can be shared with tools such as `purs-loader`
```
jhrcek: There are still 5 occurrences of `output-path` in this PR, that should be changed to `path`.

danieljharvey:
Alright, this should be sorted now, works as follows:

```
$ spago path
output: ./some/output/folder
```

```
$ spago path output
./some/output/folder
```

Does this seem right?

.gitignore
```diff
@@ -6,3 +6,4 @@ test/spago-test
 **/__pycache__
 templates/docs-search-app.js
 templates/purescript-docs-search
+.DS_Store
```
turboMaCk: I see the file commited. You should use `git rm` to get rid of it.

danieljharvey: Ahh yes. Sorry about that, removed now!

danieljharvey:
Ace, made those changes.

f-f:
Thank you @danieljharvey!

PR #485, opened 2019-11-06
Make global config path absolute

### LINKED ISSUE #472: --config option doesnt accept absolute path
if you do

```
YOUR_PROJECT_DIR_FULLPATH=$HOME/example
YOUR_PROJECT_DIR_RELATIVE_PATH=example
cd $SOME_OTHER_DIR
```

then 

```
$ spago --config $YOUR_PROJECT_DIR_FULLPATH/spago.dhall list-packages --filter transitive --json
spago:
Error: Invalid input

(input):1:3:
  |
1 | .//home/srghma/projects/spago2nix/spago.dhall
  |   ^
unexpected '/'
expecting expression or whitespace
```

and 

```
$ spago --config $YOUR_PROJECT_DIR_RELATIVE_PATH/spago.dhall list-packages --filter transitive --json
↳ ./packages.dhall

Error: Missing file ./packages.dhall

21│     ./packages.dhall

21:5
```

_______

real example

```
[nix-shell:~/projects]$ spago --config $HOME/projects/spago2nix/spago.dhall list-packages --filter transitive --json
spago:
Error: Invalid input

(input):1:3:
  |
1 | .//home/srghma/projects/spago2nix/spago.dhall
  |   ^
unexpected '/'
expecting expression or whitespace


[nix-shell:~/projects]$ spago --config spago2nix/spago.dhall list-packages --filter transitive --json
spago:
↳ ./packages.dhall

Error: Missing file ./packages.dhall

21│     ./packages.dhall

21:5

[nix-shell:~/projects]$
````

_____

the second error `Missing file ./packages.dhall` is a bit different, I believe that path should be derived from `spago.dhall` `packages` option

```
{-
Welcome to a Spago project!
-}
{ sources =
    [ "src/**/*.purs", "test/**/*.purs" ]
, name =
    "my-project"
, dependencies =
    [ "console"
    , "generics-rep"
    , "kishimen"
    , "node-fs-aff"
    , "parallel"
    , "prelude"
    , "simple-json-utils"
    , "sunde"
    , "validation"
    ]
, packages =
    ./packages.dhall
}
```

### PR DESCRIPTION
**Pull request by elliotdavies**
### Description of the change

Fix #472 by making the global config path absolute. As a result of this we also need to make the output path absolute in order for `spago test` to continue working; and this in turn means a bunch of the test fixtures needed updating.

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ Bugfixes:
 - Fix dynamic libraries compatibility problems by publishing a statically linked executable for Linux (#427, #437)
 - `--clear-screen` (usable e.g. with `spago build --watch`) now also resets cursor position, so the rebuild message always appears at top left of the screen (#465)
 - Fix `--config` option when config file is in another directory (#484)
+- Make `--config` work correctly with both relative and absolute paths (#485)
 
 Other improvements:
 - Speed up test suite by replacing couple of end 2 end bump-version tests with unit/property tests
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -354,7 +354,9 @@ runWithEnv GlobalOptions{..} app = do
         $ setLogUseColor True
         $ setLogVerboseFormat True
         $ logOptions'
-  let configPath = fromMaybe Config.defaultPath globalConfigPath
+
+  configPath <- fmap Text.pack . makeAbsolute . Text.unpack $ fromMaybe Config.defaultPath globalConfigPath
+
   logDebug'  "Running `getGlobalCacheDir`"
   globalCache <- getGlobalCacheDir
   withLogFunc logOptions $ \logFunc ->
--- a/src/Spago/Build.hs
+++ b/src/Spago/Build.hs
@@ -210,14 +210,12 @@ runBackend
   -> Spago ()
 runBackend maybeBackend defaultModuleName maybeSuccessMessage failureMessage maybeModuleName buildOpts extraArgs = do
   logDebug $ display $ "Running with backend: " <> fromMaybe "nodejs" maybeBackend
-  let postBuild = maybe (nodeAction =<< getOutputPath buildOpts) backendAction maybeBackend
+  let postBuild = maybe (nodeAction =<< getOutputPathOrDefault buildOpts) backendAction maybeBackend
   build buildOpts (Just postBuild)
   where
     moduleName = fromMaybe defaultModuleName maybeModuleName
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

**Comment by elliotdavies** [comment]
If this goes in after #484 I can add a test like the one in that PR but with an absolute path.

**Comment by elliotdavies** [comment]
Hmm this change apparently breaks `spago test` because the pretty printing inside `findRootOutputPath` seems to mangle absolute paths... I will investigate more later

**Comment by elliotdavies** [comment]
@f-f Should be good to go now I think

**Comment by f-f** [comment]
@elliotdavies the Windows CI seems still unhappy about this, I'm not sure what might be up with it 🤔

And sorry for the bunch of merge conflicts, but we needed #486 to be in as soon as possible as that's swapping out the whole logging system so it would conflict with any other PR anyways

**Comment by elliotdavies** [comment]
@f-f No worries, I can rebase!

Not sure about the Windows CI though :scream: Looks like it's calculating the absolute paths directly but something (Dhall?) doesn't like them.

**Comment by f-f** [comment]
@elliotdavies I'm afraid I don't have a Windows machine to test this on, but maybe @stkb might have some idea of what's happening here?

**Comment by stkb** [comment]
I will take a look

**Comment by f-f** [comment]
Hold on, I might have realized what's up here: I think Dhall doesn't support absolute paths on Windows!

I guess this could be reported upstream, but I now wonder if we can get away with a smaller fix here. E.g. what if we just change this line:

https://github.com/spacchetti/spago/blob/97a2131fbcce67c3af62255d24bddbae65a167b3/src/Spago/Config.hs#L115

..to something like

```hs
expr <- liftIO $ Dhall.inputExpr $ if isAbsolute path then path else "./" <> path
```

stkb:
@f-f While I don't understand all the code yet, that change still produces the same error on my machine

f-f:
@stkb note that I mean the change above to be on top of `master`, not on top of this PR. If you applied it on top of `master` and it still gives you that error then it's also good because it means we have found a smaller replication

elliotdavies:
> ..to something like
> 
> ```haskell
> expr <- liftIO $ Dhall.inputExpr $ if isAbsolute path then path else "./" <> path
> ```

@f-f This seems to work for me :+1: I can put another PR up with the more minimal change. (By not making the path absolute when we read it in there is a chance we'll run into this problem again in future, but I guess we can cross that bridge when we get to it.)

f-f:
@elliotdavies yes sounds great!

PR #486, opened 2019-11-06
Switch all the logging to RIO

f-f:
This PR introduces [the `rio` library](https://tech.fpcomplete.com/haskell/library/rio), mostly to rehaul our logging situation: this replaces our `logDebug`/`logError`/etc. with rio's own functions.
This is to reduce the amount of code that we have to write here to support logging, and to get some logging goodies for free, e.g. colors and timestamps in debug logs.

Sample of the new logging format:
```
$ spago init
[info] Set up a local Spago project.
[info] Try running `spago build`
```

cc @Benjmhart: this introduces a `logInfo` that logs to `stderr`, so we can still use `output` for outputting to `stdout`


Benjmhart:
Does this cover all the cases for application output vs application logging that you described, or is there still more work to be done with respect to logging?

f-f:
@Benjmhart initially I thought I'd just introduce the library here and then switch the logging in a separate PR, but things kind of snowballed (because of the test fixtures) so I guess I'll tackle everything in this PR 😄

PR #489, opened 2019-11-09
Do not error when there are no tests

### LINKED ISSUE #383: `spago test` errors when there are no tests
Running `spago test` on a project where there isn't a `test` directory results in this error:

```
Installation complete.
Build succeeded.
internal/modules/cjs/loader.js:670
    throw err;
    ^

Error: Cannot find module '../output/Test.Main'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:668:15)
    at Function.Module._load (internal/modules/cjs/loader.js:591:27)
    at Module.require (internal/modules/cjs/loader.js:723:19)
    at require (internal/modules/cjs/helpers.js:14:16)
    at Object.<anonymous> (/home/elliot/<filepath elided>/.spago/run.js:3:1)
    at Module._compile (internal/modules/cjs/loader.js:816:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:827:10)
    at Module.load (internal/modules/cjs/loader.js:685:32)
    at Function.Module._load (internal/modules/cjs/loader.js:620:12)
    at Function.Module.runMain (internal/modules/cjs/loader.js:877:12)
spago: Tests failed: 1
```

I think this should probably either succeed (because in effect 0/0 tests passed) or fail in a nicer manner.

waddlaw:
### Description of the change

- Fixes #383
- Implement module parser

I'm not familiar with PureScript, so could you check it?

### Checklist:


waddlaw:
@f-f Thank you for merging!

I read the source code of the PureScript parser. The current implementation is incomplete compared to the original parser. Is it better to implement everything? ([Language.PureScript.CST.Parser](https://github.com/purescript/purescript/blob/v0.13.4/src/Language/PureScript/CST/Parser.y))

I also considered using the PureScript Compiler API, but gave up because it didn't build. There are also concerns that dependencies will be heavy.

```haskell
moduleHeader :: { Module () }
  : 'module' moduleName exports 'where' '\{' moduleImports
      { (Module () $1 $2 $3 $4 $6 [] []) }

moduleName :: { Name N.ModuleName }
  : UPPER {% upperToModuleName $1 }
  | QUAL_UPPER {% upperToModuleName $1 }

exports :: { Maybe (DelimitedNonEmpty (Export ())) }
  : {- empty -} { Nothing }
  | '(' sep(export, ',') ')' { Just (Wrapped $1 $2 $3) }

export :: { Export () }
  : ident { ExportValue () $1 }
  | symbol { ExportOp () $1 }
  | properName { ExportType () $1 Nothing }
  | properName dataMembers { ExportType () $1 (Just $2) }
  | 'type' symbol { ExportTypeOp () $1 $2 }
  | 'class' properName { ExportClass () $1 $2 }
  | 'kind' properName { ExportKind () $1 $2 }
  | 'module' moduleName { ExportModule () $1 $2 }

ident :: { Name Ident }
  : LOWER {% toName Ident $1 }
  | 'as' {% toName Ident $1 }
  | 'hiding' {% toName Ident $1 }
  | 'kind' {% toName Ident $1 }

symbol :: { Name (N.OpName a) }
  : SYMBOL {% toName N.OpName $1 }
  | '(..)' {% toName N.OpName $1 }

properName :: { Name (N.ProperName a) }
  : UPPER {% toName N.ProperName $1 }

dataMembers :: { (DataMembers ()) }
  : '(..)' { DataAll () $1 }
  | '(' ')' { DataEnumerated () (Wrapped $1 Nothing $2) }
  | '(' sep(properName, ',') ')' { DataEnumerated () (Wrapped $1 (Just $2) $3) }
```

f-f:
@waddlaw great points!

I think that we should not depend on the compiler package, as I feel it's way too big. So I'd say we could port part of the parser here (as we're doing now) and eventually split off the parser from the compiler and depend on that one instead (the idea for this to happen is already there, see https://github.com/purescript/purescript/issues/3633)

And yes, I'd say it's better to implement everything from the actual parser, otherwise we risk failing on otherwise good source code. If you want to PR the change please base the new code on top of #492, where I did a small refactoring of the relevant code

PR #493, opened 2019-11-11
Fix Ctrl-C handling in Windows NPM shim (#483)

### LINKED ISSUE #483: Hitting Ctrl-C in "spago repl" messes up prompt (cmd.exe, powershell.exe)
Having opened up a new `cmd` shell and starting `spago repl` in a simple project if I hit <kbd>Ctrl-c</kbd> something brokes ... the REPL exits, but after that if I restart it in the same `cmd` session my input gets mangled and garbled.

Instead if I do this

`purs repl "src\**\*.purs" ".spago\*\*\src\**\*.purs"`

all is fine. <kbd>Ctrl-C</kbd> clears the current line and keeps things running, and most importantly the shell's input doesn't break. (Same things happen in `cmd.exe` and `powershell.exe`.

(Not sure if this makes sense or not, I'll try to make a little demo of this somehow when I got the time.)

stkb:
Should fix #483

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ Bugfixes:
 - Fix dynamic libraries compatibility problems by publishing a statically linked executable for Linux (#427, #437)
 - `--clear-screen` (usable e.g. with `spago build --watch`) now also resets cursor position, so the rebuild message always appears at top left of the screen (#465)
 - Fix `--config` option when config file is in another directory (#484)
+- Fix Ctrl-C handling in REPL when using NPM installation on Windows (#493, #483)
 
 Other improvements:
 - Speed up test suite by replacing couple of end 2 end bump-version tests with unit/property tests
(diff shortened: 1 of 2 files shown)


stkb:
@kohanyirobert, could you try on your machine, replacing the contents of the `spago` file (no extension) in your global `node_modules/spago` folder, with [this new version](https://raw.githubusercontent.com/spacchetti/spago/ad572c632c4b84acb94c84270b996763c336c640/npm/spago)? It should make Ctrl-C work correctly in `spago repl`.

There is unfortunately still a slight issue. If you use Ctrl-C in the repl, then after you quit it (either with `:quit` or Ctrl-D) you get 2x this annoying prompt: 
```
Terminate batch job (Y/N)?
```

This isn't caused by this change, but appears now because we're now exiting the process properly. (It's caused by the npm shims: if you use `purs repl ...` with a npm purescript installation, you get that prompt when quitting the repl too; but when you use `spago repl` you get it twice because you also have a spago npm shim). That prompt is [annoying and difficult to get rid of](https://superuser.com/questions/35698/how-to-supress-terminate-batch-job-y-n-confirmation/35741)

f-f:
@stkb I'm about to cut a release so let's merge this anyways, and we can followup with other PRs if we find issues with this fix

stkb:
@f-f ok!

Issue #393, opened 2019-08-28
Run package-set commands without `spago.dhall` present


elliotdavies:
Some commands such as `verify-set` and `list-packages` don't work if a `spago.dhall` file is missing from the current directory. This can feel weird in a monorepo setup where (for example) there's a root `packages.dhall` and multiple `spago.dhall`s in sub-directories, because it means the commands must be run from a sub-directory away from the actual package-set definition. This is also the opposite behaviour from commands like `upgrade-set` and `freeze` which can _only_ be run in the same directory as `packages.dhall`.

I guess we should go through the commands, figure out whether they're a "requires `packages.dhall`" or a "requires `spago.dhall`", and fix them appropriately - I'm not sure how much reorganisation this might require. The Spago help text already informally sorts commands into "project" and "package-set" lists so we could also sync up with that.

This likely relates to #349.

f-f:
@elliotdavies good point, and yes this kind of relates to #349, but in a funny way: if we have commands search for the `packages.dhall` then they would fail in folders where there isn't one
So we'd need the resolution from #349, or some kind of fallback like "search for the `packages.dhall` first, if not found just read the `spago.dhall` and get the `PackageSet` from there"

There is no particular reason why `verify` requires a `spago.dhall`, but the reason why `list-packages` does is that you can pass `--filter=transitive` and `--filter=direct`, in which case Spago would need the list of dependencies in the config.
On the other hand I was thinking about splitting that command into `spago ls package-set` (which wouldn't require a `spago.dhall`) and `spago ls deps` and `spago ls transitive` (which would).
The next release is going to be breaking anyways so this could be a good moment to implement this

PR #515, opened 2019-11-29
Make `verify-set` work in the presence of either `.dhall` file

elliotdavies:
### Description of the change

Fixes one specific problem highlighted in #393, though not the general case.

@f-f Wanted to check the general approach with you before I tidy it up. The error messages in particular might be a bit odd because functions like `ensureConfig` call `die` internally, but I worry about this spiralling if I start changing all of those...

### Checklist:

### DIFF
--- a/src/Spago/Packages.hs
+++ b/src/Spago/Packages.hs
@@ -350,6 +350,9 @@ verify :: Maybe CacheFlag -> CheckModulesUnique -> Maybe PackageName -> Spago ()
 verify cacheFlag chkModsUniq maybePackage = do
   logDebug "Running `spago verify`"
 
+  -- @TODO Swap the order of these to check for `spago.dhall` first, once
+  -- `ensureConfig` no longer calls `die` internally. See:
+  -- https://github.com/spacchetti/spago/pull/515#pullrequestreview-329632196
   packageSet@PackageSet{..} <- do
     -- Try to read a "packages.dhall" directly
     try (liftIO (Dhall.inputExpr $ "./" <> PackageSet.packagesPath)) >>= \case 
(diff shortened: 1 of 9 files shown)


src/Spago/Packages.hs
```diff
@@ -349,7 +349,18 @@ data CheckModulesUnique = DoCheckModulesUnique | NoCheckModulesUnique
 verify :: Maybe CacheFlag -> CheckModulesUnique -> Maybe PackageName -> Spago ()
 verify cacheFlag chkModsUniq maybePackage = do
   logDebug "Running `spago verify`"
-  Config{ packageSet = packageSet@PackageSet{..}, ..} <- Config.ensureConfig
+
+  packageSet@PackageSet{..} <- do
+    -- Try to read a "packages.dhall" directly
+    try (liftIO (Dhall.inputExpr $ "./" <> PackageSet.packagesPath)) >>= \case
```
f-f: I'd invert the order here: we should try to read a config first (as that is present most often than the `packages.dhall`), and only if that fails we try to read the `packages.dhall` file

elliotdavies: @f-f If I put `ensureConfig` first then we'll still see its `die` error message even if the verify command later succeeds because it finds a `package.json` :disappointed: 

I can leave it as-is, or we could leave this until I/we sort out the `die` situation?

f-f: Right, let's add a `FIXME` or `TODO` here and leave it as is for now

f-f:
@elliotdavies thank you for doing this! 🙂

elliotdavies:
@f-f No problem, sorry it took so long to finish it off!

PR #516, opened 2019-11-30
Test generating haddocks on CI

cdepillabout:
### Description of the change

Makes sure haddocks can be generated with stack when building spago.

The first commit (ad8c43c) enables building haddocks on CI, in order to make sure that haddock generation doesn't break again.  If you try to run `stack build --haddock --no-haddock-deps` with just this commit, you can confirm that the build will fail because of spago's incorrect use of haddocks.

The second commit (6ac17f4) fixes the haddocks in spago's source code.  This  _should_ enable CI to pass, but it looks like it is currently failing for unrelated reasons.

Fixes https://github.com/spacchetti/spago/issues/511

### Checklist:


.travis.yml
```diff
@@ -86,14 +86,14 @@ script:
     export PATH="${PATH}:$(pwd)/artifacts"
     if [ `uname` = "Darwin" ]
     then
-      stack build --test --no-run-tests --copy-bins --local-bin-path ./artifacts;
+      stack build --test --no-run-tests --copy-bins --local-bin-path ./artifacts --haddock --no-haddock-deps;
```
cdepillabout: Here (and a few other places), the `--haddock` flag has been added.

This builds the haddocks when building spago.  The build will fail if the haddocks for `spago` cannot be generated.

By default, all the haddocks for all depedencies are also built.  Unfortunately, the haddocks for [github](http://hackage.haskell.org/package/github) don't build correctly, so I had to use the `--no-haddock-deps` flag.  This makes sure haddocks all all dependencies are not built.

cdepillabout: The bug in the haddocks for github appears to be fixed in the `master` branch on the repo, so `--no-haddock-deps` could potentially be removed here in the future.

f-f: Yeah, building Haddocks should work again once we are back on the latest GHC/cabal/haddock versions (currently pinning to LTS-12 due to supporting older macOS versions), but having `--no-haddock-deps` should be fine for now!

PR #532, opened 2019-12-20
Add support for `--before`, `--then` and `--else` flags

### LINKED ISSUE #410: Feature request: Add support for --before/--then/--else flags
Pulp has a handful of very useful options to run a command when compilation starts, fails, and succeeds: `--before <string>`, `--else <string>`, and `--then <string>`, respectively. This is useful in combination with `--watch` in order to, for example, run some post-processing script on the output. Currently the only way to achieve the same is using an external file watcher and run `spago build && foobar` on every file change. Aside from the additional overhead of getting set up properly for every project, there's also the overhead of cold-starting a build every time (see also: #409, #110) which slows down rebuilds considerably.

FungusHumungus:
Fix #410

### Description of the change

Added support for commands to be run around the build. `--before` is run before the build, `--then` following a successful build and `--else` following an unsuccessful build.

I realise there were some alternative suggestions in the issue, but I personally prefer the original suggestion as I really liked the way it worked with Pulp.

### Checklist:

- [*] Added the change to the "Unreleased" section of the changelog
- [*] Added some example of the new feature to the `README`
- [*] Added a test for the contribution (if applicable)

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,25 @@ $ spago build --watch
 $ spago build --watch --clear-screen
 ```
 
+To run a command before a build you can use the `--before` flag, eg to clear the screen before a build:
+
+```bash
+$ spago build --watch --before clear
+```
+
+To run a command after the build, use `--then` for successful builds, or `--else` for unsuccessful builds:
+
+```bash
+$ spago build --watch --then "notify-send 'Built successfully'" --else "notify-send 'Build failed'"
+```
+
+Multiple commands are possible - they will be run in the order specified:
+
+```bash
+$ spago build --watch --before clear --before "notify-send 'Building'"
+```
+
+
 If you want to run the program (akin to `pulp run`), just use `run`:
 ```bash
 # The main module defaults to "Main"
--- a/README.md
+++ b/README.md
@@ -303,10 +303,10 @@ $ spago build --watch
 $ spago build --watch --clear-screen
 ```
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -303,6 +303,25 @@ $ spago build --watch
 $ spago build --watch --clear-screen
 ```
 
+To run a command before a build you can use the `--before` flag, eg to clear the screen before a build:
+
+```bash
+$ spago build --watch --before clear
```
f-f: To clear the screen we already have the `--clear-screen` option, could we pick an alternative example here? Or alternatively show them side-by-side

FungusHumungus: Good point. I'll try to think of an alternative example.

FungusHumungus:
It has also occurred to me that the build should be cancelled if the command fails in some way. Currently it is ignoring the result.

FungusHumungus:
I have changed the example just just post a notification. Inspiration for anything else just wasn't coming!

Now if the command returns an error, it will halt the build.

I have also moved the `then` command so it runs after the `postBuild`. I feel it makes more sense to run this after bundling has completed when running `bundle-app`. It could be useful to do some processing with the bundled file.

Let me know if you want me to revert either of these additional changes.

f-f:
@FungusHumungus all the new changes look great, thank you for doing this! :slightly_smiling_face:

FungusHumungus:
It was a pleasure! :-)

Issue #316, opened 2019-07-13
Build messages with spago run should go to stderr


justinwoo:
So I can do `spago run > my-output-only-pls.txt`

related to #256

f-f:
@justinwoo since we just let the output of the compiler through this is more related to https://github.com/purescript/purescript/issues/3672 right?

Or should we just pour all the output from the compiler to stderr?

justinwoo:
I would like that too, redirecting all of the compiler output to stderr

f-f:
@justinwoo oh right, I just realised that you'd like _all_ the output from spago to go to `stderr`

f-f:
Closing this as duplicated of #256, as I reported all the relevant things from here in https://github.com/spacchetti/spago/issues/256#issuecomment-513116049

Issue #170, opened 2019-04-05
Quiet/Silent option


jgoux:
Hello,

would it be possible to add a `--silent` or `--quiet` option to `spago build` so we don't have these messages : 
```
File changed, rebuilding: "/home/jgoux/Projects/eurofins/colisage/purescript/server/src/Main.purs"
Compiling Main
Build succeeded.
Success! Waiting for next file change.
Type help for available commands. Press enter to force a rebuild.
```

My use case is when using the `--watch` option, I'd like to see the app's output only when the project rebuilds.

f-f:
Hi @jgoux!

I have some questions:
- What output would you like to keep here?
  - Only the `purs` output or just nothing?
  - Or should we just hide the "File changed" part?
- What do you mean by "the app's output"? If you're doing `spago build` then the app is not running, but you're compiling the project with `purs`

If we want to just hide the "File changed" then we could emit it only when `--verbose` is specified. If we want to hide more, then we could add levels to the `--verbose` flag (as proposed in https://github.com/spacchetti/spago/issues/154#issuecomment-474883921 - so that e.g. `--verbose 0` would hide output) or just add the `--silent`

jgoux:
Hello @f-f ,

I'd like to be able to hide all the lines I showed.

I'm using nodemon to reload my index.js file when I have a change inside `output/**/*.js`

So I'd like to see only the output of my app which is what is logged from the app itself : 

![image](https://user-images.githubusercontent.com/1443499/55780782-074d7380-5aa9-11e9-821a-753aab44559b.png)

In this example, I'd like to see only the lines with the emoji, which are logged from my app's code.

Here is my package.json scripts : 
```json
"scripts": {
    "dev": "run-p dev:*",
    "dev:purescript": "spago build --watch",
    "dev:js": "nodemon -q -w output -w index.js index.js",
    "build": "spago build && ncc build index.js"
  }
```

f-f:
Thanks @jgoux, I see now! Some more queries:
- Would it work to do `spago build --watch > /dev/null` instead?
- Did you know about `spago run --watch`?

Issue #154, opened 2019-03-20
Make printing `purs` commands optional


starper:
I think it would be better to add something like `--debug` flag for it.

Because some messages can be really big. Like, for example, this one:
```
Running command: `purs compile  ".spago/aff/v5.1.0/src/**/*.purs" 
".spago/ansi/v5.0.0/src/**/*.purs" ".spago/arraybuffer-types/v2.0.0/src/**/*.purs" 
".spago/arrays/v5.2.0/src/**/*.purs" ".spago/avar/v3.0.0/src/**/*.purs" 
".spago/bifunctors/v4.0.0/src/**/*.purs" ".spago/catenable-lists/v5.0.1/src/**/*.purs" 
".spago/console/v4.2.0/src/**/*.purs" ".spago/const/v4.1.0/src/**/*.purs" 
".spago/contravariant/v4.0.0/src/**/*.purs" ".spago/control/v4.1.0/src/**/*.purs" 
".spago/datetime/v4.1.1/src/**/*.purs" ".spago/debug/v4.0.0/src/**/*.purs" 
".spago/distributive/v4.0.0/src/**/*.purs" ".spago/effect/v2.0.1/src/**/*.purs" 
".spago/either/v4.1.1/src/**/*.purs" ".spago/enums/v4.0.1/src/**/*.purs" 
".spago/exceptions/v4.0.0/src/**/*.purs" ".spago/exists/v4.0.0/src/**/*.purs" 
".spago/foldable-traversable/v4.1.1/src/**/*.purs" ".spago/foreign/v5.0.0/src/**/*.purs" 
".spago/foreign-generic/v8.1.0/src/**/*.purs" ".spago/foreign-object/v2.0.1/src/**/*.purs" 
".spago/free/v5.1.0/src/**/*.purs" ".spago/functions/v4.0.0/src/**/*.purs" 
".spago/functors/v3.1.1/src/**/*.purs" ".spago/gen/v2.1.0/src/**/*.purs" 
".spago/generics-rep/v6.1.1/src/**/*.purs" ".spago/globals/v4.0.0/src/**/*.purs" 
".spago/identity/v4.1.0/src/**/*.purs" ".spago/integers/v4.0.0/src/**/*.purs" 
".spago/invariant/v4.1.0/src/**/*.purs" ".spago/js-date/v6.0.0/src/**/*.purs" 
[...]

PR #540, opened 2020-01-13
Add logging options to facilitate json errors and general command-line use

matthew-hilty:
### Description of the change

Spago has a lot of functionality. One of its primary responsibilities, though, is to make using the `purs` executable easier -- that is, Spago calls `purs` directly, provides some output guidance, and interleaves logging metadata but otherwise stays out of the compiler's way. Because end users employ Spago during development to interact with `purs`, there have been a few issues (e.g., #154, #170, #256, #316) asking for greater separation between Spago logging and `purs` output. PR #475 does a good job of this by having Spago log to __stderr__. However, as @hdgarrood mentions in #256, this approach doesn't fully isolate Spago metadata from `purs` data, since the compiler prints json errors to the same __stderr__ stream. For common Spago use cases, this mingling of streams isn't much of a problem in itself, but it does make pipelining on the command-line difficult, since Spago's logging messages then affect downstream processing. This pull request offers a possible remedy by giving users greater control over Spago logging, including the option of dedicating a file handle exclusively to Spago output and the option of suppressing logging entirely.

Here's an example of the former CLI option. In the bash command below, logging by `purs` on __stdout__ is suppressed and logging by Spago on Spago's own file handle 3 is redirected to __stderr__. Because compilation errors, originally on __stderr__, are redirected to __stdout__, they can be piped for additional processing.

``` bash
$ spago --output-stream 3 build -n -u --json-errors 3>&2 2>&1 1>/dev/null | ./format.sh

{
  "warnings": [
    {
      "warningCode": "UnusedImport",
      "loc": "start: 4, end: 4, file: src/Main.purs",
      "msg": " The import of Data.Either is redundant "
    }
  ],
  "errors": [
    {
      "errorCode": "TypesDoNotUnify",
      "loc": "start: 10, end: 10, file: src/Main.purs",
      "msg": " Could not match type Effect with type Function Int while trying to match type Effect Unit with type Int -> t0 while checking that expression (log \"\\127837\") 3 has type Effect Unit in value declaration main where t0 is an unknown type "
    }
  ]
}
[error] Failed to build.
```

_src/Main.purs_
``` purescript
module Main where

import Prelude (Unit)
import Data.Either
import Effect (Effect)
import Effect.Console (log)

main :: Effect Unit
main = do
  log "🍝" 3
```

_format.sh_
``` bash
#!/bin/sh
sed --unbuffered 's/\\n/ /g; s/\s\+/ /g' \
  | jq --unbuffered \
  ' { warnings:
        .warnings
        | [.[]
            | { warningCode: .errorCode
              , loc: "start: \(.position.startLine), end: \(.position.endLine), file: \(.filename)"
              , msg: .message
              }
          ]
    , errors:
        .errors
        | [.[]
            | { errorCode: .errorCode
              , loc: "start: \(.position.startLine), end: \(.position.endLine), file: \(.filename)"
              , msg: .message
              }
          ]
    }
  '
```
-----
For comparison, and to show the usefulness of Spago pipelining, here's the same compilation output but without --json-errors and without formatting.

``` bash
$ spago --output-stream 3 build -n -u --json-errors 3>&2 2>&1 1>/dev/null

{"warnings":[{"position":{"startLine":4,"startColumn":1,"endLine":4,"endColumn":19},"message":"  The import of Data.Either is redundant\n","errorCode":"UnusedImport","errorLink":"https://github.com/purescript/documentation/blob/master/errors/UnusedImport.md","filename":"src/Main.purs","moduleName":"Main","suggestion":{"replacement":"","replaceRange":{"startLine":4,"startColumn":1,"endLine":4,"endColumn":19}},"allSpans":[{"start":[4,1],"name":"src/Main.purs","end":[4,19]}]}],"errors":[{"position":{"startLine":10,"startColumn":3,"endLine":10,"endColumn":12},"message":"  Could not match type\n\n    Effect\n\n  with type\n\n    Function Int\n\n\nwhile trying to match type Effect Unit\n  with type Int -> t0\nwhile checking that expression (log \"\\127837\") 3\n  has type Effect Unit\nin value declaration main\n\nwhere t0 is an unknown type\n","errorCode":"TypesDoNotUnify","errorLink":"https://github.com/purescript/documentation/blob/master/errors/TypesDoNotUnify.md","filename":"src/Main.purs","moduleName":"Main","suggestion":null,"allSpans":[{"start":[10,3],"name":"src/Main.purs","end":[10,12]}]}]}
[error] Failed to build.
```

``` bash
$ spago -o 3 build -n 3>&2 1>/dev/null

Warning found:
in module Main
at src/Main.purs:4:1 - 4:19 (line 4, column 1 - line 4, column 19)

  The import of Data.Either is redundant


See https://github.com/purescript/documentation/blob/master/errors/UnusedImport.md for more information,
or to contribute content related to this warning.


Error found:
in module Main
at src/Main.purs:10:3 - 10:12 (line 10, column 3 - line 10, column 12)

  Could not match type

    Effect

  with type

    Function Int


while trying to match type Effect Unit
  with type Int -> t0
while checking that expression (log "\127837") 3
  has type Effect Unit
in value declaration main

where t0 is an unknown type

See https://github.com/purescript/documentation/blob/master/errors/TypesDoNotUnify.md for more information,
or to contribute content related to this error.


[error] Failed to build.
```

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,19 +293,6 @@ E.g. if you wish to output your files in some other place than `output/`, you ca
 $ spago build --purs-args '-o myOutput/'
 ```
 
-It's possible to suppress or redirect Spago's logging in order to interact directly
-with `purs` output. The following examples use `jq` to format compilation errors
-rendered by `purs` as JSON. Specifically, the third example distinguishes between
-`purs` output and Spago output by configuring Spago to direct its own logging messages
-(but not those from `purs`) to file descriptor 3 for separate handling.
-
-```bash
-$ spago --quiet build --purs-args --json-errors 2> >( jq )
-$ spago --output-stream stdout build --purs-args --json-errors 2> >( jq )
-$ touch spago-log # Create a file for Spago's logging messages.
-$ spago -o 3 build -w -l -u --json-errors 3>> spago-log 4>&2 2>&1 1>&4 4>&- | jq
-```
-
 If you wish to automatically have your project rebuilt when making changes to source files
 you can use the `--watch` flag:
 
--- a/README.md
+++ b/README.md
@@ -293,6 +293,15 @@ E.g. if you wish to output your files in some other place than `output/`, you ca
 $ spago build --purs-args '-o myOutput/'
 ```
 
+It's possible to suppress or redirect Spago's logging in order to interact directly
+with `purs` output. The following two examples use `jq` to format compilation errors
+rendered by `purs` as JSON.
+
+```bash
+$ spago --quiet build --purs-args '--json-errors' 2> >( jq )
+$ spago --output-stream stdout build --purs-args '--json-errors' 2> >( jq )
+```
+
 If you wish to automatically have your project rebuilt when making changes to source files
 you can use the `--watch` flag:
 
@@ -1153,9 +1162,9 @@ that is accepted by many commands. You can either:
 
 ### Know the output path for my compiled code
 
-As there are now various factors that can affect the output path of compiled code, run 
-`spago path output` along with any flags you would pass to `spago build` (like 
-`--purs-args` or `--no-share-output`) to return the output path Spago is using. 
+As there are now various factors that can affect the output path of compiled code, run
+`spago path output` along with any flags you would pass to `spago build` (like
+`--purs-args` or `--no-share-output`) to return the output path Spago is using.
 This can be useful for sharing an output folder with `webpack`, for instance.
 
 ## Explanations
--- a/README.md
+++ b/README.md
@@ -303,7 +303,7 @@ rendered by `purs` as JSON. Specifically, the third example distinguishes betwee
 $ spago --quiet build --purs-args --json-errors 2> >( jq )
 $ spago --output-stream stdout build --purs-args --json-errors 2> >( jq )
 $ touch spago-log # Create a file for Spago's logging messages.
-$ spago -O 3 build -w -l -u --json-errors 3>> spago-log 4>&2 2>&1 1>&4 4>&- | jq
+$ spago -o 3 build -w -l -u --json-errors 3>> spago-log 4>&2 2>&1 1>&4 4>&- | jq
 ```
 
 If you wish to automatically have your project rebuilt when making changes to source files
--- a/README.md
+++ b/README.md
@@ -294,12 +294,16 @@ $ spago build --purs-args '-o myOutput/'
 ```
 
 It's possible to suppress or redirect Spago's logging in order to interact directly
-with `purs` output. The following two examples use `jq` to format compilation errors
-rendered by `purs` as JSON.
+with `purs` output. The following examples use `jq` to format compilation errors
+rendered by `purs` as JSON. Specifically, the third example distinguishes between
+`purs` output and Spago output by configuring Spago to direct its own logging messages
+(but not those from `purs`) to file descriptor 3 for separate handling.
 
 ```bash
-$ spago --quiet build --purs-args '--json-errors' 2> >( jq )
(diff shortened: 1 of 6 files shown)


matthew-hilty:
Is Spago now part of Purescript's official tooling? It looks like ownership of the project has changed. I'm wondering, because I was going to suggest that this PR has broader merit than ease of `json-errors` use alone, that Spago at times acts more like a pair (or more) of semi-independent applications than a single application, and that this two-program conceptual model perhaps warrants a non-traditional multi-"typed" streaming interface. However, now I'm not so sure.

I suppose there could still be a case made for granular output streams. Granularity would seem to make command-line use easier in general (users could project a tuple of streams onto the desired component rather than filter a single merged stream) and to better isolate Spago and compiler functionality. But if Spago and `purs` are now to be developed in concert, chances of contention between the two, present or future, are reduced and, at worst, any such conflict likely resolved within a release cycle. Component insularity, then, from among the long list of design goals, becomes less important and perhaps not worth the effort or complexity costs.

That's a reasonable argument to make, and if that's the case, does this pull request have any parts that could still be of use? Are there featurelets that I can strip out and resubmit?

matthew-hilty:
A related question about Spago's `watch` mode: What's the best channel for this interactive feature? Right now, `watch` mode's messaging is directed to __stderr__. This makes sense -- it's not compiler output; it's something else entirely. But at the same time, it doesn't seem like the diagnostic metadata of [Jesse Storimer](https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout) either; rather, it looks like the console interface of a running program. __stderr__ is probably the better choice between it and __stdout__, but I wonder if, for command-line interactivity, this design decision shouldn't be second-guessed.

f-f:
@matthew-hilty 

> chances of contention between the two, present or future, are reduced and, at worst, any such conflict likely resolved within a release cycle

That's the hope, and in this case it [looks like this should be fixed in 0.14](https://github.com/purescript/purescript/issues/3672#issuecomment-578111564)

> does this pull request have any parts that could still be of use? Are there featurelets that I can strip out and resubmit?

Yes! The `--quiet` and `--no-color` flags are more than welcome and I'd like to merge them in. Could you open a new PR just with them so we can keep the discussion about the `output-stream` here?

> A related question about Spago's watch mode: What's the best channel for this interactive feature? 

If I understand correctly: you'd like to move the interactive prompts for the watch mode to `stdout`, right? I think this is the right thing to do (for the reasons you mentioned) and I'd welcome a PR for this change :slightly_smiling_face:

matthew-hilty:
Awesome! Thanks, @f-f. I'll get right on opening a new PR for the `--quiet` and `--no-color` flags.

matthew-hilty:
As for the watch mode...

> If I understand correctly: you'd like to move the interactive prompts for the watch mode to `stdout`, right? I think this is the right thing to do (for the reasons you mentioned)

Actually, I wasn't too sure. Ideally, the interactive prompts should be moved to `stdout`, I think. However, once json errors are sent to `stdout` in 0.14, if the interactive prompts are also on `stdout`, processing json errors downstream while in watch mode would be harder.

Leaving the prompts on `stderr` seems counter to `stderr`'s purpose, and users wouldn't have the option of suppressing logging, but maybe it's the better choice. I just wanted to bring it up as something to consider.

f-f:
@matthew-hilty it looks to me that there's consensus on [fixing the issue in the compiler](https://github.com/purescript/purescript/issues/3672#issuecomment-578111564) being the way forward for now - should we close this PR?

I think your points about the logging messages while in watch mode are good, so feel free to move those to a new issue so we can track that separately

f-f:
Closing as outdated

PR #548, opened 2020-01-25
Adds logging flags

matthew-hilty:
### Description of the change

Adds `--quiet` (`-q`) flag that suppresses all Spago logging.
Adds `--no-color` (`-C`) flag that disables ANSI color escape sequences.

### Checklist:

### DIFF
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -105,13 +105,14 @@ data Command
 
 
 data GlobalOptions = GlobalOptions
-  { globalVerbose     :: Bool
+  { globalQuiet       :: Bool
+  , globalVerbose     :: Bool
   , globalVeryVerbose :: Bool
+  , globalUseColor    :: Bool
   , globalUsePsa      :: UsePsa
   , globalJobs        :: Maybe Int
   , globalConfigPath  :: Maybe Text
   }
-
 
 parser :: CLI.Parser (Command, GlobalOptions)
 parser = do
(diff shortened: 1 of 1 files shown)


f-f:
@matthew-hilty thank you for doing this! :slightly_smiling_face:

matthew-hilty:
@f-f , you're very welcome. Thank you too!

shamansir:
Seems having some output when using `spago` as a builder for [VSCode extension](https://github.com/nwolverson/vscode-ide-purescript), breaks it from building, being exact: `spago build --purs-args --json-errors` gives as an output (for version 0.13.1):

```
[info] Installation complete.
{"warnings":[],"errors":[]}
[info] Build succeeded.
```

So this is the important fix, please publish it as a version.

f-f:
@shamansir I just cut [a new release](https://github.com/purescript/spago/releases/tag/0.14.0), it should be up on npm shortly

shamansir:
@f-f Awesome, thank you a lot!

f-f:
You're welcome :slightly_smiling_face:

shamansir:
Seems spago `v0.14.0` still outputs _Compiling ..._ messages, even with `--quiet` flag.

I am running the command `spago --quiet build --purs-args --json-errors` in my project and there are lines of:

```
Compiling Data.List
Compiling Data.Maybe
. . . <hundreds of lines omitted>
{"warnings":[{"moduleName":...<here goes JSON>...}
```

Which could break the VSCode plugin parser which still expects only to get JSON.

Could be related to #540

matthew-hilty:
Hi, @shamansir.

I think the "Compiling ..." statements are ```ProgressMessage```s emitted by the purs compiler itself, not by spago. And, unfortunately, users have no option of silencing progress messages.

However, as of purs v0.13.6 and at least until purescript issue 3672 is resolved, there is a potential workaround. Since purs emits progress messages onto stdout and json-errors onto stderr, to ignore the "Compiling ..." statements while retaining json-errors, you need only redirect a couple of file descriptors: 

```bash
spago --quiet build --purs-args --json-errors 2>&1 1>/dev/null
```

f-f:
Since that output is coming straight from `purs`, something we could do in this case would be to not output that at all when `--quiet` is passed. Would that work?

matthew-hilty:
I don't think there's a command-line flag or any other hook accessible to Spago code for silencing the "Compiling ..." statements, although it seems like a good feature for `purs` developers to include. However, until that point, there doesn't seem to be any officially supported way for `purs` clients (either Spago itself or Spago's end users) to filter out progress messages.

As far as I can tell, there are only two possible solutions, both kludges:
1. Spago can pattern match on all of `purs`'s output and strip progress messages from `stdout`. (However, I wouldn't recommend this; my guess is that this tactic would be rather error prone.)
2. Command-line users like @shamansir can redirect `stdout` and its progress messages to some out-of-the-way file descriptor.

matthew-hilty:
Unless, do you mean, have Spago suppress the compiler's `stdout`-output entirely?

That would work, but it seems a tad heavy-handed. Most PureScript developers, I expect, have high technical competence and can probably redirect output streams for themselves. Having Spago do it for them might make for a better user-interface, but eventually, `purs` is going to change its `stdout`/`stderr` model anyway, so it might not be worth the bother.

PR #549, opened 2020-01-25
Fix a few watch-mode buffering and concurrency issues

matthew-hilty:
### Description of the change

In the `onChange` function, read from and _write to_ the timing TVar during the same transaction in order to prevent a race condition.

In the `redisplay` function, flush stdout after resetting the cursor since buffering might otherwise unduly delay this action.

### Checklist:


f-f:
Thank you both! :slightly_smiling_face:

matthew-hilty:
Thanks, @klntsky and @f-f  !!

Issue #545, opened 2020-01-22
Add --source-maps flag


Profpatsch:
```
> purs bundle --help
…
  --source-maps            Whether to generate source maps for the bundle
                           (requires --output).
```

```
> spago bundle-app --help
  --source-maps << why u no source map‽ 
```

It would be great if the `--source-maps` flag could be passed to purs. According to https://github.com/purescript/spago/issues/216#issuecomment-496279796 it’s possible via `purs bundle`, but I gave up reverse engineering how exactly `spago bundle-app` calls `purs` (well, it doesn’t, it calls the library, which makes it harder even).

This should be easy to add.

f-f:
@Profpatsch we don't use the library, but call the binary. The call to `bundle` happens here:
https://github.com/purescript/spago/blob/cc755561d992406533e0ca1fd2fe4087c047a7bb/src/Spago/Purs.hs#L66-L80

I'm up for adding a `--source-maps` flag to spago, and it would conditionally add a flag for `purs` as it happens in line 68.

However you can already pass things to `purs` today, this should work: `spago bundle-app --purs-args "--source-maps"`

Profpatsch:
Sadly it doesn’t, because that option passes flags on to the toplevel of purs:

```
Invalid option `--source-maps'

Usage: purs COMMAND
  The PureScript compiler and tools
```

Woody88:
Hello, 

I really appreciate this library, so I can make time this weekend to work on it!

Woody88:
Hey,

So, It implemented the feature. Everything seems to be working fine. 
I added "--source-maps" flag, and also added, "-x" as the shorter version because "s" and "m" were already taken. If you prefer another letter please let me know. 

I also added a test case: 
https://github.com/Woody88/spago/blob/aa0e2ac550ab0885e04bd774867ab49ab325e19f/test/SpagoSpec.hs#L648-L653

However, line 653 is giving me this error: 
`  1) Spago, spago bundle-app, Spago should bundle successfully with source map
[...]

PR #562, opened 2020-02-02
Add `--source-maps` flag

Woody88:
### Description of the change
Implemented issue #545  
### Checklist:


test/SpagoSpec.hs
```diff
@@ -645,7 +645,13 @@ spec = around_ setup $ do
       spago ["bundle-app", "--to", "bundle-app.js"] >>= shouldBeSuccess
       checkFixture "bundle-app.js"
 
+    it "Spago should bundle successfully with source map" $ do
 
+      spago ["init"] >>= shouldBeSuccess
+      spago ["bundle-app", "--to", "bundle-app-src-map.js", "--source-maps"] >>= shouldBeSuccess
+      checkFixture "bundle-app-src-map.js"
```
f-f: We already have a fixture for a `bundle-app` (in the test above), could we use that one instead of adding a new one?

test/fixtures/bundle-app-src-map.js
```diff
@@ -0,0 +1,31 @@
+// Generated by purs bundle 0.13.6
```
f-f: If we reuse the fixture that we already have then we would not need to add any new ones in this PR :slightly_smiling_face:

Woody88: Hmm... The problem is that the ".js" file that is generated with "--source-maps" flag appends the line below in the file:
`//# sourceMappingURL=<name-of-source-map-file>.js.map`

Here: https://github.com/Woody88/spago/blob/master/test/fixtures/bundle-app-src-map.js#L31

and here: https://github.com/Woody88/spago/blob/master/test/fixtures/bundle-module-src-map.js#L30

The test will fail with the ones that we already have. 

I noticed that added an extra file that is not needed, bundle-app-src-map.js.map, I will remove it. My bad...

f-f: Right. The reason why I'd like to reuse these files is that we have some code that automatically upgrades fixtures whenever a new purescript version is out (because the version number of the compiler appears several times in these files), here:
https://github.com/spacchetti/spago/blob/a35d100d0aef5a5fa76cc18ff2a0285bd094d58f/app/Curator.hs#L296-L310

So either:
- we reuse the fixture files and append that magic string when checking the fixtures
- ..or patch the new fixture as the existing ones in the code above

Woody88: Right, I see. Thanks for showing me this piece of code. Hmm.. I don't want to do too much processing in the test file because it seems like the flow that you guys have been following is to call functions from the Utils.hs module. I am not sure if the steps of appending the magic string belong in that module. Anyhow, just so we can get this going, I will go with your first suggestion!

f-f: Sorry if I didn't make this clearer, but my preference would be instead to just add two lines to do the `sed` thing in the code above, so that you don't have to change the test setup at all from what it is right now

f-f:
Thanks for doing this @Woody88! :smile:

Woody88:
@f-f It was a pleasure! Sorry for taking long, you know gotta balance work and life, I could only squeeze this for the weekends. I see that there are a lot of good first and second issues, I might look through the issues and see if I can tackle something else! Thanks for guiding me through this issue!

f-f:
@Woody88 you're welcome! And I know that some of these "good first issues" might be non very well specced or might be out of date by now, so feel free to ask for more guidance if you decide to pick up something else :slightly_smiling_face:

Profpatsch:
This is great, thank you for taking your time @Woody88

PR #587, opened 2020-03-05
Upgrade LTS (upgrades GHC to 8.6.5)

bbarker:
I tested this and spago could build and run. This helps keep it in line with the current version of GHC available in NixOS as well.

### Description of the change

Upgrade Stackage LTS version and GHC version. 

### Checklist:

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: lts-12.21
+resolver: lts-14.20
 packages:
 - .
 extra-deps:


f-f:
@bbarker thanks for this! We already upgraded to GHC 8.6 once, but had to downgrade again in #479 for compatibility reasons.

So I'd like to keep Spago on 8.4 until that's in the range of the last three versions of the compiler with an LTS (which is usually the moment when packages start dropping bounds), which means that I'd like to upgrade once there is an LTS with GHC 8.10 out.

arrowd:
I'm trying to build Spago with GHC 8.8.3. Judging from this thread, it is completely unsupported?

The error I got is

```
src/Spago/Dhall.hs:32:12: error:
    Not in scope: data constructor ‘Dhall.Format.Check’
    Perhaps you meant variable ‘Dhall.Format.input’ (imported from Dhall.Format)
    Module ‘Dhall.Format’ does not export ‘Check’.
   |
32 |   try (f $ Dhall.Format.Check path) >>= \case
   |            ^^^^^^^^^^^^^^^^^^

src/Spago/Dhall.hs:34:11: error:
    Not in scope: data constructor ‘Dhall.Format.Modify’
    Perhaps you meant one of these:
      ‘Dhall.Format.Format’ (imported from Dhall.Format),
      variable ‘Dhall.Format.format’ (imported from Dhall.Format)
    Module ‘Dhall.Format’ does not export ‘Modify’.
   |
34 |       f $ Dhall.Format.Modify path
   |           ^^^^^^^^^^^^^^^^^^^

```

f-f:
@arrowd it sounds like you're trying to build with `dhall-1.30`? We're still at `dhall-1.29`, but I'd welcome a PR to upgrade to the latest. However, Dhall itself might not build with 8.8 yet, see https://github.com/dhall-lang/dhall-haskell/issues/1249

arrowd:
Yep, downgrading `dhall` to `1.29` allowed me to build `spago` successfully.

f-f:
@bbarker since we're not looking to upgrade LTS any time soon, I'll just close this PR for now

PR #609, opened 2020-04-12
Do not exclude the metadata package from the set

f-f:


f-f:
Closing this in favor of #610, which includes this

PR #637, opened 2020-05-08
Add note about using single quotes for --purs-args

milesfrain:
It's confusing when porting commands from other tools which don't require quotes around purs args. For example:
This works:
```
psc-package build -- --codegen js,corefn
```
This does not:
```
spago build --purs-args --codegen js,corefn
```
This is required instead:
```
spago build --purs-args '--codegen js,corefn'
```


f-f:
Thank you for this! :slightly_smiling_face: 
Why not double quotes though? Single quotes do not work on Windows..

milesfrain:
Was basing it off of this example in the readme:
> spago build --purs-args '-o myOutput/'

Ugg... so annoying you can't link to lines in a .md file.

I'll change this PR to just say "quotes"

f-f:
Oh, thanks for noting, maybe we should change that example too :smile:

PR #647, opened 2020-06-15
Bump to LTS-16 with GHC-8.8.3.

cdepillabout:
### Description of the change

Hey, I noticed that spago-0.15.3 was released, so I wanted to update `spago` in nixpkgs.

I went to bump the version to 0.15.3, but I noticed that spago was no longer building.  This is because nixpkgs recently moved to the LTS-16 package set (GHC-8.8.3).  I needed to figure out how to get spago building with the LTS-16 package set.

This PR is the few small changes I needed to make.

I thought that you might be interested in this if you ever wanted to update spago to LTS-16, but if not, please feel free to just close this!

### Checklist:


src/Spago/Prelude.hs
```diff
@@ -95,7 +94,6 @@ import           Data.Sequence                         (Seq (..))
 import           Data.Text.Prettyprint.Doc             (Pretty)
 import           Data.Text.IO.Utf8                     (readFile, writeFile)
 import           Dhall.Optics                          (transformMOf)
-import           Lens.Family                           ((^..))
```
cdepillabout, line 98: This is apparently exported by RIO in more recent versions of RIO.

cdepillabout:
Here's my PR bumping spago to 0.15.3 in nixpkgs: https://github.com/NixOS/nixpkgs/pull/90470

f-f:
Thank you for this! I guess we could try moving to `lts-16`, so let's try to merge this one :slightly_smiling_face: 

Would you like to look at CI?
I can take care of the Linux CI, but here are the failures for the other targets:
- macOS fails with `spec: unexpected argument `./.stack-work/dist/x86_64-osx/Cabal-3.0.1.0/build/spec/spec'`. It might be enough to run tests locally to see what's up?
- Windows fails with `Module 'GHC.IO.Encoding.Iconv' does not export 'localeEncodingName'` while building `with-utf8`. I suppose we could pin its version?

cdepillabout:
@f-f Really, I just needed this for getting spago in nixpkgs (since the recent jump to lts-16).  If this is going to cause you a bunch of trouble to update, I don't know if it would be worth it.

I don't have a windows machine to play around on, so I'm afraid I won't be a big help figuring out the errors there.

I  do have an OSX machine, but I wasn't even able to get `stack` installed when I went to look into this!  (It kept telling me that I needed XCode installed...) So I don't think I will be able to help you much for non-Linux problems.

Again, please feel free to just close this (or wait for someone to come along with more experience on windows/osx).

f-f:
@cdepillabout no worries, I can take a look at this in a bit! Btw I think macOS CI failure should be debuggable on Linux, since it should be only about fixing tests :blush:

cdepillabout:
> I think macOS CI failure should be debuggable on Linux, since it should be only about fixing tests

Oh, I'm sorry, I can definitely take a look at that then!

I've been somewhat stressed out at work lately, but I'll try to get to this sometime this week.

f-f:
Yeah no worries at all - I'm in the same boat, have quite some stuff going on so I cannot look at things as much as I'd like to :sweat_smile:

cdepillabout:
@f-f I took a look at this.

I wasn't able to reproduce this error when running the tests locally on Linux, so I took at look at CI.

My guess is that MacOSX CI is failing from this line:

https://github.com/purescript/spago/blob/da6d91c19b23f06f3ede793f78599a6589c9e7cd/.travis.yml#L108-L109

https://travis-ci.org/github/purescript/spago/jobs/698489890#L8226

I'm guessing that there is more than one executable (or just file?) that matches ./.stack-work/dist/*/*/build/spec/spec, so one of them is executed and gets the other one as an argument.

f-f:
@hdgarrood it's fine by me, but I'll also point out that having NixOS build "from scratch" helps to:
1. gauge how "unreasonable" our build is (TL;DR: a bit. We have a piece of template haskell that does HTTP calls during the build. It's handy but really doesn't help with stuff like Nix builds :smile:)
2. update to newer compilers (as NixOS will usually trail in front of us with that), which helps also other distros that want to compile with the builting GHC, such as it happened with Voidlinux: they can just apply the patch in this PR and be done with it

Of course ultimately the decision should be taken by @cdepillabout since he's been maintaining the package in nixpkgs :slightly_smiling_face:

cdepillabout:
@hdgarrood Yeah, putting spago in nixpkgs as a binary is definitely one possibility.

With a little work, I've been able to keep spago compiling from source in nixpkgs, so I haven't needed to add it as a binary.  However, if I ever get behind with this (or keeping spago updated to compile in nixpkgs ever becomes a lot of work), I think it makes sense to add it as a binary similar to `purescript`.

There is also one advantage that Fabrizio didn't mention: Building spago from source makes it possible to use nixpkgs to produce different types of `spago` binaries.  For example, using musl instead of glibc, statically-linked `spago`,  `spago` that runs on aarch64, etc.

hdgarrood:
Thanks both, that’s useful to know.

f-f:
I think this is superseded by #695, so I'll close this.
However @cdepillabout let me know I missed something over there (i.e. do we still need patches for `nixpkgs` after that went in?)

PR #653, opened 2020-06-20
Fix README: webpack.config.js template

nineonine:
### Description of the change

In README we suggest to manually include the bundle script in `index.html` but in `webpack.config.js` we do not tell `HtmlWebpackPlugin` to avoid injection. As a result, the bundle will be injected twice.

See https://stackoverflow.com/a/38292765/3067181


nineonine:
> Looks great, thanks! Just added a small comment to clarify why we need this

Thanks! I executed.

f-f:
Thanks! :slightly_smiling_face:

PR #665, opened 2020-07-08
Ignore .gitignore files in `--watch`

### LINKED ISSUE #477: Ignore emacs tempfiles in `--watch`
As noted in https://github.com/spacchetti/spago/issues/465#issuecomment-546569568, emacs creates some temporary files that trigger unnecessary rebuilds when using the  `--watch` flag. We should just ignore these files (unless we find a usecase for not ignoring them, in which case we can make this configurable)

ozkutuk:
### Description of the change
Fixes #477

### Checklist:

### DIFF
--- a/src/Spago/CLI.hs
+++ b/src/Spago/CLI.hs
@@ -130,22 +130,23 @@ parser = do
     noColor     = CLI.switch "no-color" 'C' "Log without ANSI color escape sequences"
 
     -- Note: the first constructor is the default when the flag is not provided
-    force       = bool NoForce Force <$> CLI.switch "force" 'f' "Overwrite any project found in the current directory"
-    watch       = bool BuildOnce Watch <$> CLI.switch "watch" 'w' "Watch for changes in local files and automatically rebuild"
-    noInstall   = bool DoInstall NoInstall <$> CLI.switch "no-install" 'n' "Don't run the automatic installation of packages"
-    depsOnly    = bool AllSources DepsOnly <$> CLI.switch "deps-only" 'd' "Only use sources from dependencies, skipping the project sources."
-    noSearch    = bool AddSearch NoSearch <$> CLI.switch "no-search" 'S' "Do not make the documentation searchable"
-    clearScreen = bool NoClear DoClear <$> CLI.switch "clear-screen" 'l' "Clear the screen on rebuild (watch mode only)"
-    noBuild     = bool DoBuild NoBuild <$> CLI.switch "no-build" 's' "Skip build step"
-    srcMapFlag  = bool WithoutSrcMap WithSrcMap <$> CLI.switch "source-maps" 'x' "Whether to generate source maps for the bundle"
-    jsonFlag    = bool JsonOutputNo JsonOutputYes <$> CLI.switch "json" 'j' "Produce JSON output"
-    dryRun      = bool DryRun NoDryRun <$> CLI.switch "no-dry-run" 'f' "Actually perform side-effects (the default is to describe what would be done)"
-    usePsa      = bool UsePsa NoPsa <$> CLI.switch "no-psa" 'P' "Don't build with `psa`, but use `purs`"
-    openDocs    = bool NoOpenDocs DoOpenDocs <$> CLI.switch "open" 'o' "Open generated documentation in browser (for HTML format only)"
-    noComments  = bool WithComments NoComments <$> CLI.switch "no-comments" 'C' "Generate package.dhall and spago.dhall files without tutorial comments"
-    configPath  = CLI.optional $ CLI.optText "config" 'x' "Optional config path to be used instead of the default spago.dhall"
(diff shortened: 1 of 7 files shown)


src/Spago/Watch.hs
```diff
@@ -191,3 +201,13 @@ globToParent glob = go pathHead pathRest
     go acc ("**":_rest) = acc
     go acc [_file]      = acc
     go acc (h:rest)     = go (acc </> h) rest
+
+
+-- | Check if the path is ignored by git
+--
+-- `git check-ignore` exits with 1 when path is not ignored, and 128 when
+-- a fatal error occurs (i.e. when not in a git repository).
+isIgnored :: MonadIO m => Text -> m Bool
```
f-f, line 210: Let's add a `FIXME` about unifying this with the other implementation that we have in `Spago.Git`, together with the reason why there is another implementation

src/Spago/Watch.hs
```diff
@@ -76,29 +78,37 @@ fileWatchConf watchConfig shouldClear inner = withManagerConf watchConfig $ \man
...
+          unless pathIgnored $ do
+            rebuilding <- liftIO $ atomically $ do
+              globs                <- readTVar allGlobs
+              (lastTime, lastPath) <- readTVar lastEvent
+
+              let matches glob = Glob.match glob eventPath
+              -- We should rebuild if at least one of the globs matches the path,
+              -- and the last event either has different path, or has happened
+              -- more than `debounceTime` seconds ago.
+              let shouldRebuild =
```
f-f, line 98: Can we add the `isPathIgnored` check to the `shouldRebuild` instead of wrapping the whole block with it? By doing so we'd keep all the logic about "should we rebuild given a path" in a single place :slightly_smiling_face:

src/Spago/CLI.hs
```diff
@@ -130,22 +130,23 @@ parser = do
...
-    dryRun      = bool DryRun NoDryRun <$> CLI.switch "no-dry-run" 'f' "Actually perform side-effects (the default is to describe what would be done)"
-    usePsa      = bool UsePsa NoPsa <$> CLI.switch "no-psa" 'P' "Don't build with `psa`, but use `purs`"
-    openDocs    = bool NoOpenDocs DoOpenDocs <$> CLI.switch "open" 'o' "Open generated documentation in browser (for HTML format only)"
-    noComments  = bool WithComments NoComments <$> CLI.switch "no-comments" 'C' "Generate package.dhall and spago.dhall files without tutorial comments"
-    configPath  = CLI.optional $ CLI.optText "config" 'x' "Optional config path to be used instead of the default spago.dhall"
-    chkModsUniq = bool DoCheckModulesUnique NoCheckModulesUnique <$> CLI.switch "no-check-modules-unique" 'M' "Skip checking whether modules names are unique across all packages."
-    transitive  = bool NoIncludeTransitive IncludeTransitive <$> CLI.switch "transitive" 't' "Include transitive dependencies"
+    force        = bool NoForce Force <$> CLI.switch "force" 'f' "Overwrite any project found in the current directory"
```
f-f, line 133: Could we not realign all of this? It's fine to have things misaligned if we get a nicer diff!

src/Spago/Watch.hs
ozkutuk: That was my initial plan. However, `shouldRebuild` is in the `STM` block. I don't know much about `stm` myself, however AFAIK there is no safe way to execute `IO` actions within `STM`. If there is any other way than the current implementation, I would like to know about it!

src/Spago/Watch.hs
ozkutuk: What would be a good way to unify these though? To be honest I pretty like the idea of `HasGit` constraint. However as you mentioned before, I can't think of a way to make it work with a CLI flag right now. What could be a potential way to unify these, without ditching the one in `Spago.Git`? 🤔

src/Spago/CLI.hs
klntsky: Github UI allows to hide whitespace changes in diffs: https://github.blog/2018-05-01-ignore-white-space-in-code-review/

f-f: Oh wonderful thanks, I didn't know that. Nevermind then! :slightly_smiling_face:

src/Spago/Watch.hs
f-f: Let's move this one to the `Spago.Git` module, rename it `unsafeIsIgnored`, have `Spago.Git.isIgnored` just be an alias to it, and add a `FIXME` comment to remember to remove the unsafe variant. I have some ideas how to make things tidy (e.g. port the codebase to `generic-lens` so that we can just pass in constraints dynamically without having to use custom typeclasses), but it's kind of a big refactoring so this is probably the quickest way to be done with this.

src/Spago/Watch.hs
f-f: Oh great point. @klntsky any idea on this? :thinking:

PR #668, opened 2020-07-22
Supply --run argument to alt backends

nwolverson:
### Description of the change

Add `--run Module.main` to the alt backend invocation (applies to `run` and `test`). As the main module is already resolved there is no need to answer the question in #434 about the default to `--run`, `spago` can just always supply it. Note that as per the proposed API on #434, the main function is also passed, since `spago` already cares about this for JS this seems appropriate to me.

cc @andyarvanitis  @csicar @thautwarm

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Breaking changes (😱!!!):
 
 Other improvements:
 - Docs: updated package addition/overriding syntax to use `with` syntax (#661)
+- Pass main function argument to `--run` for alternate backends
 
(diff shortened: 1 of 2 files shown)


thautwarm:
I'd appreciate it.
Previously, we noticed this and treated it by adding a `main-module` field to our backend-specific config file.
The new change can reduce the complexity of alt backend CLI and improve the consistency of different backends.

csicar:
Thank's for implementing this :tada: 
BTW: pskt should support this as well

PR #671, opened 2020-07-24
Allow upgrade of alternative package sets

nwolverson:
### Description of the change

Allow upgrade via `spago upgrade-set` when the package set is other than `purescript/package-sets`.

Motivation:
* Alternate backends which use an alternative package set (https://github.com/purerl/package-sets)
* Organisation-private package sets

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ New features:
 Other improvements:
 - Docs: updated package addition/overriding syntax to use `with` syntax (#661)
 - Pass main function argument to `--run` for alternate backends
+- Support `upgrade set` for alternative package-set repositories
 
 ## [0.15.3] - 2020-06-15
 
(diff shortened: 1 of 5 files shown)


nwolverson:
Thanks, I'll add it to the changelog - couldn't think of a straightforward way of adding a test as the non-alt version uses `init` which of course sets up for the wrong repo

f-f:
You mean that it points to the default package set? In tests we already [point to a custom release](https://github.com/purescript/spago/blob/357dcbd9fc023b9a4fc3282263bbc0e7c6af7c56/test/SpagoSpec.hs#L574), so I guess you could point to a release of the Erlang package set and test it? Or am I missing something?

nwolverson:
If I understand that test, it does an init which sets packages.dhall to point to the current release, overwrites that with a specific old release, does an upgrade, and checks that it is back in the starting position, ie the current release. What I could do is check that after upgrading it *has changed*, which I didn't think of when I wrote the earlier comment, but not that it is at the current release (without embedding that in the test).

f-f:
@nwolverson oh that's right, what you propose sounds great to me :+1:

PR #672, opened 2020-07-27
Ensure the existence of global cache directory

### LINKED ISSUE #667: Error after run "spago repl"
Operating System: macOS Catalina 10.15.5 

After I installed spago, I checked the version

```
$ spago version
0.15.3
```

then run the repl

```
› spago repl
spago: /Users/mrkong/.cache/spago/spago-repl-tmp-1b7495f97ae0a4d7: createDirectory: does not exist (No such file or directory)
```

then I went to `~/.cache/` find out there isn't a directory named "spago", so I `mkdir` it. Try the repl agian, is showed up.

Maybe spago should take care of the directory `~/.cache/spago` ?

ozkutuk:
### Description of the change

Fixes #667 

Some commands (i.e. `spago init`) trigger the creation of global cache directory, while some others (i.e. `spago repl`) don't. With this PR, the existence of global cache directory is checked for every Spago command, and the directory is created if it doesn't exist already.

### DIFF
--- a/src/Spago/GlobalCache.hs
+++ b/src/Spago/GlobalCache.hs
@@ -149,7 +149,9 @@ getMetadata cacheFlag = do
 --   - (on Windows) the folder pointed by `LocalAppData`
 getGlobalCacheDir :: MonadIO m => m FilePath.FilePath
 getGlobalCacheDir = do
-  liftIO $ getXdgDirectory XdgCache "spago" <|> pure ".spago-global-cache"
+  globalCache <- liftIO $ getXdgDirectory XdgCache "spago" <|> pure ".spago-global-cache"
+  assertDirectory globalCache
+  pure globalCache
 
 
 -- | Fetch the tarball at `archiveUrl` and unpack it into `destination`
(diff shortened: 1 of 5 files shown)


ozkutuk:
Ideally I want to write a test to cover this, but it involves fiddling with the cache directory, which didn't seem like a good option to me. I can think of two options:
1. Write a `MonadFilesystem` class and a fake in-memory filesystem for testing. Something like the example in [this post](https://lexi-lambda.github.io/blog/2017/06/29/unit-testing-effectful-haskell-with-monad-mock/).
2. Make the cache directory configurable via a CLI option and use it to make use of a local cache directory for testing. (Although I am not sure making it configurable via a CLI flag makes sense, having the option to configure it via a configuration file seems nice to me).

Since both of these options require a bit of work, I thought we should better discuss before proceeding further. And of course just forgetting about the test is a valid option too :D

f-f:
Thanks for the PR! I think we could skip a test on this, since as you noted it's a bit fiddly - however if you want to configure the location of the global cache directory, note that it's controlled by the value of the `XDG_CACHE_DIR` environment variable

PR #675, opened 2020-08-01
Make outputs of `version` and `--version` uniform

### LINKED ISSUE #649: spago --version also prints help text
`spago version` prints the version number whereas `spago --version` prints the version number followed by three blank lines and the full help text.

Is this intended?

ozkutuk:
### Description of the change

Fixes #649 

### Checklist:

### DIFF
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -8,7 +8,6 @@ import qualified Paths_spago         as Pcli
 import           Main.Utf8           (withUtf8)
 import           Spago.CLI           (Command(..))
 
-import qualified Data.Text           as Text
 import qualified System.Environment  as Env
 import qualified Spago.Build
 import qualified Spago.GitHub
@@ -31,10 +30,19 @@ main = withUtf8 $ do
   Env.setEnv "GIT_TERMINAL_PROMPT" "0"
 
   (command, globalOptions@GlobalOptions{..}) 
-    <- CLI.optionsExt "" "" "Spago - manage your PureScript projects" (Text.pack spagoVersion) CLI.parser
+    <- CLI.options "Spago - manage your PureScript projects" CLI.parser
+
+  let handleDefault :: MonadIO m => ShowVersion -> m ()
+      handleDefault = \case
+        DoShowVersion -> CLI.echo spagoVersion *> exitSuccess
+        NoShowVersion -> pure ()
 
   Run.withEnv globalOptions $
     case command of
+
+      -- ### Commands that need no env
+      Default shouldShowVersion
+        -> handleDefault shouldShowVersion
 
       -- ### Commands that need only a basic global env
       Init force noComments
--- a/src/Spago/CLI.hs
+++ b/src/Spago/CLI.hs
@@ -166,6 +166,10 @@ parser = do
     buildOptions  = BuildOptions <$> watch <*> clearScreen <*> allowIgnored <*> sourcePaths <*> srcMapFlag <*> noInstall
                     <*> pursArgs <*> depsOnly <*> beforeCommands <*> thenCommands <*> elseCommands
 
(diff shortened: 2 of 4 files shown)


ozkutuk:
As discussed in the corresponding issue, I have took the dhall implementation as reference and finally put together something that feels good enough to me. The trick is to add a final "catch-all" default parser, and add the `--version` flag to that.

app/Spago.hs
```diff
@@ -31,11 +30,20 @@ main = withUtf8 $ do
   Env.setEnv "GIT_TERMINAL_PROMPT" "0"
 
   (command, globalOptions@GlobalOptions{..}) 
-    <- CLI.optionsExt "" "" "Spago - manage your PureScript projects" (Text.pack spagoVersion) CLI.parser
+    <- CLI.options "Spago - manage your PureScript projects" CLI.parser
+
+  let handleDefault :: MonadIO m => ShowVersion -> m ()
+      handleDefault = \case
+        DoShowVersion -> CLI.echo spagoVersion *> exitSuccess
+        NoShowVersion -> pure ()
```
f-f, line 38: I think we should show the help message and fail here?

app/Spago.hs
```diff
@@ -31,11 +30,20 @@ main = withUtf8 $ do
...
 
   Run.withEnv globalOptions $
     case command of
 
+      -- ### Commands that need no env
+      Default shouldShowVersion
+        -> handleDefault shouldShowVersion
```
f-f, line 45:
```suggestion
      Default showVersion -> handleDefault showVersion
```

src/Spago/CLI.hs
```diff
@@ -162,6 +166,8 @@ parser = do
     buildOptions  = BuildOptions <$> watch <*> clearScreen <*> allowIgnored <*> sourcePaths <*> srcMapFlag <*> noInstall
                     <*> pursArgs <*> depsOnly <*> beforeCommands <*> thenCommands <*> elseCommands
 
+    versionOpt  = Opts.flag' DoShowVersion (Opts.long "version" <> Opts.help "Show spago version")
```
f-f, line 169: Is there a reason why we don't have to pass `NoShowVersion` here?

app/Spago.hs
ozkutuk: `showVersion` clashes with the [one from `Data.Version`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Version.html#v:showVersion). I was between this choice and making `Data.Version` a qualified import. If it looks better to you, we can go with the latter :+1:

src/Spago/CLI.hs
ozkutuk: Yes. [`Opts.flag'`](https://hackage.haskell.org/package/optparse-applicative-0.15.1.0/docs/Options-Applicative.html#v:flag-39-) creates kind of a "partial" parser. That is, it does not have a default value. This is intended. We want this parser to fail if it does _not_ get a `--version` flag, rather than defaulting to `NoShowVersion`. We rely on the parser failure to show the help message. This is also the reason why we just bail out of the `handleDefault` function with `pure ()`. Because (as far as I could see) there is no way to get and propagate the help message to `handleDefault` if we successfully parse the command. So we rely on parser failure instead.

app/Spago.hs
ozkutuk: Please see the explanation below.

app/Spago.hs
f-f: Oh I see, this is fine then :+1:

src/Spago/CLI.hs
f-f: Yep, I figured something like this was happening - could you add this explanation as a comment above this line?

PR #680, opened 2020-08-14
Enable `spago init` and `spago upgrade-set` to use user-specified package-set tag rather than latest release

### LINKED ISSUE #370: Add `--tag` flag to `spago upgrade-set`
Right now doing `spago init` tries to upgrade to the latest release of the set, so in order to have a replicable setup one needs to know the right version of `package-sets` and set that by hand in the `packages.dhall`

This is uncomfortable to do from a CLI-only perspective, so we should have some way to set the right package-sets version from CLI, and I propose something like
```
$ spago upgrade-set --tag psc-0.13.2-20190725
```

This would just set the version we pass in the `packages.dhall` (that is, without doing a lookup to see if this is an actual tag in there)

The usecase for this is e.g. to make life easier for [blog posts](https://thomashoneyman.com/articles/practical-profunctor-lenses-optics/), where you really want the code there to be working for long time and there's really no space to paste the entire `packages.dhall` (cc @thomashoneyman, does this make sense?)

JordanMartinez:
### Description of the change

Fixes #370 by implementing the feature but doesn't have Spago check whether the user-specified tag is valid.

### Checklist:


src/Spago/CLI.hs
```diff
@@ -148,6 +148,7 @@ parser = do
     usePsa       = bool UsePsa NoPsa <$> CLI.switch "no-psa" 'P' "Don't build with `psa`, but use `purs`"
     openDocs     = bool NoOpenDocs DoOpenDocs <$> CLI.switch "open" 'o' "Open generated documentation in browser (for HTML format only)"
     noComments   = bool WithComments NoComments <$> CLI.switch "no-comments" 'C' "Generate package.dhall and spago.dhall files without tutorial comments"
+    tag          = CLI.optional $ CLI.optText "tag" 't' "Optional package set tag to be used instead of the latest one."
```
f-f, line 151: I wouldn't like to have a shorthand for this flag (as we found `t` to be clashing with other flags several times already), so let's use optparse-applicative directly here:

```suggestion
    tag          = CLI.optional $ Opts.strOption (Opts.long "tag" <> Opts.help "Optional package set tag to be used instead of the latest one.")
```

src/Spago/PackageSet.hs
```diff
@@ -40,6 +42,100 @@ makePackageSetFile force comments = do
...
   Dhall.format packagesPath
 
+-- | Use the specified version of the package set (if specified).
+--   Otherwise, get the latest version of the package set if possible
+modifyPackageSetVersion
+  :: forall env
+  .  (HasLogFunc env, HasGlobalCache env)
+  => Maybe Text
+  -> RIO env ()
+modifyPackageSetVersion = maybe upgradePackageSet useSpecificRelease
```
f-f, line 52:
```suggestion
updatePackageSetVersion
  :: forall env
  .  (HasLogFunc env, HasGlobalCache env)
  => Maybe Text
  -> RIO env ()
updatePackageSetVersion = maybe upgradePackageSet useSpecificRelease
```

This is a minor nit, but I feel "update" fits better what we're doing here

src/Spago/Packages.hs
```diff
@@ -35,17 +35,18 @@ import qualified Spago.Templates          as Templates
 --   - create an example `test` folder (if needed)
 initProject 
   :: (HasGlobalCache env, HasLogFunc env, HasConfigPath env)
-  => Force -> Dhall.TemplateComments 
+  => Force -> Dhall.TemplateComments -> Maybe Text.Text
```
f-f, line 38:
```suggestion
  => Force -> Dhall.TemplateComments -> Maybe Text
```

src/Spago/PackageSet.hs
JordanMartinez: I've updated the code to use that name, but I don't think one will always use it to "upgrade" the package set. One could also use `spago upgrad-set --tag olderVersion` to downgrade the package set to an earlier one.

JordanMartinez:
I've rebased the code on master and accounted for your feedback.

When renaming `modifyPackageSetVersion` to `updatePackageSetVersion`, I also renamed `updatePackageSet` to `useLatestRelease`. This would be a breaking change if anyone is using `spago` as a library. I can revert if you'd like, but I thought the name better describes what the code is doing.

JordanMartinez:
A few thoughts on this. In [your comment in #370](https://github.com/purescript/spago/issues/370#issuecomment-585730933), you list the following things that would make this done:
1. add a `--tag` flag to the `init` and `upgrade-set` commands
1. Spago patches the `packages.dhall` file with the specified tag
1. before patching, Spago verifies that a package set with that tag exists

1 and 2 have been done. How should number 3 be done, and what should my involvement with that be?

f-f:
@JordanMartinez in the issue you link I proposed that we check the GitHub releases of the repo before patching, to make sure that the tag exists. And the reason why I proposed such solution is only so that we're able to give a decent error message to the user, and prevent silent failure.
I'm not attached to any particular solution as long as we take care of this concern, so that we can introduce this feature while keeping a good UX.

Currently this are the error messages that the user gets when they input a tag that doesn't exist (this is for `upgrade-set`, but the message for `init` is similar):
```
$ spago upgrade-set --tag psc-nonexistent
[info] Changing the package set version to "psc-nonexistent"
[info] Package-set upgraded to latest tag "psc-nonexistent"
Fetching the new one and generating hashes.. (this might take some time)
spago: 
Error: Remote file not found

HTTP status code: 404

URL: https://github.com/purescript/package-sets/releases/download/psc-nonexistent/packages.dhall

Message:

1│ 
2│ 
3│ 
4│ 
5│ <!DOCTYPE html>
6│ <html lang="en">
7│   <head>
…│     <meta charset="utf-8">
```

It's good that we fail immediately _and_ do not edit the file with the wrong tag, but the message could probably be better?

I think we should merge this as is, and we can take care of error message improvements in another pull request (it could be you or someone else, and ideally it would happen before the next release)

About strategies for improving the error message:
- the above message comes from Dhall getting a 404 for the import. We _might_ be able to pattern match on this specific error and return an error message with a suggestion to "please check that the tag you passed in is correct" (though I'm not sure myself which function call actually throws this error)
- if it turns out we cannot catch the HTTP error from Dhall, we can just try to fetch the file ourselves before Dhall does that, and handle the 404 there

f-f:
@JordanMartinez re breaking changes: we don't support at all people using Spago as a library, since we don't even publish to Hackage. The current versioning system only cares about breaking changes from the CLI perspective - that's the only API that we guarantee at the moment, so it's fine to change internal things

JordanMartinez:
>  in the issue you link I proposed that we check the GitHub releases of the repo before patching, to make sure that the tag exists. And the reason why I proposed such solution is only so that we're able to give a decent error message to the user, and prevent silent failure.

Sorry, let me be clearer. I understand (and agree!) that we want a better error message and to verify that the package set actually exists before we patch `packages.dhall`. However, you described three ways we could do that: check it via GitHub, store the latest list via a release, or modify the Curator. Out of all of them, the Curator approach sounded like the approach with the best tradeoffs. However, the Curator is a separate project, so I was wondering whether I should be involved in that or whether you would work on that and then I could update this PR once that work is done.

If this PR was merged, I would want to update the message above to say "unsafely patching `packages.dhall` to the specified tag regardless of whether that tag actually exists or not" beforehand."

f-f:
@JordanMartinez 

> ..we want a better error message and to verify that the package set actually exists before we patch packages.dhall

> If this PR was merged, I would want to update the message above to say "unsafely patching packages.dhall to the specified tag regardless of whether that tag actually exists or not" beforehand."

Note that we are already throwing an error before we patch the file, which is the behaviour we want here.
I.e. this is a perfectly safe operation because if the package set file does not exists then the `packages.dhall` will not be patched.

> However, you described three ways we could do that: check it via GitHub, store the latest list via a release, or modify the Curator. Out of all of them, the Curator approach sounded like the approach with the best tradeoffs. However, the Curator is a separate project, so I was wondering whether I should be involved in that or whether you would work on that and then I could update this PR once that work is done.

I apologize for not explaining myself clearly in my previous message: I think that all the three ways I described in the past and that you quote here are overly complicated and should not be implemented, and we can get by with implementing one of the two solutions I mentioned above (i.e. catching the error that Dhall gives us or verifying ourseves that we get a 404 when fetching a non-existent package set - I can detail more if you'd like), which are dramatically simpler to implement.

JordanMartinez:
I've done some digging and here's what I've found.

First, in `Dhall.purs`'s `writeRawExpr`, the above error is thrown via `Dhall.Import.load expr`, which ultimately calls [`loadWith`](https://hackage.haskell.org/package/dhall-1.34.0/docs/Dhall-Import.html#v:loadWith)

Second, the error message we get is an [`(HTTPExceptionRequest _ ConnectionFailure)` exception](https://github.com/dhall-lang/dhall-haskell/blob/master/dhall/ghc-src/Dhall/Import/HTTP.hs#L53-L59) based on the error message it produces.

Since I'm not familiar with Haskell, can these exceptions be caught?

JordanMartinez:
Current error message:
```
spago init --tag psc-nope
[info] Current tag is different from specified tag "psc-nope"
Fetching the new one and generating hashes.. (this might take some time)
[error] Package-set tag "psc-nope" in the repo "purescript/package-sets" does not exist.
Reverting change back to "psc-0.13.8-20200724
```
The `init` command should still work, but right now the program exits. I'll fix that tomorrow.

src/Spago/Messages.hs
```diff
@@ -146,6 +146,18 @@ upgradingPackageSet newTag = makeMessage
   , "Fetching the new one and generating hashes.. (this might take some time)"
   ]
 
+changingToSpecificPackageSet :: Text -> Text
+changingToSpecificPackageSet newTag = makeMessage
+  [ "Current tag is different from specified tag " <> surroundQuote newTag
```
f-f, line 151: I think we can unify this with the previous message:
```suggestion
  [ "Updating package-sets tag to: " <> surroundQuote newTag
```

src/Spago/Messages.hs
```diff
@@ -146,6 +146,18 @@ upgradingPackageSet newTag = makeMessage
...
+  , "Fetching the new one and generating hashes.. (this might take some time)"
+  ]
+
+nonExistentPackageSet :: Text -> Text -> Text -> Text -> Text
+nonExistentPackageSet org repo oldTag newTag = makeMessage
+  [ "Package-set tag " <> surroundQuote newTag <> " in the repo " <> surroundQuote (org <> "/" <> repo) <> " does not exist."
+  , "Reverting change back to " <> surroundQuote oldTag
```
f-f, line 158: Technically we're not reverting anything here, we're just "not updating the file". The wording with "reverting" might be a bit scary, so I'd like something softer here

src/Spago/PackageSet.hs
```diff
@@ -1,5 +1,7 @@
 module Spago.PackageSet
-  ( upgradePackageSet
+  ( useLatestRelease
+  , useSpecificRelease
+  , updatePackageSetVersion
```
f-f, line 4: We're not using these anywhere so they shouldn't really be exported:

```suggestion
  ( updatePackageSetVersion
```

src/Spago/PackageSet.hs
```diff
@@ -40,6 +48,112 @@ makePackageSetFile force comments = do
...
+-- | Use the specified version of the package set (if specified).
+--   Otherwise, get the latest version of the package set if possible
+updatePackageSetVersion
+  :: forall env
+  .  (HasLogFunc env, HasGlobalCache env)
+  => Maybe Text
+  -> RIO env ()
+updatePackageSetVersion = maybe useLatestRelease useSpecificRelease
+
+useSpecificRelease
```
f-f, line 60: I feel like we could reduce code duplication a bit here, as it's exactly the same in many places so it might be really confusing to maintain - I would like to:
- add a `where` block in `updatePackageSetVersion`
- inline in that block `useSpecificRelease`, `useLatestRelease`, `updateTag`, `getCurrentTag`, `upgradeImports`.
  In this way we'd only get one definition for these subroutines. It's also a good occasion to remove the old code that has the `TODO: remove` there (i.e. we should keep the new definitions you added, not the old ones)

src/Spago/PackageSet.hs
```diff
@@ -40,6 +48,112 @@ makePackageSetFile force comments = do
...
+          | currentTag == specificTag
+            -> logDebug $ "Skipping package set version upgrade, already using specified version: " <> display quotedTag
+        Just (header, expr) -> do
+          logDebug $ "Specified tag is different than current tag"
+          let newExpr = fmap (upgradeImports org repo specificTag) expr
+          logInfo $ display $ Messages.changingToSpecificPackageSet specificTag
+          resolvedExpr <- verifyPackageSetExists newExpr
+          liftIO $ Dhall.writeRawExprPostVerify packagesPath (header, newExpr) resolvedExpr
+          -- If everything is fine, refreeze the imports
+          freeze packagesPath
```
f-f, line 97: I'd try to catch that exception simply when calling `writeExpr` here, e.g. something like this:
```suggestion
          try (liftIO $ Dhall.writeRawExpr packagesPath (header, newExpr) newExpr) >>= \case
            Right _ -> do
              -- If everything is fine, refreeze the imports
              freeze packagesPath
            Left (err :: PrettyHttpException) -> do
              -- ...rest of the code here
```

Makes sense?

f-f (commented):
Thank you for digging into this! I left some refactoring comments, but otherwise looks great :+1:

src/Spago/Messages.hs
JordanMartinez: Done!

src/Spago/Messages.hs
JordanMartinez: Good point. I've updated it to a different message now (will post soon).

src/Spago/PackageSet.hs
JordanMartinez: Done.

src/Spago/PackageSet.hs
JordanMartinez: Done.

src/Spago/PackageSet.hs
JordanMartinez: Yeah, that reads a lot better than what I had.

JordanMartinez:
I updated the code a bit before I read your messages. I only print the 'package set does not exist' message when we get a 404 specifically. Otherwise, someone might be mislead by a 5xx error.

Current output:
```
$ spago init --tag psc-none
[info] Updating package-set tag to "psc-none"
Fetching the new one and generating hashes.. (this might take some time)
[warn] Package-set tag "psc-none" in the repo "purescript/package-sets" does not exist.
Will ignore user-specified tag and continue using current tag: "psc-0.13.8-20200724"
[info] Set up a local Spago project.
[info] Try running `spago build`

$ spago upgrade-set --tag psc-please
[info] Updating package-set tag to "psc-please"
Fetching the new one and generating hashes.. (this might take some time)
[warn] Package-set tag "psc-please" in the repo "purescript/package-sets" does not exist.
Will ignore user-specified tag and continue using current tag: "psc-0.13.8-20200724"
```

JordanMartinez:
...might help if I pushed the commits I just made...

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+New features:
```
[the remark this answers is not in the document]
JordanMartinez: Should I revert that change? I feel sort of bad because you just made a breaking change...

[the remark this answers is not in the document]
JordanMartinez: I've used your message.

JordanMartinez:
I looked at the tests for a bit and wasn't quite sure how to write it. I'll likely need more guidance. We should likely add the tests here so the entire things is self-contained, but it's up to you.

JordanMartinez:
Ah, nevermind. I think I got the hang of if now.

CHANGELOG.md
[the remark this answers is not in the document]
f-f: Before 1.0 we can do all the breaking changes we need as long as they are gradual.
In this specific case people had ~1y to upgrade things, and if they still wish to do so they can go through Spago 0.16 before upgrading to this one :slightly_smiling_face:

JordanMartinez:
Besides coming to an agreement as to how certain things should be rendered, I think this PR is done!

f-f (approved):
It is done indeed! Thank you for doing this :slightly_smiling_face: 

We squash-merge as a rule, so I'll added you as a collaborator to the repo and you can reshuffle the commit message and merge yourself.

PR #684, opened 2020-09-03
Drop dependency on libtinfo

f-f:
As per title, we disable the inclusion of the `terminfo` library when compiling, which in turn removes our dependency on `libtinfo`. This should remove a whole lot of issues about `libtinfo` compatibiity across distros, and remove the _urgent_ need to have a static build.

This is a harmless change, since we inherit the dependency transitively from Dhall, which includes that for the REPL, which we don't use anywhere.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -44,3 +44,8 @@ extra-deps:
 allow-newer: true
 nix:
   packages: [zlib]
+
+# This is so we don't depend on libtinfo
+flags:
+  haskeline:
+    terminfo: false


hdgarrood (approved):
🎉

hdgarrood:
What’s the output of running `ldd` on a binary with this change, just out of interest?

f-f:
@hdgarrood 

	linux-vdso.so.1 (0x00007fff7dc7f000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f966fae9000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f966f8fb000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f966f8df000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f966f8bc000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f966f8b1000)
	libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007f966f8ac000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f966f8a4000)
	libgmp.so.10 => /lib/x86_64-linux-gnu/libgmp.so.10 (0x00007f966f823000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f966f632000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f966fc60000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f966f618000)

PR #688, opened 2020-09-20
fix: to correct value of pursArgs in findFlag in Path.hs

knight9999:
### Description of the change

This PR is to fix the bug that the user can not `spago run` command when he specifies the output directory different from `output`.

- The Issue I found

Support user wants to specify output directory, for example, `output2` directory.

The build can be done by the command
```
$ spago build -u "-o output2"
```

However, the run leads to error by the command.
```
$ spago run -u "-o output2"
```

- Reason and how to fix in this PR
 
This issue is because `findFlag` function in `Path.hs` has a problem.
If we use
```
$ spago run -u "-o output2"
```
, the `pursArgs` is `["-o output2"]`, not `["-o", "output2"]`.
 
Therefore `(x:y:xs)` can not match the `y = "output2"`.

### Checklist:

### DIFF
--- a/src/Spago/Command/Path.hs
+++ b/src/Spago/Command/Path.hs
@@ -53,11 +53,12 @@ getOutputPath buildOpts = do
 -- | Try to find the content of a certain flag in a list of PursArgs
 findFlag :: Char -> Text -> [PursArg] -> Maybe Text
 findFlag char string = \case
-  [] -> Nothing
-  [_] -> Nothing
-  (x:y:xs) -> if isFlag x
-              then Just (unPursArg y)
-              else findFlag char string (y : xs)
+  (x:xs) -> if isFlag x
+              then case Text.words (unPursArg x) of
+                (_:value:_) -> Just value
+                _ -> Nothing
+              else findFlag char string xs
+  _ -> Nothing
   where 
     isFlag :: PursArg -> Bool
     isFlag (PursArg a)
--- a/src/Spago/Command/Path.hs
+++ b/src/Spago/Command/Path.hs
@@ -1,4 +1,4 @@
-module Spago.Command.Path (showPaths, getOutputPath) where
+module Spago.Command.Path (showPaths, getOutputPath, findFlag) where
 
 import Spago.Prelude
 import Spago.Env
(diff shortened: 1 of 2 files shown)


f-f:
Hi @knight9999, thanks for contributing the fix! :clap: 

This was sort of a known bug for a while, but not enough people stumbled on the edge cases for it to be annoying. In fact you can see I left a "TODO add tests" just a few lines above your fix:
https://github.com/purescript/spago/blob/7dcd30d810d29c2bf2ae61c7947d183255b1844e/src/Spago/Command/Path.hs#L46-L51

It would be great if you could add these tests as part of this PR, as it might help catch any other bugs we might be introducing (for example, there seems to be a regression in the current tests)

knight9999:
Thanks @f-f for the valuable advice.
Now I have just added the test case and improved my PR to support `--output=something` case.

There are some test issue other than my PR,
Please test by using `Spec.hs` as

```
module Spec (spec) where

import Test.Hspec

import qualified Spago.Command.PathSpec as PathSpec

spec :: Spec
spec = do
  describe "PathSpec"     PathSpec.spec
```

instead of the origin one

```
{-# OPTIONS_GHC -F -pgmF hspec-discover -optF --module-name=Spec #-}
```

knight9999:
Thanks @f-f for the valuable advice.
Now I have just added the test case and improved my PR to support `--output=something` case.

There are some test issue other than my PR,
Please test by using `Spec.hs` as

```
module Spec (spec) where

import Test.Hspec

import qualified Spago.Command.PathSpec as PathSpec

spec :: Spec
spec = do
  describe "PathSpec"     PathSpec.spec
```

instead of the origin one

```
{-# OPTIONS_GHC -F -pgmF hspec-discover -optF --module-name=Spec #-}
```

src/Spago/Command/Path.hs
```diff
@@ -43,7 +43,7 @@ getOutputPath buildOpts = do
     Just path -> Text.unpack path
 
 
--- TODO tests:
+-- tests are followings: See test/Spago/Command/PathSpec.hs
```
f-f, line 46:
```suggestion
```

f-f (approved):
I left a small comment but otherwise this looks great :clap:

knight9999:
Hi @f-f . Thank you very much.

f-f:
Thank you :slightly_smiling_face:

PR #695, opened 2020-10-21
Move CI to GitHub Actions, upgrade to GHC 8.6.5 and Dhall 1.36

f-f:


cdepillabout:
@f-f 

From https://github.com/purescript/spago/pull/647#issuecomment-725004951:

> let me know I missed something here (i.e. do we still need patches for nixpkgs after this went in?)

Thanks for the update!

When spago is released next, I'll update it in nixpkgs.  I imagine that now it is on ghc-8.8 and using a recent version of dhall, it will be much easier to keep it up to date in nixpkgs.

f-f:
It's using the latest Dhall and GHC 8.6.5, so I imagine that if we're not already compatible with GHC 8.8 it should at least be much easier to do that.
I did actually try to upgrade to GHC 8.8, but the Windows build hangs. So I'll skip this and instead try to move to GHC 8.10 once an LTS is out for that.

Issue #709, opened 2020-12-12
add alternative name for "node-args" for other backends


joprice:
I was surprised to find that `node-args` works for any backend: https://github.com/andyarvanitis/purescript-native/issues/67. Could there be a new alias for the flag that doesn't mention node? Looking at the code, it seems that it's already represented by the type `BackendArg`, just not exposed that way.

f-f:
@joprice definitely! As you suggested we could just add a `backend-args` flag that works alongside `node-args`. This should be straightforward to implement in our CLI code, so I'll mark it as up for grabs

joprice:
I'd like to try this one myself.

joprice:
I started by building the project and running against an existing test repo of mine, and it looks like there was a backward incompatible change in dhall parsing:

```
spago:
↳ ./spago.dhall
  ↳ ./../packages.dhall

Error: Invalid input

/Users/josephprice/dev/purescript-native-test/main/../packages.dhall:131:7:
    |
131 |       with parser = ./parser/spago.dhall as Location
    |       ^
unexpected 'w'
expecting ->, :, end of input, or whitespace


21│                  ../packages.dhall

/Users/josephprice/dev/purescript-native-test/main/spago.dhall:21:18
```

```dhall

  let upstream =
        https://github.com/purescript/package-sets/releases/download/psc-0.13.8-20200922/packages.dhall sha256:5edc9af74593eab8834d7e324e5868a3d258bbab75c5531d2eb770d4324a2900

  let overrides = {=}

  let additions = {=}

  in      upstream
      //  overrides
      //  additions
        with parser = ./parser/spago.dhall as Location
        with main = ./main/spago.dhall as Location
```

Adding an extra let for the merged upstream overrides and additions fixes it.

f-f:
Fixed in #710

PR #710, opened 2020-12-13
add backend args flag

joprice:
### Description of the change

This allows the more general "backend-args" to be used instead of "node-args", since there are now multiple backends and the flag is not actually specific to node. It addresses https://github.com/purescript/spago/issues/709. 

### Checklist:

### DIFF
--- a/src/Spago/CLI.hs
+++ b/src/Spago/CLI.hs
@@ -155,6 +155,7 @@ parser = do
     docsFormat  = CLI.optional $ CLI.opt Purs.parseDocsFormat "format" 'f' "Docs output format (markdown | html | etags | ctags)"
     jobsLimit   = CLI.optional (CLI.optInt "jobs" 'j' "Limit the amount of jobs that can run concurrently")
     nodeArgs         = many $ CLI.opt (Just . BackendArg) "node-args" 'a' "Argument to pass to node (run/test only)"
+    backendArgs      = many $ CLI.opt (Just . BackendArg) "backend-args" 'b' "Argument to pass to the backend (run/test only)"
     replPackageNames = many $ CLI.opt (Just . PackageName) "dependency" 'D' "Package name to add to the REPL as dependency"
     sourcePaths      = many $ CLI.opt (Just . SourcePath) "path" 'p' "Source path to include"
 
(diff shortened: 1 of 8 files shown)


src/Spago/CLI.hs
```diff
@@ -193,16 +193,19 @@ parser = do
       , Repl <$> replPackageNames <*> sourcePaths <*> pursArgs <*> depsOnly
       )
 
+    firstNonEmpty x y = if null x then y else x
+    execArgs = firstNonEmpty <$> backendArgs <*> nodeArgs
```
f-f, line 197: Why do we have to pick one? I'd expect we could just merge them?

src/Spago/CLI.hs
```diff
@@ -193,16 +193,19 @@ parser = do
...
+
     test =
       ( "test"
       , "Test the project with some module, default Test.Main"
-      , Test <$> mainModule <*> buildOptions <*> nodeArgs
+      , Test <$> mainModule <*> buildOptions <*> execArgs
```
f-f, line 202: I'm now wondering if we should now call the new flag `--exec-args` or `--run-args` rather than `--backend-args`? :thinking:

f-f (commented):
@joprice thank you for doing this, looks great! Just left some minor comments, it's almost good to go.

src/Spago/CLI.hs
joprice: I have no strong preference. I had it as merge first, but thought that would more surprising than preferring one or the other. For instance, if one script written by one person uses node-args, and another adds backend-args on top, you might expect it to override the first one, since you'd be treating them as equivalent. But, if the semantics of repeating node-args is to combine them, then having both would imply combining for the same reason.

src/Spago/CLI.hs
joprice: Yea, I like `exec-args` better. I guess my variable names betray that.

joprice:
Should I go forward with those two changes - merge instead of override and "exec-args"?

f-f:
> Should I go forward with those two changes - merge instead of override and "exec-args"?

@joprice sorry for delay, I was travelling yesterday - yes let's go forward with these!

joprice:
Updated. Let me know if I missed anything

f-f (approved):
Looks great, thank you! :slightly_smiling_face:

PR #712, opened 2020-12-19
Feature: spago script

### LINKED ISSUE #683: Feature request: spago script
I'd :heart:  to have an analog of `stack`'s script feature:
https://docs.haskellstack.org/en/stable/GUIDE/#script-interpreter
https://www.fpcomplete.com/haskell/tutorial/stack-script/#running-a-script

Example:
```sh
$ head myscript.ps
```
outputs:
```purescript
#!/usr/bin/env spago
-- spago script
import Prelude
...
```
then
```sh
$ chmod u+x myscript.ps
$ ./myscript.ps
```

colinwahl:
### Description of the change

Fixes #683 by implementing a `script` command, which can be used to run a standalone PureScript file.

To see an example, you can run this command: `spago script -d node-fs ~/path/to/script.purs`
on this file: https://gist.github.com/colinwahl/ea8f9cc1f13b0ac9e9ba80614d8aff24
to log to the console and write a text file.

This will install the `node-fs` dependency (along with the default `console`, `effect`, and `psci-support` dependencies), compile the files, and run the `main` function within the provided file.

It is similar to `spago repl`, except that it executes the given script instead of opening a repl.

### Checklist:

### Open Questions:

* How much logging should we do? 
  * Right now, installation, compilation, warnings, and info logs are included. Given that this may be used in a pipeline, we may want to suppress those logs
*  Are there any particular types of tests you'd like to see for this?
* I have a few other questions particular to the implementation, I will attach them to the code snippet.


src/Spago/Build.hs
```diff
@@ -200,33 +202,117 @@ run
...
+    let tmpDir = Text.pack dir
+    Turtle.cd (Turtle.fromText tmpDir)
+
+    config@Config { packageSet = PackageSet{..}, ..}
+      <- Packages.initProject NoForce Dhall.WithComments tag
+
+    let newConfig :: Config
+        newConfig = config
+          { Config.dependencies = dependencies <> packageDeps
+          , Config.configSourcePaths = []
```
colinwahl, line 236: By default, `initProject` creates the `src` and `test` directories, along with the `Main.purs` files within them.

This causes some issues, specifically, it is very likely that the module being run as a script is called `Main`, which will conflict with the generated `src/Main.purs`

src/Spago/Build.hs
```diff
@@ -200,33 +202,117 @@ run
...
+    Just modName -> pure modName
+
+  GlobalCache cacheDir _ <- view (the @GlobalCache)
+  currentDir <- Turtle.pwd
+  Temp.withTempDirectory cacheDir "spago-script-tmp" $ \dir -> do
+    let tmpDir = Text.pack dir
+    Turtle.cd (Turtle.fromText tmpDir)
+
+    config@Config { packageSet = PackageSet{..}, ..}
+      <- Packages.initProject NoForce Dhall.WithComments tag
```
thomashoneyman, line 231: `initProject` creates a `spago.dhall`, `packages.dhall`, `src/Main.purs`, `test/Main.purs`, `.purs-repl`, `.gitignore`, and other files as part of creating a new project. However, the `repl` command (when used outside a project) and the `script` command don't need to create any of these files. The package sets tag and the config can be handled in memory (in fact, once we read from the config for the first time we do manipulate it in memory only -- see `newConfig` below).

The `src/Main.purs` file in particular causes problems for the script command. As @colinwahl noted below, this module is included in the config sources by default and has the module name `Main`, which is very likely to cause a duplicate module compilation error with user scripts (which are also likely to be named `Main`). For this reason we edit the config (below) to strip out the sources so these files are not included in compilation.

But it seems even better to just not generate any of these files in the case we are doing a temporary project (`spago repl` outside a project or `spago script` anywhere). We didn't attempt this in this PR because it's a far-reaching change, but I wanted to suggest that we introduce some kind of `Temporary | Permanent` flag that can be used to control whether we write these files in the first place.

src/Spago/Build.hs
```diff
@@ -200,33 +202,117 @@ run
...
+
+    let newConfig :: Config
+        newConfig = config
+          { Config.dependencies = dependencies <> packageDeps
```
thomashoneyman, line 235: By default all projects include `psci-support`, `effect`, and `console`. I think it merits discussion what packages are sensible to include by default for the `script` command. The last two seem reasonable, and there's good sense in keeping consistent, but `psci-support` seems obviously unnecessary.

src/Spago/Build.hs
```diff
@@ -200,33 +202,117 @@ run
...
+          , allowIgnored = DoAllowIgnored
+          , sourcePaths = [ SourcePath absoluteModulePath ]
+          , withSourceMap = WithoutSrcMap
+          , noInstall = NoInstall
+          , pursArgs = []
+          , depsOnly = AllSources
+          , beforeCommands = []
+          , thenCommands = []
+          , elseCommands = []
+          }
```
thomashoneyman, line 264: @f-f we could use a careful look here -- we weren't sure what to set for some of these, specifically the `pursArgs` field.

src/Spago/Build.hs
```diff
@@ -200,33 +202,117 @@ run
...
+      runBackend Nothing (Just dirs) moduleName Nothing "error" buildOpts []
+
+    parseModuleName contents = do
+      Parse.PsModule name _ <- hush $ parse Parse.moduleDecl "" (encodeUtf8 contents)
+      hush $ ModuleName <$> decodeUtf8' name
+
+
+data RunDirectories = RunDirectories { tmpDir :: Text, executeDir :: FilePath }
```
thomashoneyman, line 272: We've adjusted the `runBackend` function to accommodate the case in which the project is being executed in a temporary directory. In that case we need to track the temporary directory (as it contains the `run.js` file) as well as the directory where `spago script` was invoked (as this is the context where the script should be run).

For example: you run `spago script` on your desktop and expect to write a file to the current directory. We create a temporary directory, init a project, install dependencies, and compile. At this point we can't just execute the script as usual -- we need to change back to your desktop to execute the script. Otherwise you'll write your file into the temporary directory which will then be removed when the script exits.

You can see the effects of this change propagate through `runBackend`, which uses absolute paths when `RunDirectories` is provided and otherwise behaves as it did before.

src/Spago/CLI.hs
```diff
@@ -157,6 +161,8 @@ parser = do
     nodeArgs         = many $ CLI.opt (Just . BackendArg) "node-args" 'a' "Argument to pass to node (run/test only)"
     backendArgs      = many $ CLI.opt (Just . BackendArg) "exec-args" 'b' "Argument to pass to the backend (run/test only)"
     replPackageNames = many $ CLI.opt (Just . PackageName) "dependency" 'D' "Package name to add to the REPL as dependency"
+    scriptPackageNames = many $ CLI.opt (Just . PackageName) "dependency" 'd' "Package name to add to the script as a dependency"
+    scriptSource = CLI.arg Just "source" "Source file to run as script"
```
thomashoneyman, line 165: In the example given in #683 the command used `-d` to indicate a dependency. However, the `repl` command above uses `-D` for dependencies. It seems better to be consistent and use the same for both commands -- it's `repl` that seems to be the odd one out, though, as the other commands (like `nodeArgs` and `backendArgs`) use lowercase characters.

thomashoneyman:
It's also worth noting that this PR doesn't make any attempt to do the following:

- Support installing backend-specific dependencies (ie. no installing via `npm install`)
- Support caching (ie. projects are initialized from scratch every time `spago script` is run)
- Support shebangs or comments (ie. you won't be able to specify dependencies in the .purs file or chmod +x it).

Those are all well worth doing in followup PRs, but we wanted to keep this small and get the feature in its smallest form implemented.

src/Spago/Build.hs
f-f: Sounds great. I think we can just use this in all cases - i.e. without the `Maybe` - and use absolute paths everywhere?

src/Spago/CLI.hs
f-f: Yes I think consistency between the two commands is very important. I'm fine changing the `repl` to use lowercase `d`, if there are no other flags using that letter

src/Spago/CLI.hs
```diff
@@ -321,6 +333,7 @@ parser = do
 
     projectCommands = CLI.subcommandGroup "Project commands:"
       [ initProject
+      , script
```
f-f, line 336: "project commands" are meant to be "commands to do things in a project", so it feels like this belongs more to "other commands" since you wouldn't use this in a project

f-f (commented):
@colinwahl @thomashoneyman thank you so much for doing this, looks great! There are some things to adjust and I left some comments, thank you for leaving some review pointers too

f-f:
@colinwahl to answer your questions:
1. more logging is better - we log to `stderr` so the logs shouldn't mix up with the program's output if people use the `stdout`, and people can anyways use the `quiet` flag if needed
2. a good test for this could be the one you posted in your first comment, where we run a script and then check that it writes a file or something like that. Also note that current tests are failing, I suppose because of the changes to the `runBackend` function

src/Spago/CLI.hs
colinwahl: Hmm, it looks like lowercase `d` is taken by `deps-only`. 

Since that is the case, maybe we should make both `repl` and `script` use uppercase `D`?

src/Spago/CLI.hs
colinwahl: Gotcha, thanks for clearing that up, I wasn't sure. I'll update it!

src/Spago/Build.hs
```diff
@@ -200,33 +202,117 @@ run
...
+    config@Config { packageSet = PackageSet{..}, ..}
+      <- Packages.initProject NoForce Dhall.WithComments tag
+
+    let newConfig :: Config
+        newConfig = config
+          { Config.dependencies = dependencies <> packageDeps
+          , Config.configSourcePaths = []
+          }
+
+    Run.withInstallEnv' (Just newConfig) installAction
```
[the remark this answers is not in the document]
colinwahl: Great!

src/Spago/Build.hs
```diff
@@ -200,33 +202,117 @@ run
...
+          , pursArgs = []
+          , depsOnly = AllSources
+          , beforeCommands = []
+          , thenCommands = []
+          , elseCommands = []
+          }
+      runBackend Nothing (Just dirs) moduleName Nothing "error" buildOpts []
+
+    parseModuleName contents = do
+      Parse.PsModule name _ <- hush $ parse Parse.moduleDecl "" (encodeUtf8 contents)
```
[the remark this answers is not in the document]
colinwahl: Okay!

One reason @thomashoneyman and I wanted to include this was as a sort of sanity check that we have indeed been given a PureScript file so we fail fast before installing things.

Do you think we should still try to do something in that vein? If not, I agree we can assume that the module will be `Main` for now.

src/Spago/CLI.hs
f-f: @colinwahl sounds great!

src/Spago/Build.hs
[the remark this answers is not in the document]
f-f: I'd say we should not worry about this, installing things is fast and mostly cached, so the compiler will fail fast if the file it's not good

src/Spago/Build.hs
colinwahl: I was working on this, and things are straightforward except I am not sure how to get a `PackageSet` from the tag in memory.

I think I know how to construct the URL for a specific tag, and I know how to fetch the latest tag from github (when `tag = Nothing`), but I don't know how to take a `tag :: Text` and get a `PackageSet`. It looks like all the other code paths rely on there being some `packages.dhall` file in memory that they modify.

I think that I will need to adapt some functions work on a Dhall expr in memory, then use something similar to `Spago.RunEnv.getPackageSet` to get out a `PackageSet`.

src/Spago/Build.hs
colinwahl: I think `clear`, `purs-args`, `before`, `then`, and `else` make sense.

I tried to see what it would be like to enable `Watch`, and I hit problems where I think the temp directory was deleted before a second build got kicked off, and `purs compile` couldn't find the dependencies anymore. I'm not sure why...

src/Spago/Build.hs
f-f: @colinwahl maybe something like this could work?

```haskell
Dhall.inputExpr "https://raw.githubusercontent/...etc.." >>= Spago.Config.parsePackageSet
```

colinwahl:
I made some progress on this last night and updated some comments. I'll be back later today to continue working on it.

src/Spago/CLI.hs
f-f: Thinking more about this, I think it would be nicer to use the `d` letter for passing in dependencies - I think it would be fine to deprecate the shorthand for `deps-only` - allowing only the long form there

src/Spago/Build.hs
colinwahl: Yeah, something like that that works!

src/Spago/Build.hs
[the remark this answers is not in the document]
colinwahl: Will do!

src/Spago/CLI.hs
colinwahl: Sure.

src/Spago/Build.hs
f-f: @colinwahl no worries, let's give up on the `watch` for now - it's more important to ship this first, and we can revisit the `watch` later

src/Spago/Build.hs
colinwahl: As part of the changes to construct the config in memory, I went ahead and changed it to include `effect`, `console`, and anything else the user specifies as a dependency.

Does this seem reasonable to y'all?

f-f: Sounds great! I'd also add `prelude` in there by default

colinwahl:
@f-f I've run into a bit of an issue with implementing a test to check program output:

It looks like the `Compiling <module_name` logs actually get logged to `stdout`, causing some issues in my test, and possibly issues down the line when trying to pipe this output to other commands.

Do you think that is worth prioritizing now? I'm not sure how to get those logs to go somewhere else.

f-f:
@colinwahl these lines are outputted by the compiler, which uses `stdout` pre-0.14 release, but will use `stderr` as well from 0.14 on - see https://github.com/purescript/purescript/issues/3672 for more info.

0.14 is about to come out, but I'd also be fine with a crappy workaround to make these tests pass as long as we have enough comments pointing out the hack to be removed as soon as we upgrade to the new release. Makes sense?

colinwahl:
> @colinwahl these lines are outputted by the compiler, which uses `stdout` pre-0.14 release, but will use `stderr` as well from 0.14 on - see [purescript/purescript#3672](https://github.com/purescript/purescript/issues/3672) for more info.
> 
> 0.14 is about to come out, but I'd also be fine with a crappy workaround to make these tests pass as long as we have enough comments pointing out the hack to be removed as soon as we upgrade to the new release. Makes sense?

Sounds good! My workaround for now is to not have a test that checks the actual output of a script, but instead one that checks a file created by a script matches what it should. I'll put a comment about wanting some additional tests once 0.14 is out.

colinwahl:
It looks like moving to using absolute paths via `RunDirectories` in all cases caused some issues for the windows builds, in particular, it looks like we are now trying to require a module `'C:projectsspago\testspago-test-ff9151978c8dbd41/myOutput/Test.Main'` from the file `C:\\projects\\spago\\test\\spago-test-ff9151978c8dbd41\\.spago\\run.js`.

I'm not sure exactly why this is happening but it seems to be something to do with `nodeContents` and probably `fromFilePath`.

I'm going to keep investigating but I'm having a bit of trouble. @f-f @thomashoneyman If you've seen something like this before, and tips would be appreciated!

f-f:
@colinwahl in my experience trying to work with Windows absolute paths is always extremely fiddly.
Looking at the logs I have a suspicion that we should escape the backslashes inside the `require` because they are in a JS string?

If that doesn't help, according to [these docs](https://nodejs.org/api/modules.html#modules_file_modules) together with [these docs](https://nodejs.org/api/path.html#path_windows_vs_posix) it looks like we might be able to use unix-like absolute paths with `require` on Windows too

colinwahl:
> @colinwahl in my experience trying to work with Windows absolute paths is always extremely fiddly.
> Looking at the logs I have a suspicion that we should escape the backslashes inside the `require` because they are in a JS string?
> 
> If that doesn't help, according to [these docs](https://nodejs.org/api/modules.html#modules_file_modules) together with [these docs](https://nodejs.org/api/path.html#path_windows_vs_posix) it looks like we might be able to use unix-like absolute paths with `require` on Windows too

@f-f The unix-like absolute paths do work - I ended up replacing the backslashes (that were improperly escaped) with forward slashes and now things are working on windows.

colinwahl:
@f-f / @thomashoneyman 

I've fix the tests, resolved review comments & finished my checklist - I think this is ready for a final review!

README.md
```diff
@@ -401,6 +402,20 @@ $ spago repl --purs-args "--port 3200"
 ```
 
 
+### Run a standalone PureScript file as a script
+
+You can run a standalone PureScript file as a script via `spago script`.
+Note: The module name must be `Main`, and it must export an effectful function `main`.
```
f-f, line 408:
```suggestion
Note: The module name must be `Main`, and it must export a function `main :: Effect Unit`.
```

CHANGELOG.md
```diff
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Deprecate `-d` flag for `deps-only` - instead only support `--deps-only` long form
+- Switch to using `-d` for specifying dependencies within `spago repl`
+
+New Features:
+
+- Add `spago script` command. It can be used to run standalone PureScript files as scripts
+
+Other Improvements:
+
+- `spago repl` will no longer create a (mostly unused) full project skeleton in a temporary directory when a config is not found
```
f-f, line 21: Small style adjustments:
```suggestion
Breaking changes (😱!!!):
- **Deprecate `-d` flag for `deps-only` (#712)** - instead only support the `--deps-only` long form
- **Switch from `-D` to `-d` as shorthand for specifying dependencies within `spago repl` (#712)**

New features:
- Add `spago script` command. It can be used to run standalone PureScript files as scripts (#712, #683)

Other improvements:
- `spago repl` will no longer create a (mostly unused) full project skeleton in a temporary directory when a config is not found (#712)
```

README.md
```diff
@@ -401,6 +402,20 @@ $ spago repl --purs-args "--port 3200"
...
+
+By default, the following dependencies are installed: `effect`, `console`, `prelude`.
+
+You can run the script via the following:
+
+```bash
+$ spago script path/to/script.purs
+```
```
f-f, line 416:
```suggestion
You can run a script via the following, optionally specifying a package set to use, and additional dependencies to pull from there:

```bash
$ spago script --tag psc-13.8 -d node-fs path/to/script.purs
```
```

File: src/Spago/CLI.hs [hunk 13]
```diff
@@ -156,14 +160,16 @@ parser = do
...
+    dependencyPackageNames = many $ CLI.opt (Just . PackageName) "dependency" 'd' "Package name to add as a dependency"
+    scriptSource = CLI.arg Just "source" "Source file to run as script"
     sourcePaths      = many $ CLI.opt (Just . SourcePath) "path" 'p' "Source path to include"
 
     packageName     = CLI.arg (Just . PackageName) "package" "Specify a package name. You can list them with `ls packages`"
     packageNames    = many $ CLI.arg (Just . PackageName) "package" "Package name to add as dependency"
     pursArgs        = many $ CLI.opt (Just . PursArg) "purs-args" 'u' "Arguments to pass to purs compile. Wrap in quotes."
     buildOptions  = BuildOptions <$> watch <*> clearScreen <*> allowIgnored <*> sourcePaths <*> srcMapFlag <*> noInstall
                     <*> pursArgs <*> depsOnly <*> beforeCommands <*> thenCommands <*> elseCommands
+    scriptBuildOptions = ScriptBuildOptions <$> clearScreen <*> pursArgs <*> beforeCommands <*> thenCommands <*> elseCommands
```
f-f, line 172: The `clearScreen` flag is useful only when paired with the `watch` flag, so we don't need it here:

```suggestion
    scriptBuildOptions = ScriptBuildOptions <$> pursArgs <*> beforeCommands <*> thenCommands <*> elseCommands
```

src/Spago/Types.hs
```diff
@@ -147,6 +147,16 @@ data BuildOptions = BuildOptions
   , elseCommands   :: [Text]
   }
 
+-- TODO: Figure out how `Watch` would work for `spago script` and include it
+data ScriptBuildOptions = ScriptBuildOptions
+  { shouldClear    :: ClearScreen
+  , pursArgs       :: [PursArg]
```
f-f, line 153:
```suggestion
  { pursArgs       :: [PursArg]
```

f-f (approved):
@colinwahl this is almost perfect now! Really great work! :clap: :tada: 

I left only small suggestions, so once we resolve these we can merge :slightly_smiling_face:

thomashoneyman (approved):
Agree with @f-f's feedback, and I'm really excited this has come together. Great work, @colinwahl!

colinwahl:
Thanks @f-f  for the feedback & tips, it's all ready now :D

Issue #633, opened 2020-04-27
Spago should handle purescript compiler install


reactormonk:
Since purescript is pinning parts of the stdlib, it should also handle the compiler version itself. How to get it to PATH of any editor is TBD, via `direnv` might be one solution.

f-f:
I'm afraid our plate is quite too full to even think of supporting this. We'll leave for anyone that would like to write a `pursup` tool

PR #721, opened 2020-12-27
Improve the way we pick the package-set to upgrade to

### LINKED ISSUE #657: Feature request: spago init chooses most recent package set compatible with available compiler
To demonstrate the problem when running `spago init`:

```
installed `purs` version:    0.13.6
minimum package-set version: 0.13.8
There are a few ways to solve this:
- install a compatible `purs` version (i.e. in the same 'semver range' as the one in the package set)
- if the `purs` version is 'too new', you can try using `spago upgrade-set` to upgrade to the latest package set
- if you know what you're doing and you want to avoid this check, you can override the `version` of the `metadata` package in the packages.dhall:
  let overrides = { metadata = upstream.metadata // { version = "v0.13.6" } }
```

As at it happens 0.13.6 is currently the most recent version in nixpkgs. There are workarounds, but I'll argue that this is more problematic for certain cases, such as the one I have: A training server with some interactive materials, that I would like to leave up without monitoring. However, I just noticed that the exercises no longer work, since each time a user runs an exercise for Purescript, `spago init` is called under the hood.

An alternative that could also fix the problem but have distinct pros and cons might be to allow a parameter specifying the package set to be used with `spago init`.

### LINKED ISSUE #666: How can I depend on a lower version purs?
```
spago build           
[error] Oh noes! It looks like the PureScript version installed on your system is not compatible with the package-set you're using.

installed `purs` version:    0.13.6
minimum package-set version: 0.13.8
```

Getting the very latest of purs installed sounds like a bit of work, if I can instead downgrade the dependency how do I do this? I assume I need to change this line:
```
let upstream =
      https://github.com/purescript/package-sets/releases/download/psc-0.13.8-20200708/packages.dhall sha256:df5b0f1ae92d4401404344f4fb2a7a3089612c9f30066dcddf9eaea4fe780e29
```

Though it's unclear what the sha256 refers to? How do I get this value for an older release like https://github.com/purescript/package-sets/releases/tag/psc-0.13.6-20200507 ?

### LINKED ISSUE #687: Error handling in getLatestPackageSetsTag could be improved
There are a couple of slight oddities with the error handling in Spago.GitHub.getLatestPackageSetsTag.

Firstly, in getLatestRelease1 (the one which uses the GitHub API), if the request fails, the string "rel1" is prepended to the message which is logged as a warning (i.e. is visible without passing the -v flag). If you turn off your wifi / unplug your ethernet cable, delete (or temporarily rename) your global cache, and run `spago init`, you get output like this:

```
[...]
[warn] rel1Could not reach GitHub. Error:

HTTPError (HttpExceptionRequest Request {
  host                 = "api.github.com"
[...]
```

which seems unintentional. The string `rel2` is included in a debug message in getLatestRelease2, which suggests that maybe the `rel1` was intended to be a debug message too.

Secondly, the errors are printed unconditionally, which means that if they all fail you get five errors printed (or ten, with `-v`). Since it's not actually fatal if all of these attempts to fetch the latest tag file fail, perhaps we should turn all of these warning log messages into debug log messages, and also only log the last failure if they all fail? I think only logging the last failure if they all fail may have been the intention at one point anyway, judging by lines 84-86:

```haskell
    liftIO (downloadTagToCache env) >>= \case
      Left (err :: SomeException) -> logDebug $ display $ Messages.failedToReachGitHub err
      Right releaseTagName -> writeTagCache releaseTagName
```

(note that the retrying happens inside `downloadTagToCache`).

f-f:
As we prepare the 0.14 release of the compiler, it looks like we won't be able to put out a matching package set for some time - this means that people pulling in the toolchain for the first time will get the new compiler and try to use it with the 0.13.8-compatible set, which right now results in a confusing error.

I believe the "proper" fix would be #633, but that's a fairly complex change, while this PR implements a simpler, but smarter logic to pick the set version (see #657 for details) on `init` and `upgrade`.
This includes properly warning the user when they use a compiler for which we don't yet have a set out. This is the error message they'd get if trying to use the 0.13.8-compatible set with the newer purs 0.14:

```
[error] It seems that you're using a compiler version that is not supported by package-sets at the moment.
Please install one of the following versions of the compiler and try again: 
- 0.12.0
- 0.12.1
- 0.12.2
- 0.12.3
- 0.12.4
- 0.12.5
- 0.13.0
- 0.13.2
- 0.13.3
- 0.13.4
- 0.13.5
- 0.13.6
- 0.13.8
```

Fix #687, fix #657, fix #666


thomashoneyman:
Should we reverse this list so that recent compiler versions are listed first?

f-f:
@thomashoneyman excellent idea :clap:

hdgarrood (approved):

thomashoneyman (approved):

PR #724, opened 2020-12-29
Reuse the same temp folder across many script runs

f-f:
As per title, this is a followup on #712 to avoid reinstalling the same dependencies over and over when running scripts

cc @colinwahl @thomashoneyman for review


thomashoneyman (approved):
Great! This is a nice follow-up improvement

Issue #723, opened 2020-12-29
New command: `spago graph`


f-f:
It should be basically an alias for `purs graph $(spago sources)`

The only caveat here is that we should probably check that the version of `purs` is at least `0.13.8`, since `graph` shipped with that version.

colinwahl:
I had to use `purs graph` internally to implement #730 - there is now a function `Spago.Purs.graph` which handles calling out to `purs graph` and decoding the console output.

If anyone is interested in working on this, I am happy to help you, just let me know!

f-f:
I was thinking it would be cool to have both `spago graph modules` and `spago graph packages` - we should already have all the functionality for this

colinwahl:
Yeah, `spago graph modules` is pretty much what's implemented by `purs graph`, so it shouldn't be bad.

`spago graph packages` would involve resolving files to package names, and then rebuilding a graph structure using package names as the nodes. This would be a little more involved. There is code to resolve a filename to a package name here: https://github.com/purescript/spago/blob/master/src/Spago/Build.hs#L80-L108

PR #730, opened 2021-01-07
Detect module imports from transitive dependencies

### LINKED ISSUE #598: Command to check that all symbols come from declared packages
As far as I understand, with Spago and Purescript it is possible to use data types and functions coming from transitive dependencies (dependencies of declared dependencies).

This may cause problems if my declared dependency decides to remove one of its dependency which I'm using but not declaring.

To avoid such issues, it would be nice to have in Spago a command which checks that every function and every data type used in the project comes from a package declared as a dependency in `spago.dhall`. Moreover, the same command could check whether some declared dependency is actually not used by the project and could be removed.

colinwahl:
This intends to close #598

### Description of the change

Adds a step to `build` to check if a source file imports a module that is part of a transitive dependency.
In that case, we fail the build and report those packages to the user.

Also, adds a stop to `build` which warns on unused dependencies.

### Checklist:

- [x] ~Added some example of the new feature to the `README`~


src/Spago/Build.hs
```diff
@@ -63,6 +65,22 @@ build BuildOptions{..} maybePostBuild = do
   let allPsGlobs = Packages.getGlobs   deps depsOnly configSourcePaths <> sourcePaths
       allJsGlobs = Packages.getJsGlobs deps depsOnly configSourcePaths <> sourcePaths
 
+      checkImports globs = do
+        graph <- Purs.graph globs
+        case graph of
+          Left err -> die [ displayShow err ]
+          Right (ModuleGraph moduleGraph) -> do
+            let
+              getPackage = fmap PackageName . headMay . drop 1 . dropWhile (/= ".spago") . Text.split (== '/')
```
colinwahl, line 74: I'm not sure if this is the best way to reify the package name from a module path - but it's what came to mind.

I have concerns particularly about whether or not the build could technically be run from a descendent of a directory `.spago`, in which case this check wouldn't work.  @f-f do you think this is a valid concern?

src/Spago/Build.hs
```diff
@@ -82,6 +100,7 @@ build BuildOptions{..} maybePostBuild = do
               shell backendCmd empty >>= \case
                 ExitSuccess   -> pure ()
                 ExitFailure n -> die [ "Backend " <> displayShow backend <> " exited with error:" <> repr n ]
+        checkImports globs
```
colinwahl, line 103: I put this check after the compilation step so that I could ignore capturing/properly displaying lex/parse errors - I could move it up with some help on how to properly render the errors (is it as simple as logging them and then failing the build with a message?)

src/Spago/Messages.hs
```diff
@@ -228,3 +228,7 @@ globsDoNotMatchWhenWatching patterns = makeMessage $
 
 makeMessage :: [Text] -> Text
 makeMessage = Text.intercalate "\n"
+
+sourceImportsTransitiveDependency :: [Text] -> Text
+sourceImportsTransitiveDependency transitive =
+  "Source files import from transative dependency: " <> Text.intercalate ", " transitive
```
colinwahl, line 234: This error message could definitely use some work - I just put this up as a first pass to get people thinking about it.

colinwahl:
Notably left out of this is the check for unused dependencies. 

I was talking about that with @thomashoneyman and he mentioned that we may not want to fail the build on detection of unused dependencies - for instance, someone may have `debug` as a dependency, and they probably don't want to add/remove it from their dependency list when they want to debug something.

colinwahl:
Right now I've only included the import check as part of `build` - is there anywhere else we'd like to include it? (such as repl)

src/Spago/Build.hs
```diff
@@ -63,6 +65,22 @@ build BuildOptions{..} maybePostBuild = do
...
+        graph <- Purs.graph globs
+        case graph of
+          Left err -> die [ displayShow err ]
+          Right (ModuleGraph moduleGraph) -> do
+            let
+              getPackage = fmap PackageName . headMay . drop 1 . dropWhile (/= ".spago") . Text.split (== '/')
+              spagoPackages = Map.fromList $ mapMaybe (\((moduleName, ModuleGraphNode{..})) -> (moduleName,) <$> getPackage path) $ Map.toList moduleGraph
+              sourceModuleDeps = Set.fromList (depends . snd =<< filter (isNothing . getPackage . path . snd) (Map.toList moduleGraph))
+              importedPackages = Set.fromList $ mapMaybe (flip Map.lookup spagoPackages) (Set.toList sourceModuleDeps)
+              importedTransitive = importedPackages `Set.difference` Set.fromList dependencies
```
f-f, line 78: This is good, but we'll need a bit more logic for this to be correct.
First of all - as you might have noticed from the failing tests - we should classify the graph nodes in three kinds:
1. project modules: well, the modules in your project
2. direct dependency modules: nodes that are children of project modules but are not project modules themselves
3. transitive dependency modules: all the other nodes (note: this includes also nodes that might be in "direct dependency packages", but that are not directly imported in project files)

What we care about for this check is the set of (2). From that set of modules we can get a set of package names - let's call that `B`. If we call `A` the set of declared dependencies (the ones in your config), then:
- if a package is in `A` but not in `B`, then it's unused and we should emit a warning about it
- if a package is in `B` but not in `A`, then we're importing a transitive dependency and we should fail the build

src/Spago/Build.hs
f-f: I think it's fine to have this check after building, so that users can take care of compilation problems before build management problems :+1:

src/Spago/Messages.hs
f-f: Maybe something like this?
```
Some of your project files import modules from packages that are not in the direct dependencies of your project.
To fix this error add the following packages to the list of dependencies in your config:
- ...
```

f-f (commented):
Wooow, great work here @colinwahl! :clap: 

The implementation is not yet covering all the cases, so I left some comments to detail the missing parts, but we're definitely on the right path.

Another minor note: as I specified in #723 (which after this will become a piece of cake!), it's important that we skip this check if the `purs` version is lower than `0.13.8`.
I'd store the version (note that [we already have code to get it](https://github.com/purescript/spago/blob/960a310d6efca3bb40009eb06d88382e4670ccef/src/Spago/Purs.hs#L104)) inside the `PursCmd` - in this way we could figure it out at startup, [here](https://github.com/purescript/spago/blob/960a310d6efca3bb40009eb06d88382e4670ccef/src/Spago/RunEnv.hs#L156)

colinwahl:
I got quite busy and haven't had a chance to update this branch yet - I plan on having some more time this week to keep moving on this.

colinwahl:
@f-f I was running this on my work codebase and also creating a test when I realized that `purs graph` seems to be including the transitive closure of the module imports for a given module in the `depends` entry, instead of only listing the modules that are actually imported from that module.

I'm not sure if `purs graph` was meant to do that - I did some looking in https://github.com/purescript/purescript/pull/3781 and this is the behavior of `sortModules` because of https://github.com/purescript/purescript/blob/master/src/Language/PureScript/ModuleDependencies.hs#L47 

That being said, I don't think that we can implement this check with the current implementation of `purs graph` - I can't distinguish between a module importing a module or one of a modules module's imports importing a module.

Was this behavior of `purs graph` intended? If not, I can log an issue for https://github.com/purescript/purescript that we can look in to.

colinwahl:
With the release of PureScript 0.14, I can continue work on this. I hope to pick it back up soon!

f-f:
@colinwahl is there something I can do to help with this?

colinwahl:
@f-f To be honest I had lost where I was on this!

I started with the tests: I was having trouble getting the fixture file to match the output, even though the output was what I wanted. I've fixed that now.

Another thought I had: `spago init` immediately drops the user in to a project that will no longer build with this change! That is because `src/Main.purs` imports `Prelude`, but `prelude` isn't specified as a dependency. Should I update the default dependencies?

test/SpagoSpec.hs
```diff
@@ -420,6 +420,28 @@ spec = around_ setup $ do
...
+    describe "import checking" $ do
+
+      it "Spago should fail on direct project imports from transitive dependencies" $ do
+        spago ["init"] >>= shouldBeSuccess
+        rm "spago.dhall"
+        writeTextFile "spago.dhall" $ "{ name = \"check-imports\", dependencies = [\"effect\"], packages = ./packages.dhall }"
+        rm "src/Main.purs"
+        writeTextFile "src/Main.purs" "module Main where\nimport Prelude\nmain = unit"
+        rm "test/Main.purs"
+        spago ["build"]
```
colinwahl, line 432: `spago build` was spitting out all of the "Compiling X" messages, and I didn't want to capture all that in the fixtures. I edited the test to run the build once, then run it again and capture the smaller error/warning (but I'm open to changing it to call `spago build` once, and I can sit down and figure out the fixtures)

thomashoneyman: There’s also the —quiet flag; would that help?

f-f: I think the pattern "call `build` once and then call again to get small output" is generally what we use in many other tests, so that sounds good 👍

f-f:
> Should I update the default dependencies?

@colinwahl yes please! It shows that this change is needed so badly if we got the dependencies wrong even in the default project 😄😄

colinwahl:
Ah, the tests are failing because the `purs` version we use in tests in 0.13.8 - this code doesn't actually run on that version.

Should I bump the version in the tests? I don't know if there's a plan to test against multiple versions of `purs` or anything. Maybe that'd be worth it - and I could edit the test to do the version check before expecting failures

colinwahl:
In looking at the test failures I realized that they are now caused by the tests being incompatible with 0.14. Some of them are just because they now fail with the new import check, and others are because versions have changed.

Do you want me to go through them and try to?

If so I can go ahead and fix the ones that now fail because of this change and start updating the snapshot-like tests that failed because of versions.

f-f:
@colinwahl yes, we'd need to upgrade CI and tests to 0.14 - would you mind doing that in a separate PR so we keep this one smaller?

Re testing with different versions of `purs`: that might be good, but that's a bigger change that is not directly related to the change here, so I'd recommend taking that up in a subsequent PR after we get this merged

colinwahl:
👍 Ok! I'll put up a separate PR that jump bumps `purs` to 0.14.0 and fixes the tests. Then I'll come back to this

colinwahl:
#759 Updates CI to 0.14

Once that's in, I'll need to update some of the tests that now fail due to the checks implemented here, a lot of them specify one dependency and then the example code imports from transitive dependencies.

Pending that, I think this is getting close!

colinwahl:
I'm going through fixing the tests now, and I realized an unfortunate side-effect of these changes:

We will now warn that `psci-support` is unused in the following command sequence:
`spago init`
`spago build`

I don't really like that we warn on this by default, but I know that `psci-support` is included on purpose to support `psci`. 

@f-f Do you have any opinions on what to do about this?

f-f:
@colinwahl I noticed as well while trying this out on my projects - I'd say we should make an  exception for `psci-support`, because it's a really special package: it's included in all the apps but never gets imported (by the user at least). This is unfortunate, because that bit of code that computes the sets is really beautiful right now :smile: 

Note that we should take it away from the list of warnings, but still error out if the user imports it while not declaring the dependency

colinwahl:
@f-f Yeah, that sounds like the best options. I've done that and tests pass locally.

I'm going to update the README and changelog, then I think I'm happy with this!

colinwahl:
Hah, I'm not sure what's going on with appveyor, it is failing because it thinks we are running `purescript@0.8.2` :laughing:

colinwahl:
Aha! It just hit me - I think the appveyor failure is caused by me.

I changed `pursVersion` to take the `purs` command as an input - https://github.com/purescript/spago/pull/730/files#diff-f5cfb5054fe7614682b6a1be9572e2b1755df350ac60006146918c2df34afc41R131

I did this because wasn't sure if always using `purs` was correct after looking at this: https://github.com/purescript/spago/pull/730/files#diff-a5bfe0cf964f0616330d294f84914d1927f4b3fa820ce40b21b77a340ccfa47bR180

This means that if psa is installed, it's actually running `psa --version`, and what is the latest psa version? `v0.8.2`!

colinwahl:
I've backed out the change to `Spago.Purs.version` and I'm thinking that it should fix the tests

colinwahl:
Okay, I did some investigation into the build failures, and it's caused by how I'm handling `UsePsa` - looking into it also brought up another confusion that I wanted to clarify:

My confusion is with the implementation of `Spago.Purs.compile` - it uses the `pursCmd`, in this case that's `psa`. Then it calls `psa compile ...` - `psa` already is a wrapper around `purs compile`, so I think that will always complain about `purs compile: No files found using pattern: compile`

Should it handle when `psa` is available by removing the `compile` portion of the command?

The failures in the appveyor are caused because `pursCmd` is set to `psa`, and then it is trying to run `psa graph` - which isn't a command. 

I think that I'll modify `PursCmd` to be 

```
{ purs :: Text
, psa :: Maybe Text
, compilerVersion :: Version.SemVer
} deriving (Generic)
```

And if `UsePsa` is enabled, I'll properly handle calling `psa` for `Spago.Purs.compile`, else I'll use `purs`. Does that sound reasonable?

thomashoneyman:
That explains the issue with `purs compile: No files found using pattern: compile` that's been cropping up. I do think that should be fixed as you described. I can't speak for @f-f, but what you've suggested seems reasonable to me.

src/Spago/Build.hs
```diff
@@ -168,9 +235,9 @@ repl newPackages sourcePaths pursArgs depsOnly = do
           [ "The package called 'psci-support' needs to be installed for the repl to work properly."
           , "Run `spago install psci-support` to add it to your dependencies."
           ]
-      let globs = Packages.getGlobs deps depsOnly (configSourcePaths <> sourcePaths)
+      let globs = Packages.getGlobs deps depsOnly configSourcePaths <> sourcePaths
```
f-f, line 238: I am not sure why remove the parentheses here since this means a different thing?

src/Spago/Env.hs
```diff
@@ -119,6 +126,7 @@ data InstallEnv = InstallEnv
   , envConfigPath :: !ConfigPath
   , envGlobalCache :: !GlobalCache
   , envPackageSet :: !PackageSet
+  , envPursCmd :: !PursCmd
```
f-f, line 129: Adding the `purs` command to the `InstallEnv` is a bit smelly here (like, `install` should just "download packages", why do we need `purs` available for that?), so I've been wondering if there's any way to avoid it.
It looks like it's because `Spago.FetchPackage.fetchPackages` calls `PackageSet.checkPursIsUpToDate`. Maybe we should take away that check from there and just call it in all the places where `fetchPackages` is called?
I don't have a strong opinion on this though, maybe this is fine - I'm just voicing my reasoning to see if any of this makes sense to you

src/Spago/RunEnv.hs
```diff
@@ -71,19 +72,22 @@ withPackageSetEnv app = do
...
   -> RIO InstallEnv a
   -> RIO env a
-withInstallEnv' maybeConfig app = do
+withInstallEnv' maybeConfig usePsa app = do
   Env{..} <- getEnv
   envConfig@Config{..} <- case maybeConfig of
     Just c -> pure c
     Nothing -> getConfig
   let envPackageSet = packageSet
+  envPursCmd <- getPurs usePsa
```
f-f, line 84: It looks like most/all the places where we call this pass `NoPsa`, so maybe we should just have it here without having to thread it everywhere?

src/Spago/RunEnv.hs
```diff
@@ -155,13 +169,14 @@ getConfig = Config.ensureConfig >>= \case
...
-    UsePsa -> findExecutable "psa" >>= \case
-      Just _  -> pure "psa"
-      Nothing -> pure "purs"
-  PursCmd <$> findExecutableOrDie pursCandidate
+  purs <- findExecutableOrDie "purs"
+  psa <- case usePsa of
+    NoPsa -> pure Nothing
+    UsePsa -> findExecutable "psa"
+  compilerVersion <- Purs.pursVersion >>= \case
+    Left _ -> die [ "Failed to fetch purs version" ]
```
f-f, line 177: Should we print this error instead of ignoring it?

f-f (commented):
Great call on refactoring the `PursCmd` type, I think this is exactly what we need!

I had a last round of review and I left a bunch of minor comments, I think we're good to merge once we take care of them. Great work! :slightly_smiling_face:

README.md
```diff
@@ -403,6 +405,18 @@ $ spago repl --purs-args "--port 3200"
...
+### Detect unused dependencies
+
+`spago build` automatically warns you when your project contains unused dependencies.
+
+
+### Detect direct imports from transitive dependencies
+
+`spago build` will fail when your source files directly import from a transitive dependency.
+This can be help avoid confusing errors when upgrading dependencies,
+such as if one of your dependencies removes the dependency on the package you are importing.
```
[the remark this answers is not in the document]
colinwahl: Yeah, the errors tell the user exactly what to do, so it's probably ok to leave it off. I just did it to check the box in the PR template :D

src/Spago/Purs/Graph.hs
```diff
@@ -0,0 +1,14 @@
+module Spago.Purs.Graph where
```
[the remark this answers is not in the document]
colinwahl: Sure, I think I anticipated this being more complicated than it was :)

src/Spago/Prelude.hs
```diff
@@ -249,13 +249,13 @@ pretty = PrettyText.renderStrict
...
+findExecutable :: MonadIO m => String -> m (Maybe Text)
 findExecutable x =
   case OS.buildOS of
     OS.Windows -> Directory.findExecutable (x <> ".cmd") >>= \case
-      Nothing -> Directory.findExecutable x
-      success -> pure success
-    _ -> Directory.findExecutable x
+      Nothing -> fmap (fmap Text.pack) $ Directory.findExecutable x
+      success -> pure $ fmap Text.pack success
+    _ -> fmap (fmap Text.pack) $ Directory.findExecutable x
```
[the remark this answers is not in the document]
colinwahl: Yeah :sweat_smile:

src/Spago/Packages.hs
```diff
@@ -81,14 +86,28 @@ initProject force comments tag = do
...
+
 getGlobs :: [(PackageName, Package)] -> DepsOnly -> [SourcePath] -> [SourcePath]
-getGlobs deps depsOnly configSourcePaths
-  = map (\pair
-          -> SourcePath $ Text.pack $ Fetch.getLocalCacheDir pair
-          <> "/src/**/*.purs") deps
-  <> case depsOnly of
-    DepsOnly   -> []
-    AllSources -> configSourcePaths
+getGlobs deps depsOnly configSourcePaths = getGlobsSourcePaths $ getGlobs' deps depsOnly configSourcePaths
```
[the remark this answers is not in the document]
colinwahl: Sounds good.

src/Spago/Build.hs
colinwahl: Oh, good catch, I don't remember why I did that...

src/Spago/Build.hs
```diff
@@ -56,13 +59,76 @@ build
...
       allJsGlobs = Packages.getJsGlobs deps depsOnly configSourcePaths <> sourcePaths
 
+      checkImports globs = do
+        minVersion <- case Version.semver "0.14.0" of
+          Left _ -> die [ "Unable to parse min version for imports check" ]
+          Right minVersion -> pure minVersion
+        when (compilerVersion >= minVersion) $ do
+          graph <- Purs.graph globs
+          case graph of
+            Left err -> die [ displayShow err ]
```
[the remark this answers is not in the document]
colinwahl: Yeah, in general I don't think shit should happen, but logging a warning that it didn't work would be preferable to crashing.

src/Spago/Build.hs
```diff
@@ -56,13 +59,76 @@ build
...
+            Right (ModuleGraph moduleGraph) -> do
+              let
+                anyMatch :: Sys.FilePath -> [SourcePath] -> Bool
+                anyMatch path = any (\sourcePath -> Glob.match (Glob.compile (Text.unpack (unSourcePath sourcePath))) path)
```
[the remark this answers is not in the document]
colinwahl: I went back through this and clarified the variable names for what they actually do, and moved out some shared code, I think it should make it more clear, but I can also add a doc if it needs further clarification.

src/Spago/Build.hs
```diff
@@ -56,13 +59,76 @@ build
...
+                getImports :: ModuleName -> Set ModuleName
+                getImports = maybe Set.empty (Set.fromList . depends) . flip Map.lookup moduleGraph
+
+                -- All package modules that are imported from our project files
+                importedPackageModules :: Set ModuleName
+                importedPackageModules = foldMap getImports projectModules `Set.difference` Set.fromList projectModules
+
+                getPackage :: Text -> PackageName
+                getPackage path =
+                  maybe undefined id
```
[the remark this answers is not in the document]
colinwahl: Ah! We aren't - I forgot to remove this.

It's only used in a function that handles `Maybe` anyway, so I can just have this return `Maybe PackageName`

src/Spago/Env.hs
colinwahl: Right, so the reason I originally moved `purs` into `InstallEnv` was a change that I made to `pursVersion`: I made it take in a reference to `purs` https://github.com/purescript/spago/pull/730/commits/7ff3fce7c10460ab229f73150780388693da731d#diff-f5cfb5054fe7614682b6a1be9572e2b1755df350ac60006146918c2df34afc41L131

I did this because I wasn't sure if using "purs" was always correct after seeing a few places reference "purs.cmd" for Windows. I've since reverted that change, it was working before so I shouldn't change it.

Since I've reverted that change, I can go through and restore the old behavior and then `purs` doesn't have to be stored in `InstallEnv`.

I didn't realize that the install continues if it can't verify the purs version against the package set - I thought it failed. Now that I know it just doesn't check it, I think you're right, it's a bit smelly to have `purs` be in `InstallEnv`. I'll try to untangle it

src/Spago/RunEnv.hs
colinwahl: This should be better now that `PursCmd` isn't in `InstallEnv`.

src/Spago/RunEnv.hs
colinwahl: Yeah, good idea, I'll do what's done elsewhere in this file to print it.

colinwahl:
@f-f I've addressed your review comments and updated the changelog to reflect everything that got done!

f-f (approved):
@colinwahl thank you for the great work, this is perfect now! :clap: :clap: 

I added you as a collaborator, so you can merge yourself once CI is all green :blush:

colinwahl:
Thanks @f-f !

I'm not seeing permission to merge, not sure what that's about - but I'll merge once it shows up! We usually squash-and-merge, right?

f-f:
@colinwahl 
1. we only do squash-merge
2. check your email, you should find an invitation from GitHub to join the repo

colinwahl:
Oh! Duh, didn't think to check my email :D

colinwahl:
Thanks @f-f and @thomashoneyman for the help on this!

PR #739, opened 2021-02-02
Bump to dhall-1.38.0. Let dhall format pick the output encoding based on the input

cdepillabout:
### Description of the change

Dhall-1.38 has been released: https://hackage.haskell.org/package/dhall-1.38.0/

This has a change in the input format detection in the [`Dhall.Format.Format`](https://hackage.haskell.org/package/dhall-1.38.0/docs/Dhall-Format.html) datatype: https://hackage.haskell.org/package/dhall-1.38.0/changelog

> - dhall format will now preserve the character set of the formatted file by default. In other words, if the file uses ASCII punctuation then dhall format will format the file using ASCII punctuation.
> - If the file contains both ASCII and Unicode punctuation it will prefer Unicode by default
> - This is a breaking change because the Lam / Pi / Combine / CombineTypes, and Prefer constructors now take an additional argument to record which character set was used

- Version 1.37.1:

    ```haskell
    
    data Format = Format
      { chosenCharacterSet :: CharacterSet
      ,  ...
      }
    ```

- Version 1.38.0:

    ```haskell
    
    data Format = Format
      { chosenCharacterSet :: Maybe CharacterSet
      ,  ...
      }
    ```

The `spago` codebase is passing [`Dhall.Pretty.ASCII`](https://hackage.haskell.org/package/dhall-1.37.1/docs/Dhall-Pretty.html#t:CharacterSet) (which is a `CharacterSet`) to all the pretty-printing functions.  This PR updates the code to pass `Just ASCII`, which causes spago to render all dhall files in ASCII (even if the input dhall code uses unicode symbols).

I did this because passing `Nothing` (which should cause dhall format to use ASCII or Unicode based on the input file) causes one of the tests to fail.  However, at some point it might make sense to change this test and have spago also output either ascii or unicode depending on the format of the input dhall file.

### Checklist:


f-f (approved):
Looks fantastic, thank you 🙂

toastal:
Personally, I'd like to see that `Nothing` because if I use `spago upgrade-set` right now, my files get reformatted from what our project/formatter are set to.

PR #740, opened 2021-02-03
Fix finding psa on Windows (#693)

### LINKED ISSUE #693: PSA integration changed?
It seems the `psa` integration isn't working for me anymore.  My build script uses `spago build --purs-args "--stash"`.  This works using spago 0.14.0.  As soon as I update to 0.15.2, 0.15.3, or 0.16.0, I get the error:
```
> spago build --purs-args "--stash"

[info] Installation complete.
Invalid option `--stash'

Usage: purs.bin COMMAND
  The PureScript compiler and tools
[error] Failed to build.
```
Which I'm pretty sure means it's not finding `psa` anymore and trying to use purs.bin directly.  Is this a regression, or did something change about the integration with psa? I have psa installed globally and local to the project.  I also have spago installed globally and local to the project.  I've tried this with psa versions 0.7.3 and 0.8.0.

OS: Windows 10
spago version: 0.15.2-0.16.0
purescript version: 0.13.6
psa version: 0.7.3-0.8.0

Thanks!

stkb:
In RunEnv.hs:getPurs we were just using the standard Directory.findExecutable to look for `psa`, forgetting to check for `psa.cmd` too. In windows this was looking for `psa.exe`, which doesn't exist.

To help prevent this happening again, a self-written `findExecutable` function in Prelude is added, instead of just exporting
`Directory.findExecutable`. This new version always first checks for a `.cmd` version of the given executable name on Windows.

Fix #693

### DIFF
--- a/src/Spago/Prelude.hs
+++ b/src/Spago/Prelude.hs
@@ -63,7 +63,7 @@ module Spago.Prelude
   , systemStrictWithErr
   , viewShell
   , findExecutableOrDie
-  , Directory.findExecutable
+  , findExecutable
 
   -- * Other
   , Dhall.Core.throws
@@ -81,6 +81,7 @@ import qualified Data.Text.Prettyprint.Doc.Render.Text as PrettyText
 import qualified Data.Time                             as Time
 import           Dhall                                 (Text)
 import qualified Dhall.Core
+import qualified Distribution.System                   as OS
 import qualified RIO
 import qualified System.FilePath                       as FilePath
 import qualified System.IO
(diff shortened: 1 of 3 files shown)


stkb:
Waiting for CI tests. It's working for me in standard Windows/Cmd.

```
> spago -v build --purs-args "--stash"
[debug] Running `getGlobalCacheDir`
[debug] Transformed config is the same as the read one, not overwriting it
[debug] Ensuring that the package set is frozen
[debug] Running `spago build`
[debug] Getting transitive deps
[debug] Running `fetchPackages`
[debug] Checking if `purs` is up to date
[debug] Compiling with "psa"
[debug] Running command: `psa compile --stash ".spago/console/v4.4.0/src/**/*.purs" ".spago/effect/v2.0.1/src/**/*.purs" ".spago/prelude/v4.1.1/src/**/*.purs" ".spago/psci-support/v4.0.0/src/**/*.purs" "src/**/*.purs" "test/**/*.purs"`
           Src   Lib   All
Warnings   0     0     0
Errors     0     0     0
[info] Build succeeded.
```

src/Spago/Prelude.hs
```diff
@@ -246,11 +247,22 @@ pretty = PrettyText.renderStrict
...
 
+-- | Return the full path of the executable we're trying to call. On Windows we
+--   first try the `.cmd` version.
+findExecutable :: MonadIO m => String -> m (Maybe String)
+findExecutable x =
+  case OS.buildOS of
+    OS.Windows -> Directory.findExecutable (x <> ".cmd") >>= \case
+      Nothing -> Directory.findExecutable x
+      success -> pure success
+    _ -> findExecutable x
```
f-f, line 258: Should this be
```suggestion
    _ -> Directory.findExecutable x
```

otherwise this will loop forever?

f-f (approved):
Thank you @stkb!!

I left a small comment, but otherwise I think we're good to merge :rocket: 

Also CI seems to be failing because of some leftover import (we have `--pedantic` set when building PRs on GitHub Actions)

f-f:
@ntwilson would you be able to verify that this works for you? (to try out this version of Spago it should be enough to checkout the branch and running `stack install`)

ntwilson:
> @ntwilson would you be able to verify that this works for you? (to try out this version of Spago it should be enough to checkout the branch and running `stack install`)

Yes, it worked!!  Thanks, @stkb!
I only tested the global version.  Do we have any reason to expect it would work differently through `npm`/`npx`?  Is there a way to test that from a local build?

stkb:
@f-f Oops yeah missed that 🤦‍♂️. Also I wasn't seeing the redundant import error on my machine because I was still using an old Stack version. Anyway it's all looking good now 👍 

@ntwilson Good to hear and I don't see any reason why it shouldn't work if installed through npm.

f-f:
Thank you both! :slightly_smiling_face:

PR #743, opened 2021-02-04
Upgrade to GHC 8.10 and lts-17

f-f:


f-f:
Windows build seems to be hanging. Running on Windows with `stack build --ghc-options '-v3 -keep-tmp-files'` builds fine until the last module, but then hangs after these lines:
```
!!! systool:as: finished in 0.00 milliseconds, allocated 0.208 megabytes
Upsweep completely successful.
link(batch): upsweep (partially) failed OR
    Main.main not exported; not linking.   
```

f-f:
It turns out the Windows build problem was actually the funky TemplateHaskell-build-time-network-fetching, and #775 fixed it :tada:

f-f:
cc @cdepillabout: I suspect the Nixpkgs build will not need many pins in the next release

PR #745, opened 2021-02-17
Using jsdelivr.net as CDN to fetch cache metadata

imcotton:
### Description of the change

Leverage CDN distribution to boost up metadata fetching.

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+Other improvements:
+- Fetch global cache metadata file via CDN (_jsDelivr_)
+
 ## [0.19.1] - 2021-02-04
 
 Bugfixes:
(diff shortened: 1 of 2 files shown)


f-f:
Sorry, unless there are specific problems with depending on GitHub's CDN to download this file I'm unlikely to accept this change, as it would add yet another service we'd depend on.

imcotton:
My bad, I should have provided more context regarding to the change: this very domain `raw.githubusercontent.com` has long broken history been blocked in China, many developers would relay on proxy or CDN mirrors to bridging the gap for ages.

To credit **jsdelivr** as CDN alone, here is one dedicated blog post back to almost 7 years:
- https://hacks.mozilla.org/2014/03/jsdelivr-the-advanced-open-source-public-cdn/

But you're right, it is indeed involves 3rd party service in between, I then double checked that setting `http_proxy` in ENV could indirect **spago** to honor the result, I think it's Okay to discard this PR for now, your thoughts?

f-f:
@imcotton thank you for the context, this makes sense!

If the `http_proxy` works for you (note that you could also use `https_proxy` envvar to go through TLS) then we can close this PR, and I'd be happy to accept a PR to document these variables for proxies, so that other people could discover this behaviour :slightly_smiling_face:

PR #754, opened 2021-03-19
Allow verify(-set) to work with alternate backends

nwolverson:
### Description of the change

Allow verify(-set) to work with alternate backends (when run in the context of a `spago.dhall` with `backend` set).

I've been using this change to verify the `purerl` package set

### Checklist:


f-f (approved):
Looks great, thank you! :clap:

PR #759, opened 2021-03-27
Use purescript@0.14.0 in CI

colinwahl:
### Description of the change

Bumps the version of purescript used in CI to 0.14

### Checklist:

- [x] ~Added some example of the new feature to the `README`~


f-f (approved):
@colinwahl looks great! I think we'll need a couple more things to get this through:
1. CI seems to be failing because of `git` on `Windows` on GitHub Actions, and [it looks like](https://stackoverflow.com/questions/22575662/filename-too-long-in-git-for-windows) we need to activate Windows longpaths with `git config --system core.longpaths true` (we already have some git config in the setup)
2. we should patch the `release.yaml` workflow in the same way

colinwahl:
@f-f I updated `release.yaml`, but I'm not sure where you are referring to when you say `(we already have some git config in the setup)` - mind pointing me in the right direction?

f-f:
@colinwahl I meant to open a PR to your branch but it looks like GitHub allowed me to commit directly to it - so I updated the `release.yml` file with the update to the git config

f-f (approved):
This is perfect now, thank you! :heart:

f-f:
No worries, we needed to update both :smile:

colinwahl:
Thanks for fixing that @f-f!

f-f:
@colinwahl you're welcome, and thanks for the patch :slightly_smiling_face:

Issue #749, opened 2021-03-07
Colour output doesn't work in cmd on Windows


garyb:
It just prints the colour codes as text:

```
←[34m[info] ←[0mInstalling 65 dependencies.←[0m
←[34m[info] ←[0mSearching for packages cache metadata..←[0m
←[34m[info] ←[0mUnable to find packages cache metadata, downloading from GitHub..←[0m
←[34m[info] ←[0mInstalling "exists"←[0m
←[34m[info] ←[0mInstalling "distributive"←[0m
```

Perhaps related to #729, but the suggestion there is there's a regression in an upstream library, and it's been this way for as long as I've been using spago. That doesn't necessarily rule out the upstream regression, as I came to it late!

garyb:
I probably should have mentioned, this isn't true for all commands. The `[info]` highlight works correctly for builds, for example.

f-f:
>  The `[info]` highlight works correctly for builds, for example.

Oh, that's funny then - if all color was wrong then we'd know that there's something wrong with the lway we do logging, but if only _some_ coloring is wrong then I am not really sure how to even debug this. Do you have any ideas? We're using the [logging machinery from `RIO`](https://hackage.haskell.org/package/rio-0.1.20.0/docs/RIO.html#g:5)

garyb:
Ah! I think it's something that `purs` is doing that makes it work actually. Any command that invokes the compiler has colour codes working correctly after "Compiling..." messages have been printed. `spago docs` is a perfect example actually, because it does this kind of thing:

```
←[34m[info] ←[0mGenerating documentation for the project. This might take a while...←[0m
Compiling documentation for Data.NaturalTransformation
Compiling documentation for Unsafe.Coerce
Compiling documentation for Record.Unsafe
Compiling documentation for Type.Proxy
... snip ...
Documentation written to: generated-docs/html
[info] Docs generation succeeded.
[info] Making the documentation searchable...
```

So before the compiler is invoked: no colour, after, colour.

garyb:
I have an idea of something to try 🙂

f-f:
Oh I see, that makes sense. Note that we just disable ANSI codes when `TERM=win`:

https://github.com/purescript/spago/blob/72315e003857e2a6c6e5904716d440925990ec21/src/Spago/RunEnv.hs#L24-L27

..but maybe we should just disable it when running on Windows at all?

stkb:
[...]

PR #768, opened 2021-04-07
Fix colors on Windows (#749)

stkb:
Looking through the Stack source code lead me to the [`hSupportsANSIWithoutEmulation`](https://hackage.haskell.org/package/ansi-terminal-0.11/docs/System-Console-ANSI.html#v:hSupportsANSIWithoutEmulation) function. Calling this on Windows 10+ returns `Just True` but also enables the processing of the Ansi control codes as a side effect.

This change also has the added benefit that (on Windows), the ANSI codes aren't present when output is redirected to a text file.

Code is just proof-of concept; feel free to modify as you wish. Haven't tested on *nix but it's working on Windows for me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Breaking changes (😱!!!):
 
 Bugfixes:
 - Properly call `psa` to avoid warnings (#730)
+- Color output now works on Windows (#749)
 
 Other improvements:
 - `spago build` now detects and warns about unused dependencies (#730, #598)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+Bugfixes:
+- Color output now works correctly or is disabled on Windows (#768, #749)
+
+Other improvements:
+- Color output is now automatically disabled when output is redirected to a file.
+  Also respects a [`NO_COLOR`](https://no-color.org/) environment variable (#768)
+
 ## [0.20.0] - 2021-04-07
 
 Breaking changes (😱!!!):
@@ -14,7 +21,6 @@ Breaking changes (😱!!!):
 
 Bugfixes:
 - Properly call `psa` to avoid warnings (#730)
-- Color output now works on Windows (#749)
 
 Other improvements:
 - `spago build` now detects and warns about unused dependencies (#730, #598)
--- a/src/Spago/RunEnv.hs
+++ b/src/Spago/RunEnv.hs
@@ -3,6 +3,7 @@ module Spago.RunEnv where
 import Spago.Prelude
 import Spago.Env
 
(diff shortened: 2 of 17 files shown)


stkb:
Oh I was just rebasing at the same time you were :) Will wait for CI then look at any errors

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Breaking changes (😱!!!):
 
 Bugfixes:
 - Properly call `psa` to avoid warnings (#730)
+- Color output now works on Windows (#749)
```
f-f, line 17: I just cut this release, so this should go in the unreleased section

src/Spago/RunEnv.hs
```diff
@@ -20,14 +21,17 @@ import qualified Spago.Purs as Purs
...
 withEnv :: GlobalOptions -> RIO Env a -> IO a
 withEnv GlobalOptions{..} app = do
+  let logHandle = stderr
   let verbose = not globalQuiet && (globalVerbose || globalVeryVerbose)
 
   -- https://github.com/purescript/spago/issues/579
   maybeTerm <- Env.lookupEnv "TERM"
   let termDumb = maybeTerm == Just "dumb" || maybeTerm == Just "win"
-  let useColor = globalUseColor && not termDumb
+  supportsAnsi <- hSupportsANSIWithoutEmulation logHandle
```
f-f, line 30:
```suggestion
  -- Calling this on Windows 10+ returns `Just True` but also enables the processing of the Ansi control codes as a side effect
  supportsAnsi <- hSupportsANSIWithoutEmulation logHandle
```

f-f (approved):
Left small suggestions but otherwise looks great! :clap:

stkb:
@f-f Ok as shown by the failing tests, this actually produces a change in behavior. For all platforms now, if the output is directed into a file, the color codes are never included, even without the `--no-color` flag.

Is this change desired? If not, we can instead just call the `hSupportsANSIWithoutEmulation` function but ignore the output.

f-f:
Good question. It feels like the new behaviour _should_ be more correct, but I don't have a strong opinion on this.
I guess a good call could be to stay consistent with what the compiler does?

stkb:
I can't find any definitive recommendation on this, but yeah my feeling too is that logfiles should be free from the color codes.

After investigating, this is indeed what the purs compiler does (as well as respecting `TERM=dumb`). However unfortunately PSA doesn't respect either, so output from that is still colored. I guess that's an issue for PSA though...

stkb:
Yeah I'm just working on making all the tests pass

stkb:
Tests are now passing; I stripped the color codes from the relevant test fixtures.

While we were here I also came across the [`NO_COLOR`](https://no-color.org/) spec that is another way to disable colors. It seems to have some critics but was very easy to add support for so I've added that too.

f-f (approved):
This is perfect now, great work! 👏👏

garyb:
Nice work @stkb!

Issue #769, opened 2021-04-08
How to fix transitive dependency errors


milesfrain:
PSA for folks troubleshooting error messages:

With the [release of 0.20.0](https://github.com/purescript/spago/releases/tag/0.20.0), which incorporates #730, you may encounter errors like this in pre-existing projects:
```
[error] Some of your project files import modules from packages that are not in the direct dependencies of your project.
To fix this error add the following packages to the list of dependencies in your config:
- integers
- maybe
- partial
- prelude
```
The error message suggests adding these to `spago.dhall`. A simple way to accomplish this is to run the `install` command with those missing packages:
```
spago install integers maybe partial prelude
```

It might be nice to autogenerate that `spago install ...` line in the error message for convenience.

f-f:
Fixed in #770

PR #770, opened 2021-04-08
Print install command to fix missing transitive dependencies

milesfrain:
### Description of the change

Prints a `spago install` command that will likely install missing transitive dependencies.
See #769 for context.

### Checklist:

- [x] **NA** - Added some example of the new feature to the `README`
  - Not extensive

### DIFF
--- a/src/Spago/Messages.hs
+++ b/src/Spago/Messages.hs
@@ -242,3 +242,5 @@ sourceImportsTransitiveDependency transitive = makeMessage $
   [ "Some of your project files import modules from packages that are not in the direct dependencies of your project."
   , "To fix this error add the following packages to the list of dependencies in your config:"
   ] <> map (\package -> "- " <> package) transitive
+  <> [ "You may install those packages by running the following command:"
+  , "spago install " <> Text.intercalate " " transitive ]
--- a/src/Spago/Messages.hs
+++ b/src/Spago/Messages.hs
@@ -242,5 +242,5 @@ sourceImportsTransitiveDependency transitive = makeMessage $
   [ "Some of your project files import modules from packages that are not in the direct dependencies of your project."
   , "To fix this error add the following packages to the list of dependencies in your config:"
   ] <> map (\package -> "- " <> package) transitive
-  <> [ "You may install those packages by running the following command:"
+  <> [ "You may add these dependencies by running the following command:"
   , "spago install " <> Text.intercalate " " transitive ]
--- a/test/fixtures/check-direct-import-transitive-dependency.txt
+++ b/test/fixtures/check-direct-import-transitive-dependency.txt
@@ -5,4 +5,6 @@ These dependencies are unused. To fix this warning, remove the following package
 - effect[0m
 [31m[error] [0mSome of your project files import modules from packages that are not in the direct dependencies of your project.
 To fix this error add the following packages to the list of dependencies in your config:
-- prelude[0m
+- prelude
+You may install those packages by running the following command:
+spago install prelude[0m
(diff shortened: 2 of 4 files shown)


src/Spago/Messages.hs
```diff
@@ -242,3 +242,5 @@ sourceImportsTransitiveDependency transitive = makeMessage $
   [ "Some of your project files import modules from packages that are not in the direct dependencies of your project."
   , "To fix this error add the following packages to the list of dependencies in your config:"
```
milesfrain, line 243: Comment related to the `unusedDependency` block above (must expand due to this github feature gap https://github.com/isaacs/github/issues/284):

It would be nice if we could also offer the same convenience for unused dependencies here with a `spago uninstall` or `spago remove` line. I recall `uninstall`/`remove` being proposed before, but I can't find the issue for it.

test/fixtures/check-direct-import-transitive-dependency.txt
```diff
@@ -5,4 +5,6 @@ These dependencies are unused. To fix this warning, remove the following package
 - effect[0m
 [31m[error] [0mSome of your project files import modules from packages that are not in the direct dependencies of your project.
 To fix this error add the following packages to the list of dependencies in your config:
-- prelude[0m
+- prelude
+You may install those packages by running the following command:
+spago install prelude[0m
```
milesfrain, line 10: A better test would involve 2 or more missing dependencies to ensure that `intercalate` is working correctly.

src/Spago/Messages.hs
f-f: IIRC it was a thread on slack. I don't think providing a command here is as urgent as with the other error, as this is just a warning and won't fail your build

test/fixtures/check-direct-import-transitive-dependency.txt
f-f: Yep, I agree that testing with more than one dep would be very nice here

test/fixtures/check-direct-import-transitive-dependency.txt
```diff
@@ -5,4 +5,6 @@ These dependencies are unused. To fix this warning, remove the following package
 - effect[0m
 [31m[error] [0mSome of your project files import modules from packages that are not in the direct dependencies of your project.
 To fix this error add the following packages to the list of dependencies in your config:
-- prelude[0m
+- prelude
+You may install those packages by running the following command:
```
colinwahl, line 9: Do you think it would be worth changing the wording here slightly to be something like "You may add these dependencies by running the following command:"?

The reason I ask is that these dependencies are already installed at this point (because they are transitive dependencies and those all get installed, and the build would have failed if they were not), so really all you need to do is list them as a dependency.

test/fixtures/check-direct-import-transitive-dependency.txt
```diff
@@ -1,8 +1,8 @@
 purs compile: No files found using pattern: test/**/*.purs
 [34m[info] [0mBuild succeeded.[0m
-[33m[warn] [0mNone of your project files import modules from some projects that are in the direct dependencies of your project.
-These dependencies are unused. To fix this warning, remove the following packages from the list of dependencies in your config:
-- effect[0m
```
milesfrain, line 5: This is covered in the following `Spago should warn on unused dependencies` test.
Could add this back in if this is something we want to test, or perhaps make another test for both unused and missing dependencies.

f-f: Yep I'd like to keep this warning here too, I'd say we could add `effect` back

test/fixtures/check-direct-import-transitive-dependency.txt
```diff
@@ -8,4 +8,4 @@ To fix this error add the following packages to the list of dependencies in your
 - maybe
 - prelude
 You may add these dependencies by running the following command:
-spago install maybe prelude[0m
+spago install maybe prelude[0m
```
milesfrain: Don't know what non-printable characters changed here. Guess the lesson moving forward is to not rely on web-UI for conflict resolution.

f-f (approved):
This looks great now, thank you @milesfrain!

PR #771, opened 2021-04-08
Note purescript-psa testing incompatibility

milesfrain:
### Description of the change

I noticed I needed to uninstall `purescript-psa` for tests to pass. Don't know if there's another way to get this to work.

https://github.com/purescript/spago/pull/730#issuecomment-812027464 seems related.

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+Other improvements:
+- Docs: note `purescript-psa` testing incompatibility
+
 ## [0.20.0] - 2021-04-07
 
 Breaking changes (😱!!!):
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,6 +86,7 @@ $ stack test
 $ stack test --test-arguments='--match "/Spago/spago run/Spago should use exec-args"'
 ```
 
+Note that you may need to uninstall [`purescript-psa`](https://github.com/natefaubion/purescript-psa) for tests to pass.
 
 ## Merging changes
 


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+Other improvements:
+- Docs: note `purescript-psa` testing incompatibility
```
milesfrain, line 11: Hopefully this doesn't imply an incompatibility with `spago test`. Maybe this should be reworded as:
> note `purescript-psa` `stack test` incompatibility

f-f:
Tests should pass when `psa` is installed. We even have it as part of CI (on Appveyor), to verify that we don't break compatibility with it. What issue did you encounter?

stkb:
Yeah I have this too. These tests fail if you have `psa` installed, because psa outputs extra information (the table with number of errors/warnings etc):
* spago build/import checking/Spago should fail on direct project imports from transitive dependencies
* spago build/import checking/Spago should warn on unused dependencies
* spago test/Spago should test successfully

The Appveyor script doesn't encounter these because it only runs tests in the `spago run` section. The GitHub Actions scripts don't install PSA(?)

colinwahl:
Right, the github actions tests never install `psa`, and so they never exercise the tests with it enabled.

The tests that fail all fail because they have snapshots that slightly change when `psa` is enabled.

I think the best way to fix it is to either somehow detect if `psa` is enabled and check different snapshots, or see if there is a way to check stderr vs stdout etc to check the snapshots properly

f-f:
> I think the best way to fix it is to either somehow detect if psa is enabled and check different snapshots

Yep, we already do this for some tests in fact:
https://github.com/purescript/spago/blob/7dfd2236aff92e5ae4f7a4dc336b50a7e14e4f44/test/SpagoSpec.hs#L675-L677

stkb:
Can we just add `--no-psa` for those tests? All psa really does in those cases is add some noise

colinwahl:
Oh, I never saw that check for `psa`! Yeah, it shouldn't be too hard to adapt something like that and have two snapshots - one for if `psa` exists and one for if it does not.

I wonder if the GitHub Actions tests should also be extended to install `psa` and rerun tests?

f-f:
> Can we just add `--no-psa` for those tests? All psa really does in those cases is add some noise

Oh yeah makes sense, let's just go for this

f-f:
Superseded by #772

PR #773, opened 2021-04-10
Refactor the graph functionality to remove the custom module name parser

f-f:
This refactoring PR orchestrates a few things:
- we remove the custom module name parser that was introduced in #489, using the new functionality from `purs graph` instead
- we now call `purs graph` when instantiating a `BuildEnv`, so that we get the graph only once at the beginning, and have it handy anytime we are building
- the `init` command has its own module (we should eventually do this with all the commands)


src/Spago/Build.hs
```diff
@@ -254,38 +240,29 @@ repl newPackages sourcePaths pursArgs depsOnly = do
...
-  Config.Config { alternateBackend, configSourcePaths } <- view (the @Config)
-  liftIO (foldMapM (Glob.glob . Text.unpack . unSourcePath) configSourcePaths) >>= \paths -> do
-    results <- forM paths $ \path -> do
-      content <- readFileBinary path
-      pure $ Parse.checkModuleNameMatches (encodeUtf8 $ unModuleName moduleName) content
-    if or results
-      then do
+  Config.Config { alternateBackend } <- view (the @Config)
+  maybeGraph <- view (the @Graph)
+  -- We check if the test module is included in the build and spit out a nice error if so (see #383)
```
colinwahl, line 248:
```suggestion
  -- We check if the test module is included in the build and spit out a nice error if it isn't (see #383)
```

This comment doesn't quite match implementation

src/Spago/Build/Parser.hs
```diff
@@ -1,143 +0,0 @@
-module Spago.Build.Parser
```
colinwahl, line 1: :tada:

colinwahl (approved):
Look great @f-f, just saw one minor improvement.

colinwahl:
> the init command has its own module (we should eventually do this with all the commands)

Do you also intend for all of the commands in `Spago.Build` to be pulled into their own modules down the line? I could help with moving things around if I know where they should go!

f-f:
Thank you for the review @colinwahl!

> Do you also intend for all of the commands in `Spago.Build` to be pulled into their own modules down the line? I could help with moving things around if I know where they should go!

That would be great! 
So far I created such modules only to break circular dependencies, but I think ideally all the toplevel commands should have their `Spago.Command.*` module.
Something I'm still not sure about is the granularity of the effort, e.g. if every exported function in the `Spago.Build` module would go to its own module, or if that would end up being confusing and we should go for bigger modules. In any case the `Spago.Build` module might not disappear, but remain to hold common things such as `runBackend`. Or something like that :slightly_smiling_face:

PR #775, opened 2021-04-15
Update docs-search to the latest version, simplify including it in the build

f-f:
This PR:
- prepares for including `0.0.11` of `docs-search`
- removes the Template Haskell logic that makes it hard to understand which version is being included in the build, see https://github.com/purescript/purescript-docs-search/pull/46#issuecomment-820230672
- conditionally includes the right version of `docs-search` depending on the compiler version

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -79,7 +79,9 @@ jobs:
         shell: bash
 
       - name: Build
-        run: stack build --pedantic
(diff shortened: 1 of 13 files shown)


f-f:
cc @cdepillabout: this will probably require adjustments in the nixpkgs build

cdepillabout:
Thanks for letting me know!

PR #778, opened 2021-04-19
Add more useful comments in spago.dhall

### LINKED ISSUE #708: Update opening comment in spago.dhall to refer back to Spago's Readme
The `spago.dhall` file has this comment added automatically:
```
{-
Welcome to a Spago project!
-}
```

Should this comment be removed by default? It doesn't add anything besides notifying someone unfamiliar with spago and package sets in general that it's a Spago-related item. However, I feel like Spago has become more of the norm now that even that reason isn't much of a reason now.

JordanMartinez:
### Description of the change

Fixes #708

### Checklist:

### DIFF
--- a/templates/packages.dhall
+++ b/templates/packages.dhall
@@ -42,6 +42,7 @@ in  upstream
   with halogen.repo = "https://example.com/path/to/git/repo.git"
 
   with halogen-vdom.version = "v4.0.0"
+  with halogen-vdom.dependencies = ["extra-dependency"] # halogen-vdom.dependencies
 -------------------------------
 
 ### Additions
--- a/templates/packages.dhall
+++ b/templates/packages.dhall
@@ -4,13 +4,6 @@ Welcome to your new Dhall package-set!
 Below are instructions for how to edit this file for most use
 cases, so that you don't need to know Dhall to use it.
 
-## Warning: Don't Move This Top-Level Comment!
-
-Due to how `dhall format` currently works, this comment's
-instructions cannot appear near corresponding sections below
-because `dhall format` will delete the comment. However,
-it will not delete a top-level comment like this one.
-
 ## Use Cases
 
 Most will want to do one or both of these options:
--- a/templates/packages.dhall
+++ b/templates/packages.dhall
@@ -42,7 +42,7 @@ in  upstream
   with halogen.repo = "https://example.com/path/to/git/repo.git"
 
   with halogen-vdom.version = "v4.0.0"
-  with halogen-vdom.dependencies = ["extra-dependency"] # halogen-vdom.dependencies
+  with halogen-vdom.dependencies = [ "extra-dependency" ] # halogen-vdom.dependencies
 -------------------------------
 
 ### Additions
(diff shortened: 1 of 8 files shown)


templates/spago.dhall
```diff
@@ -1,6 +1,15 @@
 {-
 Welcome to a Spago project!
 You can edit this file as you like.
+
+Need help? See the following resources:
+- Spago project:        https://github.com/purescript/spago
+- Package-Sets project: https://github.com/purescript/package-sets
```
f-f, line 7:
```suggestion
```
I'd remove this link because I think it would distract more than be useful - we link to this anyways from the Spago docs in the relevant places

templates/spago.dhall
```diff
@@ -1,6 +1,15 @@
 {-
 Welcome to a Spago project!
 You can edit this file as you like.
+
+Need help? See the following resources:
+- Spago project:        https://github.com/purescript/spago
```
f-f, line 6:
```suggestion
- Spago documentation: https://github.com/purescript/spago
```

templates/spago.dhall
```diff
@@ -1,6 +1,15 @@
...
+- Package-Sets project: https://github.com/purescript/package-sets
+- Dhall language:       https://dhall-lang.org/
```
f-f, line 8:
```suggestion
- Dhall language tour: https://docs.dhall-lang.org/tutorials/Language-Tour.html
```
In the same philosophy as above, linking to the website might be a bit distracting (albeit neat because of the inline interactive examples, etc), while I suspect the main question one might have about it is "how to I use this thing", so the Language Tour might be more useful

templates/packages.dhall
```diff
@@ -49,6 +42,7 @@ in  upstream
   with halogen.repo = "https://example.com/path/to/git/repo.git"
 
   with halogen-vdom.version = "v4.0.0"
+  with halogen-vdom.dependencies = ["extra-dependency"] # halogen-vdom.dependencies
```
f-f, line 45:
```suggestion
  with halogen-vdom.dependencies = [ "extra-dependency" ] # halogen-vdom.dependencies
```

JordanMartinez:
I've addressed your comments.

f-f (approved):
Thank you, this looks great!
It looks like CI is failing because of some fixtures needing an update, but otherwise we're good to go :clap:

JordanMartinez:
When trying to build/test this locally, I ran into issues due to not having built a file for [docs-search](https://github.com/purescript/purescript-docs-search). The contributing file should be updated to clarify how to do this to make things easier for people.

JordanMartinez:
Latest commit fixes fixtures. Not sure if I got every one of them, but I think I did.

f-f:
Great, thank you!

f-f:
> The contributing file should be updated to clarify how to do this to make things easier for people

@JordanMartinez that's a great point - In the contributing file we [do instruct to run `make` before running the build](https://github.com/purescript/spago/blob/master/CONTRIBUTING.md#developing-spago), but I wonder how can this be made clearer?

JordanMartinez:
@f-f Ah... No, it's clear enough. I just missed it because I skimmed it too quickly. I saw `stack` and somehow missed the `make` before it.

PR #784, opened 2021-04-29
Implement `spago graph`

### LINKED ISSUE #723: New command: `spago graph`

f-f:
TODO

Fix #723


colinwahl:
Nice :) If you'd like I can give this a review when you're ready

f-f:
Keeping this open as a reminder to get to feature parity in spago-next

PR #790, opened 2021-05-06
Fix broken test command on older purs versions

### LINKED ISSUE #789: Invoking `spago test` does nothing (purs 0.13.8, spago 0.20.1)
Is spago 0.20.x supposed to work with PureScript 0.13.x or is it 0.14.x only? When invoking `spago test` with PureScript 0.13.8 and spago 0.20.1 nothing happens. I expect the build and test to run. I'm guessing this may somehow be related to `psa` but I'm unsure. Let me know if you need more info.

To reproduce:

Nothing happens when `spago test` invoked.
```
$ mkdir foo && cd foo
$ npm init -y
$ npm install --save purescript@0.13.8 spago@0.20.1
$ npx spago init
$ npx spago test
<<no output; no build>>
```

Expected output:
```
$ npm install --save spago@0.19.2
$ npx spago test
           Src   Lib   All
Warnings   0     0     0
Errors     0     0     0
[info] Build succeeded.
🍝
You should add some tests.
[info] Tests succeeded.
```

f-f:
Fix #789

### DIFF
--- a/src/Spago/Build.hs
+++ b/src/Spago/Build.hs
@@ -242,16 +242,18 @@ repl newPackages sourcePaths pursArgs depsOnly = do
 --   (or the provided module name) with node
 test :: HasBuildEnv env => Maybe ModuleName -> [BackendArg] -> RIO env ()
 test maybeModuleName extraArgs = do
+  logDebug "Running `Spago.Build.test`"
   let moduleName = fromMaybe (ModuleName "Test.Main") maybeModuleName
   Config.Config { alternateBackend } <- view (the @Config)
-  maybeGraph <- view (the @Graph)
+
   -- We check if the test module is included in the build and spit out a nice error if it isn't (see #383)
-  for_ maybeGraph $ \(ModuleGraph moduleMap) -> case Map.lookup moduleName moduleMap of
-    Nothing -> die [ "Module '" <> (display . unModuleName) moduleName <> "' not found! Are you including it in your build?" ]
-    Just _ -> do
-        sourceDir <- Turtle.pwd
-        let dirs = RunDirectories sourceDir sourceDir
-        runBackend alternateBackend dirs moduleName (Just "Tests succeeded.") "Tests failed: " extraArgs
+  maybeGraph <- view (the @Graph)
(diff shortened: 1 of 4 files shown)


src/Spago/Build.hs
```diff
@@ -242,16 +242,18 @@ repl newPackages sourcePaths pursArgs depsOnly = do
...
+
   -- We check if the test module is included in the build and spit out a nice error if it isn't (see #383)
-  for_ maybeGraph $ \(ModuleGraph moduleMap) -> case Map.lookup moduleName moduleMap of
-    Nothing -> die [ "Module '" <> (display . unModuleName) moduleName <> "' not found! Are you including it in your build?" ]
-    Just _ -> do
-        sourceDir <- Turtle.pwd
-        let dirs = RunDirectories sourceDir sourceDir
-        runBackend alternateBackend dirs moduleName (Just "Tests succeeded.") "Tests failed: " extraArgs
+  maybeGraph <- view (the @Graph)
+  for_ maybeGraph $ \(ModuleGraph moduleMap) -> when (isNothing $ Map.lookup moduleName moduleMap) $
```
colinwahl, line 251: Hrm, since the graph is only supported in `0.14` and up, I think we run into the risk of `moduleName` not existing, but us not catching it up front for anything below 0.14.

f-f: That's correct. I think it's a fine tradeoff: people will get the check for `purs` version older than `0.14.0` and any `spago` version, or for `spago` versions earlier than `0.20` and any `purs` version.

The demographic exposed to the "confusing error when the test module doesn't exist" is only composed by people like the bug reporter, that presumably have an existing build and are just upgrading `spago`, but do not want to yet upgrade `purs`. Since one runs into these errors only when setting up a build and getting the module name wrong, I believe the inconvenience for users is going to be minimal as this is not going to show up in practice.

We used to have a module name parser to provide this check before the graph command, and I'd not like to reintroduce that

colinwahl: Okay! Yeah, the tradeoff seems fine. I agree that avoiding adding the module name parser back in is very preferable :D

colinwahl (approved):
This should do it 👍

PR #799, opened 2021-06-25
Output additional info when failing to access cache dir

### LINKED ISSUE #751: Document how to override the location of the global cache
When trying to compile with `spago` on a linux env that does not have a `/home` directory, there is an error message `Directory "/home" is not accessible`. Is there a setting that would make it possible to use `spago` in an environment with out `/home`?

samhh:
Fixes #751.

I'm not at all experienced with exceptions in Haskell yet, let me know if there's a better way to approach this. I considered instead providing a hook in `assertDirectory` for modifying the error message but it felt like misuse of exceptions.

You can test this like so (assuming default XDG location):

```
$ chmod 400 ~/.cache/spago/
$ spago build
[error] Directory "/home/user/.cache/spago" is not accessible. Permissions {readable = True, writable = False, executable = False, searchable = False}
[error] hint: The global cache is located at $XDG_CACHE_HOME/spago. Set $XDG_CACHE_HOME to control its location.
```

### DIFF
--- a/src/Spago/GlobalCache.hs
+++ b/src/Spago/GlobalCache.hs
@@ -147,11 +147,16 @@ getMetadata = do
 --   `$XDG_CACHE_HOME`, otherwise it uses:
 --   - (on Linux/MacOS) the folder pointed by `$HOME/.cache`, or
 --   - (on Windows) the folder pointed by `LocalAppData`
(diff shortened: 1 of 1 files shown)


f-f (approved):
This looks great, thank you!

PR #803, opened 2021-07-07
README fixes

mhmdanas:
### Description of the change

Made some changes in the README. Each change's rationale is either obvious or explained by a comment.

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,6 @@ $ node .
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
-
 - [Design goals and reasons](#design-goals-and-reasons)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -634,15 +633,7 @@ Fetching the new one and generating hashes.. (this might take some time)
...
 ```
 
-If you wish to detach from tags for your package-set, you can of course point it to a
-specific commit. Just set your `upstream` to look something like this:
-
-```haskell
-let upstream =
-      https://raw.githubusercontent.com/purescript/package-sets/bd72269fec59950404a380a46e293bde34b4618f/src/packages.dhall
-```
-
```
mhmdanas, line 644: This is the same as the section below.

README.md
```diff
@@ -634,15 +633,7 @@ Fetching the new one and generating hashes.. (this might take some time)
...
-#### ... to a specific tag manually
+#### ...to a specific tag manually
```
mhmdanas, line 636: This is more consistent with the titles of the previous sections.

f-f (approved):
Thank you!

PR #809, opened 2021-08-04
Update Discord link to remove explicit invite

thomashoneyman:
### Description of the change

After some discussion with the infrastructure team, we've decided to implement a redirect from `purescript.org/chat` to the Discord invite link, so that resources that provide a link to the Discord don't have to update if anything were to happen with that particular invite link (for example, it is accidentally deactivated, or the channel moves, or the server owner leaves).

Related:
https://github.com/purescript/purescript/pull/4170


JordanMartinez (approved):
Not sure why CI isn't building, but the change is good.

PR #811, opened 2021-08-15
Implement Targets

### LINKED ISSUE #681: Feature Request: remove need for multiple `spago.dhall` config files by defining output targets in `spago.dhall`
In an attempt to solve #416, I described [in this comment](https://github.com/purescript/spago/issues/416#issuecomment-613634984) the idea of "output targets" (since `target` already refers to the output file one produces when bundling an app).

## Problem

If you want to add "developer dependencies," "test dependencies," or do a mono-repo, you have to define one `spago.dhall` config file for each output target. Hence, one may find a `spago.dhall` file, a `test.dhall` file, an `examples.dhall` file, and sometimes a `benchmark.dhall` file. 

Rather than duplicating our work, this feature would define all of those targets within a single `spago.dhall` file.

Below is an example of what `spago.dhall` would look like after this change:
```dhall
{-
Welcome to a Spago project!
-}
let packages = ./packages.dhall
let main = { packages = packages
           , dependencies = [ "dep1" ,"dep2" ]
           , sources = [ "src/**/*.purs" ]
           }
let test = { packages = packages
           , dependencies = main.dependencies # [ "spec" ]
           , sources = main.sources # [ "test/**/*.purs" ]
           }
-- let otherTarget = --
in { name = "my-project"
   , outputs =
      { main = main
      , test = test
  --  , otherTarget = otherTarget
      }
   }
```

## Unanswered Questions

1. When would we want to implement this change and how would we handle the transition process since it will be a big breaking change?
2. How should publishing change in light of this? Could we publish multiple projects from the same repo? (e.g. Halogen Hooks and it's corresponding testing library)
3. What would the command line argument look like?

For 3, we could "desugar" commands:
- `spago build` -> `spago --output main build`
- `spago run` -> `spago -o main run`
- `spago test` -> `spago -o test run`

I don't think this feature should remove the `-x config.dhall` part. Both features could exist at the same time.

JordanMartinez:
### Description of the change

Fixes #681 

Currently a WIP and I'll be gone for a week after submitting this.

### Checklist:

### DIFF
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -75,8 +75,8 @@ main = withUtf8 $ do
         $ Spago.Packages.install packageNames
       ListDeps _ jsonFlag transitiveFlag -> Run.withInstallEnv
         $ Ls.listPackages transitiveFlag jsonFlag
-      Sources _ -> Run.withInstallEnv
-        $ Spago.Packages.sources
+      Sources targetName -> Run.withInstallEnv
+        $ Spago.Packages.sources targetName
 
       -- ### Commands that need a "publish env": install env + git and bower
       BumpVersion dryRun spec -> Run.withPublishEnv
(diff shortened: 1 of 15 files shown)


JordanMartinez:
This PR should be refactored to build off of #815.

PR #814, opened 2021-08-28
Parse `dependencies` and `source` expressions of `list1 # list2`

### LINKED ISSUE #813: Failed to add dependencies. The `dependencies` field wasn't a List of Strings.

The spago.dhall is:

```dhall

in  { name = "monorepo"
    , dependencies =
          [ "aff"
          , "affjax"
          ]
        # package1.dependencies
    , packages = ./packages.dhall
    , sources =
      [ "packages/**/src/**/*.purs"
      ]
    }
```

JordanMartinez:
### Description of the change

Fixes #813

### Checklist:


JordanMartinez:
:smile: Might help if I was fixing the actual problem...

PR #815, opened 2021-08-29
Make `spago install` work when more advanced Dhall expressions are used

### LINKED ISSUE #813: Failed to add dependencies. The `dependencies` field wasn't a List of Strings.

JordanMartinez:
### Description of the change

Fixes #813

### Checklist:


JordanMartinez:
Ok. The failing test now passes. I'm not sure whether the change I made will affect other things or not.

src/Spago/Config.hs
```diff
@@ -398,6 +492,27 @@ withConfigAST transform = do
...
+withConfigAST'
+  :: (HasLogFunc env, HasConfigPath env)
+  => (Expr -> RIO env Expr) -> RIO env Bool
+withConfigAST' transform = do
+  ConfigPath path <- view (the @ConfigPath)
+  rawConfig <- liftIO $ Dhall.readRawExpr path
+  case rawConfig of
+    Nothing -> die [ display $ Messages.cannotFindConfig path ]
+    Just (header, expr) -> do
+      newExpr <- transform $ Dhall.Core.denote expr
```
JordanMartinez, line 508: This duplicates `withConfigAST` but doesn't use the `transformMOf Dhall.subExpressions``

JordanMartinez:
Looks like this PR will also need to be able to handle parsing of
- `With` - `{ a = "bar" } with a = "baz"`
- `Prefer` - `{ a = "bar" } // { b = "baz" }`
- `Field e (FieldSelection _ x _ )` - `e.x`

JordanMartinez:
Ok, the reason why this last test fails is because the root-level expression is a `Prefer (Embed ...) RecordLit`. After failing to update the `RecordLit` because it doesn't have a `dependencies` field, the code turns towards the `Embed`. However, there isn't a case for it, so it fails.

In this situation, I guess we do this?
```dhall
-- before
./spago.dhall // { sources = ["foo"] }
-- after
let config = ./spago.dhall 
in config // { sources = ["foo"], dependencies = config.dependencies # ["new"] }

**Comment by f-f** [comment]
@JordanMartinez I wouldn't like to edit user configs in such depth, so I think in this case it's fine to just fail?
I'm not sure how useful such a detailed error message is though 🤔

**Comment by JordanMartinez** [comment]
I can remove the detailed error message, but I still think this case should be supported.

**Comment by JordanMartinez** [comment]
As a sanity check, I think the returned expression should be normalized to verify that the AST modification did not produce an invalid configuration file. Then we fail if it produces an invalid file but continue if it's still valid.

**Comment by JordanMartinez** [comment]
Well, I think `Embed` shouldn't break `spago install` if it's at the root or top level expressions. I don't think what I have written works in all circumstances, but I think it does fix the more common ones.

Assuming this is good enough (and that you agree with this change), what tests should I add to verify that various expressions work correctly?

Also, what should the changelog say now since more than `ListAppend`s are supported now?

**Comment by JordanMartinez** [comment]
Latest commit is a much cleaner way to manipulate the AST and provides a much higher guarantee that we did it correctly.

**Comment by JordanMartinez** [comment]
@f-f I'm pretty confident in the latest changes I've made. I should still add some tests to verify some things work correctly, but I'm confident they'll pass the first time. 

Also, if we added targets on top of this, it would be easy to update the AST modification. We'd only need to add a new `SearchingForTargets` case in the `ExprLevel` type and then update the respective `Expr`s to account for that case.

**Comment by JordanMartinez** [comment]
This is ready for a final review.

**Comment by JordanMartinez** [comment]
I'm now wondering if this PR should also handle `addSourcePaths` and `updateName` (which are using `withConfigAST` rather than the `withRawConfigAST` I've added). I suppose that could be done in a separate PR...

**Comment by JordanMartinez** [comment]
@f-f I would like to also account for the `addSourcePaths` and `updateName` modifications.

**Comment by JordanMartinez** [comment]
Ok, so the current code does stuff like `addDependencies packages . updateName (PscPackage.name pscPackage)`, which has type `Expr -> Expr`, so it composes easily. The code I've written has type `Expr -> RIO env Expr`, and such updates cannot be composed as easily because there are warnings that are emitted in some circumstances. In other words, if I do this change here, it'll likely force me to update a lot more tests, and I don't think that should be done in this PR.

So, instead, the latest commits just make it possible for us to
1. add source paths to a config file
2. set the `name` field to something else
3. add dependencies to a config file

The current PR only uses the 3rd one because 1 and 2 are a bit awkward and I think I should wait until I get feedback.

I'll spend some time on adding more docs to each case as a way to verify that the case can occur and how.

**Comment by JordanMartinez** [comment]
Ok, NOW this is ready for review :laughing:

File: src/Spago/Config/AST.hs [hunk 2]
```diff
@@ -0,0 +1,964 @@
...
+  | WithinField
+  -- ^ This expression is the value associated with the `dependencies` field.
+  --   It must be a `List`-like structure, so immediately append the new packages
+  --   to the expression via a `ListAppend` or merge `ListLit`s together
+  --   if it's a `ListAppend`.
+  deriving (Eq, Show)
+
+data UpdateResult
+  = Updated !Expr
+  -- ^ The expression was succesfully updated with the new packages
```
f-f, line 178:
```suggestion
  -- ^ The expression was successfully updated with the new packages
```

src/Spago/Config/AST.hs
```diff
@@ -0,0 +1,964 @@
...
+  --   that corresponds to the specified de Brujin index
+  | EncounteredEmbed
+  -- ^ We encountered an `Embed` constructor. We can make only
+  --   two assumptions about it:
+  --    1. If we are `WithinField`, then this must be an
+  --       expression that produces a `List`-like structure.
+  --    2. We are NOT `WithinField`, then this must be an
+  --       expression that produces a `Record`-like Config-schema structure.
+  --       Thus, a "dependencies" key should exist. If it doesn't, the
+  --       Dhall expression is invalid (as verified by our prior normalized expression)
```
f-f, line 190: I don't understand why are we able to make these assumptions?

src/Spago/Config/AST.hs
```diff
@@ -0,0 +1,964 @@
...
+--     The expression we need to update will often be in the `expr` (i.e. after the 'in' keyword;
+--     ``let src = [ "src" ] in { ..., dependencies = [ "old" ], sources = src }`).
+--     However, we might need to update the expression associated with the bound variable name.
+--     (e.g. `let config = { dependencies = [ "foo" ] } in config`). We won't know
+--     until we have looked at what the "dependencies" key refers to before finding out that
+--     the binding for the variable name `config` is what we need to update.
+--
+-- Since this is modifying the raw AST and might produce an invalid configuration file,
+-- the returned expression should be verified to produce a valid configuration format.
+modifyRawAST' :: HasLogFunc env => Text -> AstUpdate -> Expr -> RIO env Expr
```
f-f, line 260: Also in this case I'm not sure what's the difference between `modifyRawAST` and `modifyRawAST'`, and I wonder if we're able to find a more descriptive name for this variant?

src/Spago/Config/AST.hs
```diff
@@ -0,0 +1,964 @@
...
+modifyRawAST' initialKey astMod originalExpr = do
```
f-f, line 261: Does this function need to be in `RIO`? It looks like all branches are wrapped in `pure`, so maybe we can make it pure?

src/Spago/Config/AST.hs
```diff
@@ -0,0 +1,964 @@
...
+
+                  Just (Updated newLeft) -> do
+                    pure $ Just $ Updated $ Dhall.Prefer charSet preferAnn newLeft right
+
+                  Nothing -> pure Nothing
+
+          SearchingForField _ -> do
+            -- See Dhall.Prefer's `AtRootExpression` case
+            -- but for this situation, we're trying to find a record, so we don't match against a "dependencies" field
+            mpRight <- updateExpr level right
```
f-f, line 637: What's `mp`?

test/SpagoSpec.hs
```diff
@@ -185,12 +246,23 @@ spec = around_ setup $ do
...
-      spago ["-x", "alternative1.dhall", "install", "simple-json"] >>= shouldBeSuccess
-      checkFixture "alternative1.dhall"
+      it "... alternative1.dhall" $ do
+
+        spago ["init"] >>= shouldBeSuccess
+        writeTextFile "alternative1.dhall" "./spago.dhall // {dependencies = [\"prelude\"]}"
+        spago ["-x", "alternative1.dhall", "install", "simple-json"] >>= shouldBeSuccess
+        checkFixture "alternative1.dhall"
+
+      it "... alternative3-insane.dhall qw" $ do
```
f-f, line 258: `qw`?

f-f (commented):
This is some very impressive work Jordan! :clap: :clap: :clap: 
I feel like I don't fully understand why/how this works (even after [reading your document](https://demo.hedgedoc.org/bZjNUCWKTsCcZPIUNsnW9A?view#), which maybe we could copy here in the PR body or in the commit message when we merge?) and that I'll be pretty lost when the next bug report "Spago is messing badly with my config file" comes, but I hope we can make this testable enough so that it'd be easy to debug in the future.

I left some minor comments around, and I have a few major concerns:
1. how much of the `modifyRawAST'` function is covered by tests? The implementation is _very_ involved, and I feel like we should try to make that function pure and cover every single branch with unit tests, rather than trying to cover everything with end-to-end tests
2. how will this play out once we have multiple targets? Will it need to change drastically? Or do you think we could reuse most of this?

src/Spago/Config/AST.hs
```diff
@@ -0,0 +1,964 @@
...
+        Just Dhall.RecordField { recordFieldValue } -> case recordFieldValue of
+          Dhall.TextLit (Dhall.Chunks [] txt) -> pure $ Just txt
+          other -> do
+            logWarn $ display $ "not listlit: " <> failedToAddDepsExpectedRecordKey other
+            pure Nothing
+        _ -> do
+          logWarn $ display $ "no record lit: " <> failedToAddDepsExpectedRecordKey (Dhall.RecordLit kvs)
+          pure Nothing
+      other -> do
+        logWarn $ display $ "other: " <> failedToAddDepsExpectedRecordKey other
```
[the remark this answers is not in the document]
JordanMartinez: This is left over from when I was debugging things. I can remove them.

src/Spago/Config/AST.hs
JordanMartinez: We can make it pure. If we kept it in `RIO`, then we could add a 'logDebug` to show what decisions were made in various places to debug possible errors if `spago install` fails.

src/Spago/Config/AST.hs
JordanMartinez: Typo! That's supposed to be `mb`. And yes, I'll change this to `maybeRight`.

test/SpagoSpec.hs
JordanMartinez: Left over from when I was debugging this. When I run `stack test --fast --ta="--match='some description'"` I get a parsing failure because it gets interpreted as `stack --fast --ta="--match='some' description"`, which fails because it has no idea what CLI arg `description` refers to. So, adding `qw` was an easier way to run specific tests.

JordanMartinez:
The first three commits were things I was doing before I saw your comments. The rest account for specific issues you mentioned.

However, I still haven't made `modifyRawAST'` (which was renamed to `modifyRawDhallExpression`) a pure function. Assuming this PR does get merged and there is a bug in it, logging debug output to the console could make it easier to figure out why spago failed to install new packages.

As for your questions:

> 1. how much of the `modifyRawAST'`/`modifyRawDhallExpression` function is covered by tests? The implementation is very involved, and I feel like we should try to make that function pure and cover every single branch with unit tests, rather than trying to cover everything with end-to-end tests

If by 'cover every possible branch,' you mean, 'add a test for each branch where the root expression is that branch,' then that can be done. If you mean, 'add a test for each expression that `modifyRawDhallExpression` can handle, that would be impossible.

Right now, the tests only cover these cases:
- given a normal `RecordLit` whose `dependencies` field is [one of these expressions](https://github.com/purescript/spago/pull/815/files#diff-560aba1d76fb29566dd1ccf7e6fabe4e1cc9d5c011606bfd6de21485b70fe8cfR450-R452), do we merge the list of text values into the `ListLit` rather than adding another `ListLit`?
- given a normal `RecordLit` with `dependencies` field, if we try to install packages that are already installed, does it result in a no-op?
- given a normal `RecordLit` with `dependencies` field, if we try to install packages where some are already installed, does it only install those new packages?
- given a `./spago.dhall with sources = ["src/**/*.purs"]`, does adding new packages work?
- given the complicated `alternative3.dhall` file, does adding new packages work?

> 2. how will this play out once we have multiple targets? Will it need to change drastically? Or do you think we could reuse most of this?

The changes for the `Spago.Config.AST` would be minimal. Here's the changes I would need to make:
- change out `AtRootExpression`'s `Text` argument with a `NonEmpty Text` argument (e.g. `"dependencies"` -> `"targets" :| ["dependencies"]). (I could get rid of `ExprLevel` completely and just use `[Text]`, but the named levels probably make the coder easier to read).
- make the `AddPackages` and `AddSources` constructors for the `ConfigModification` type take a `TargetName` value in addition to their current arguments.
- update a few things in `Config.hs` to pass in the `TargetName` value.

CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Don't warn on unused deps when building --deps-only. (#794)
+- Make `spago install` work on more advanced Dhall expressions (#815)
+
+  - Support `ListAppend`: `{ dependencies = list1 # [ "package"] }`
+  - Support `Prefer`: `let config = ... in config // { dependencies = config.dependencies # [ "foo" ] }`
+  - Support `With`: `let config = ... in config with dependencies = config.dependencies # [ "foo" ]`
+  - Support `Field`: `let multiConfig = ... in multiConfig.thisProjectConfig`
+  - Support `Let`: `let config = ... in config`
+  - Support `Embed`: `./spago.dhall`
+  - Support `Var`: `let deps = ... in { ..., dependencies = deps }`
```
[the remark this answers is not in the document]
JordanMartinez: Done.

test/fixtures/spago-install-append-some-success.dhall
```diff
@@ -0,0 +1,24 @@
...
+You can edit this file as you like.
+
+Need help? See the following resources:
+- Spago documentation: https://github.com/purescript/spago
+- Dhall language tour: https://docs.dhall-lang.org/tutorials/Language-Tour.html
+
+When creating a new Spago project, you can use
+`spago init --no-comments` or `spago init -C`
+to generate this file without the comments in this block.
+-}
```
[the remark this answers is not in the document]
JordanMartinez: I've dropped them.

src/Spago/Config.hs
```diff
@@ -120,30 +121,38 @@ parseConfig = do
...
-      pure Config{..}
-    _ -> case Dhall.TypeCheck.typeOf expr of
+  mbConfig <- parseConfig' expr
+  case mbConfig of
+    Just config -> pure config
+    Nothing -> case Dhall.TypeCheck.typeOf expr of
       Right e  -> throwM $ Dhall.ConfigIsNotRecord e
       Left err -> throwM err
 
+parseConfig'
```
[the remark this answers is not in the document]
JordanMartinez: Basically, to verify that adding new packages to the raw expression worked, I parse the returned expression to see if it produced a `Config` value. To ensure I'm using the same logic as the original parser, I moved the parsing logic in `parseConfig` to `parseConfig'`. I've since renamed `parseConfig'` to `parseConfigDhallExpression`. I've also added docs to `parseConfigDhallExpression`.

So, `parseConfig` will get the normalized expression from the `./spago.dhall` file and then pass that expression into `parseConfigDhallExpression` whereas `addDependencies` will get the modified Dhall expression that includes new packages, normalizes it, and passes it into `parseConfigDhallExpression` to verify that the modification produces a valid `Config` value.

src/Spago/Config.hs
```diff
@@ -398,6 +410,26 @@ withConfigAST transform = do
...
+  :: (HasLogFunc env, HasConfigPath env)
+  => (ResolvedExpr -> Expr -> RIO env Expr) -> RIO env Bool
+withRawConfigAST transform = do
+  ConfigPath path <- view (the @ConfigPath)
+  rawConfig <- liftIO $ Dhall.readRawExpr path
+  normalizedExpr <- liftIO $ Dhall.inputExpr $ "./" <> path
+  case rawConfig of
+    Nothing -> die [ display $ Messages.cannotFindConfig path ]
+    Just (header, expr) -> do
+      newExpr <- transform normalizedExpr $ Dhall.Core.denote expr
```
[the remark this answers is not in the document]
JordanMartinez: I think newtyping a tuple would be better because it clarifies that the normalized and raw expression are the same expression in two different versions. A simple tuple does not do that.

src/Spago/Config.hs
```diff
@@ -419,7 +451,27 @@ addDependencies
...
+            pure validatedExpr
+          Nothing -> do
+            logWarn "Failed to add dependencies."
+            logDebug "Raw AST modification did not produce a valid `spago.dhall` file."
+            pure expr
+
   unless configHasChanged $
     logWarn "Configuration file was not updated."
+  where
+    notInPackageSet = filter (\p -> Map.notMember p packagesDB) newPackages
```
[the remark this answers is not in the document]
JordanMartinez: Done.

src/Spago/Config/AST.hs
```diff
@@ -0,0 +1,964 @@
...
+      Nothing -> do
+        pure originalExpr
+      Just originalName
+        | originalName == newName -> do
+            pure originalExpr
+        | otherwise -> do
+            modifyRawAST' nameText (SetText (Dhall.TextLit (Dhall.Chunks [] newName))) originalExpr
+  where
+    -- | Code from https://stackoverflow.com/questions/45757839
+    nubSeq :: Ord a => Seq a -> Seq a
```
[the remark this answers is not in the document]
JordanMartinez: Done.

src/Spago/Config/AST.hs
```diff
@@ -0,0 +1,964 @@
...
+  SetName newName -> do
+    mbName <- findTextValue nameText
+    case mbName of
+      Nothing -> do
+        pure originalExpr
+      Just originalName
+        | originalName == newName -> do
+            pure originalExpr
+        | otherwise -> do
+            modifyRawAST' nameText (SetText (Dhall.TextLit (Dhall.Chunks [] newName))) originalExpr
```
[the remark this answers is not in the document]
JordanMartinez: Done.

src/Spago/Config/AST.hs
JordanMartinez: I've updated this doc comment, but here's my understanding. If I'm misunderstanding how Dhall's `normalize` function works, then perhaps I'm making a mistake here.

If we take a Dhall expression (e.g. what's found in `./spago.dhall`) and normalize it, it gets rid of all the non-literal constructors (e.g. `ListAppend`, `Prefer`, etc.) and replaces them all with literal values (e.g. `ListLit`, `RecordLit`, etc.). For example, an expression like `["foo"] # ["bar"]` becomes `["foo", "bar"]`. Similarly (and this is my understanding), an expression like `./spago.dhall with name = "foo"` where `./spago.dhall`'s file content is `{ name = "bar", ... }` will produce `{name = "foo", ... }` where `...` are the same values stored in the `./spago.dhall` file.

So, if we have an expression like this:
```dhall
{ name = "foo", dependencies = ./spago-deps.dhall }
-- which normalizes to
{ name = "foo", dependencies = ["foo"] }
```
then [here](https://github.com/purescript/spago/pull/815/files#diff-560aba1d76fb29566dd1ccf7e6fabe4e1cc9d5c011606bfd6de21485b70fe8cfR88), where we are looking up the text values for the "dependencies" key in the normalized version of the expression, [will succeed and return `Just`](https://github.com/purescript/spago/pull/815/files#diff-560aba1d76fb29566dd1ccf7e6fabe4e1cc9d5c011606bfd6de21485b70fe8cfR128) if `./spago-deps.dhall` produces a `List Text` Dhall expression and will fail otherwise.

Knowing that in advance, if I was updating the `dependencies` key, was at the `WithinField` level, and encountered an `Embed` case, then I know the `Embed` case must produce a `List Text` expression. Thus, I can wrap that expression in a `ListAppend` even though I don't know what madness may lie behind that `Embed`.

Similarly, if I have an expression like this:
```dhall
{ outer = { config = ./spago.dhall } }.outer.config
-- which normalizes to
{ name = "foo", dependencies = [], sources = [], ... }
```
then, because the normalized expression produces a `RecordLit` where the `dependencies` key stores a `List Text` (i.e. a valid `Config` dhall expression, at least for updating the `dependencies` field), here's how my `modifyRawAST'`/`modifyRawDhallExpression` would work:

1. Make a request to update the `dependencies` field by calling `modifyRawDhallExpression` with `InsertListText "new" with level `AtRootExpression "dependencies"`.
1. Found a `Field` that selects the `config` field. Before we can find a record expression and lookup its `dependencies` field, we must first find a record expression that has a `config` field. Thus, we now transition to `SearchingForField ("config" :| ["dependencies"])`)
1. Found a `Field` that selects the `outer` field. Same as before. Thus, we now transition to `SearchingForField ("outer" :| ["config", "dependencies"])`)
1. Found a `RecordLit` and looking up the `outer` field was successful. Transition to `SearchingForField ("config" :| ["dependencies"])`
1. Found a `RecordLit` and lookingup the `config` field was successful. Transition to `SearchingForField ("dependencies" :| [])`
1. Found an `Embed`. Since we're not at `WithinField` yet, then this MUST be a record expression that stores a `Config`-like expression somewhere in it that we can find by using the rest of the keys on our stack. Otherwise, the normalization step on the original expression would have returned `Nothing` rather than `Just`. Since the keys in our stack are just `"dependencies :| []`, then this `Embed` IS a `Config`-like Dhall expression.
1. Since this is a record expression, and we're updating the `dependencies` field, we need to build off of the record expression produced by the `Embed`. Thus, we will wrap it in a `Let` binding to have access to it, and override its `dependencies` value with the record expression's original `dependencies` value plus the new changes.
1. We now return `Just Updated` and rewrap all the other expressions around this one. 

Thus...
```dhall
-- start
{ outer = 
    { config = ./spago.dhall 
    } 
}.outer.config

-- final expression produced
{ outer = 
    { config = 
        let __embed = ./spago.dhall 
        in  __embed 
                with dependencies = __embed.dependencies # ["new"] 
    }
}.outer.config
```

src/Spago/Config/AST.hs
JordanMartinez: I renamed `modifyRawAST` to `modifyRawConfigExpression` and `modifyRawAST'` to `modifyRawDhallExpression`. The latter can work on any Dhall expression that produces a `RecordLit` when normalized. The former is designed to only work on `Config` dhall expressions rather than something else (e.g. `PackageSet` expressions).

src/Spago/Config/AST.hs
JordanMartinez: I haven't addressed this yet because we might want to keep this impure initially so that if there is a bug in my code, it's easier to debug as we can see the exact path taken.

JordanMartinez:
Ok. I've addressed all feedback aside from 
- adding one test for each branch as a root expression (assuming that's what you meant)
- making the AST functions pure (I'm waiting for confirmation that we _should_ do that)

src/Spago/Config/AST.hs
JordanMartinez: Done.

f-f:
@JordanMartinez great, thanks for addressing the feedback!

> If by 'cover every possible branch,' you mean, 'add a test for each branch where the root expression is that branch,' then that can be done. If you mean, 'add a test for each expression that `modifyRawDhallExpression` can handle, that would be impossible.

I meant the [CS meaning for "branch"], so something in the middle: we should have a unit test that covers every possible case handled/covered by the implementation. E.g. in the following code
```haskell
case foo of
  Foo a -> a
  Bar b -> if b then 0 else 1
```
...there are 3 branches: one for `Foo`, one for `Bar with b == True`, and one for `Bar with b == False`.

---

We can keep the function in `RIO` since does indeed make it easier to debug and add logging. I think it would still be fine to unit-test.

src/Spago/Config.hs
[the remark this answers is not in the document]
f-f: Yeah this is great!

src/Spago/Config/AST.hs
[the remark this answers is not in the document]
f-f: Thanks!

src/Spago/Config/AST.hs
f-f: Yeah sounds good, let's keep it in RIO

src/Spago/Config/AST.hs
```diff
@@ -0,0 +1,995 @@
...
+                    pure $ Just $ Updated $ Dhall.With recordExpr field t
+
+              Just (VariableName _) -> do
+                -- This can't happen because the Dhall expression is invalid. There can't be a variable name
+                -- if there isn't a let binding.
+                --    `{ ..., dependencies = ["old"] } with dependencies = x`
+                pure maybeResult
+
+              Just (Updated newUpdate) -> do
+                pure $ Just $ Updated $ Dhall.With recordExpr field newUpdate
```
JordanMartinez, line 734: While working on the tests, I found a bug! `newUpdate` should be called `newRecordExpr`. This was likely caused by a copy-paste error.

JordanMartinez: Whoops! This is the wrong one.

src/Spago/Config/AST.hs
```diff
@@ -0,0 +1,995 @@
...
+                  --     ```
+                  SetText t -> do
+                    pure $ Just $ Updated $ Dhall.With expr (key :| []) t
+
+              Just (VariableName _) -> do
+                -- `The `Just VariableName` can't happen here because this is `AtRootExpression`
+                pure maybeResult
+
+              Just (Updated newUpdate) -> do
+                pure $ Just $ Updated $ Dhall.With recordExpr field newUpdate
```
JordanMartinez, line 802: While working on the tests, I found a bug! `newUpdate` should be called `newRecordExpr`. This was likely caused by a copy-paste error.

src/Spago/Config/AST.hs
```diff
@@ -0,0 +1,995 @@
...
+            -- to
+            --  `{ config = { ..., dependencies = [ "package", "new" ] } }.config`
+            let
+              newLevel = SearchingForField (fieldSelectionLabel :| [key])
+            maybeResult <- fmap (mapUpdated (\newRecord -> Dhall.Field newRecord selection)) <$> updateExpr newLevel recordExpr
+            case maybeResult of
+              Just EncounteredEmbed -> do
+                case astMod of
+                  InsertListText additions -> do
+                    pure $ Just $ Updated $ Dhall.Field (updateListTextByWrappingLetBinding (key :| []) additions "__embed" expr) selection
```
JordanMartinez, line 528: Tests found another bug here:
```diff
- pure $ Just $ Updated $ Dhall.Field (updateListTextByWrappingLetBinding (key :| []) additions "__embed" expr) selection
+ pure $ Just $ Updated $ Dhall.Field (updateListTextByWrappingLetBinding (fieldSelectionLabel :| [key]) additions "__embed" recordExpr) selection

File: src/Spago/Config/AST.hs [hunk 19]
```diff
@@ -0,0 +1,995 @@
...
+
+                  SetText t -> do
+                    pure $ Just $ Updated $ Dhall.Field (Dhall.With expr (key :| []) t) selection
```
JordanMartinez, line 531: This also is a bug because it uses `expr` rather than `recordExpr`

JordanMartinez:
@f-f Thoughts on why the `expectedConfig == parsedConfig` check is returning false? Does order matter in any of its values?

JordanMartinez:
To answer my own question, looks like the issue is still caused by the `publishConfig` value. Below is the output when I diff the config values:
```
[info] publish config mismatch
[info] EXPECTED: Left \ESC[1;31mError\ESC[0m: Error while reading spago.dhall:

Explanation: a record is missing a required key.

The key missing is:

\8627 license

The keys in the record are:

\8627 dependencies, name, packages, sources
[info] ACTUAL: Left \ESC[1;31mError\ESC[0m: Error while reading spago.dhall:

Explanation: a record is missing a required key.

The key missing is:

\8627 license

The keys in the record are:

\8627 dependencies, name, packages, sources
[warn] Failed to add dependencies.
[warn] Configuration file was not updated.
```

JordanMartinez:
Two tests currently fail, but I'm not sure why. I think it's because the `expectedConfig` value is still not entirely correct in some circumstances.

JordanMartinez:
Note to self. The following command fixes the `STDOUT` and `STDERR` debug messages one can enable by uncommenting the two lines in the `shouldSucceed` area.
```bash
stack test --fast --ta="--match=qw" 2>&1 | sed 's/\\n/\n/g' | sed 's/\\"/"/g'
```

JordanMartinez:
CI is now passing. :tada: However, I think there's still more tests to write to ensure this works properly.

@f-f I got a bit excited and made it possible for the `modifyRawDhallExpression` code to also update the `name` and `sources` key. However, it would be tedious to add tests for those, too. So, I am going to remove that code but edit the `ConfigModifcation` and `AstUpdate` types in such a way that these can be easily added in the future.

JordanMartinez:
Found another bug with this expression:
```dhall
let x = { key1 = ./spago.dhall } in x.key1
```
The issue not accounted for here is that `x.key1` adds the `key1` to the `keyStack`. However, when we recurse back up the expression to `let x = ...`, that key is not included on the key stack. Thus, we'll get to `{ key1 = ... }` and expect a dependencies key to be there. When that's not there, the `updateExpr` code fails, Nothing is returned, and the file is not updated.

JordanMartinez:
Example of what the control flow looks like (this would normally be [debug] not [warn]):
```
[warn] Level: AtRootExpression "dependencies" - Let( variable = "x")
[warn] Level: SearchingForField ("dependencies" :| []) - Field( fieldSelectionLabel = "key1")
[warn] Level: SearchingForField ("key1" :| ["dependencies"]) - Var(varName ="x", deBrujin Index = 0)
[warn] Level: SearchingForField ("dependencies" :| []) - Field( fieldSelectionLabel = "key1") - got: Just "VariableName x 0 [key1, dependencies]"
[warn] Level: AtRootExpression "dependencies" - Let( variable = "x") - inExpr - got: Just "VariableName x 0 [key1, dependencies]"
[warn] Level: SearchingForField ("key1" :| ["dependencies"]) - RecordLit( keys = ["key1"]
[warn] Level: SearchingForField ("dependencies" :| []) - Embed
[warn] Level: SearchingForField ("key1" :| ["dependencies"]) - RecordLit( keys = ["key1"] - got: Just "EncounteredEmbed"
[warn] Level: AtRootExpression "dependencies" - Let( variable = "x") - value - got: Just "EncounteredEmbed"
```

test/fixtures/alternative4-root-let-value-embed-field-success.dhall
```diff
@@ -1 +1,7 @@
-let x = { key1 = let __embed = ./spago.dhall in __embed.dependencies # ["newtype"] } in x.key1
\ No newline at end of file
+let x =
+      let __embed = { key1 = ./spago.dhall }
+
+      in  __embed
+        with key1.dependencies = __embed.key1.dependencies # [ "newtype" ]
+
+in  x.key1
```
JordanMartinez, line 7: Stylistically, this is not the most "efficient" update we could make. Rather, the following would be a better way to handle this:
```dhall
let x = 
  { key1 = 
    let __embed = ./spago.dhall in 
    in  __embed with __embed.dependencies # ["newtype"]
  }

in  x.key1
```

JordanMartinez: This has been resolved via the `Embed` case when level is `SearchingForField`.

test/fixtures/alternative4-root-recordlit-let-success.dhall
```diff
@@ -0,0 +1,7 @@
+{ name = "my-project"
+, dependencies =
+      (let x = [ "console", "effect", "prelude", "psci-support" ] in x)
+    # [ "newtype" ]
```
JordanMartinez, line 4: Well... I thought this would be the last test I would need to add, but stylistically, this is just stupid. It should be
```dhall
let x = [ "console", "effect", "newtype", "prelude", "psci-support" ] in x
```
However, that means the `x` could be a record where some part of it is selected:
```dhall
{ ...
, dependencies = let x = { deps = [ ... ] } in x.deps
}
```
Looks like I need to handle that possibility, too, by making all record expressions work even at the `WithinField` level.

JordanMartinez: Basically, I would need to attempt to do the update within the let binding to account for the situation above. However, I would need to stop the traversal back out of the `WithinField` level if the expression ultimately refers to some other binding outside of that scope. For example:
```dhall
-- given
let lsBinding = ["old"] ... in { dependencies =  let x = lsBinding in x }

-- Wrong: side-effect introduced if `lsBinding` is used elsewhere besides the dependencies field
let lsBinding = ["old", "new"] ... in { dependencies = let x = lsBinding in x }

-- Correct: update occurs as close to local binding as possible
let lsBinding = ["old"] ... in { dependencies = let x = lsBinding # ["new"] in x  }
```

JordanMartinez: I think for simplicity sake, let's not support this right now.

Let's say I have multiple let bindings here:
```dhall
let lsBinding = ["foo"]
{ ...
, dependencies = let x = lsBinding let y = x let z = y in z
}
```
When I hit the first let binding (i.e. `x`), I'm at level `WithinField`. When I attempt to update its `inExpr` (i.e. the `y` let binding), I'm still at level `WithinField`. When I finally get to the `z` let binding's `inExpr`, I now learn that it refers to the `y` binding's value. So what do I do?
1. Since I'm still at `WithinField`, do I just wrap the `z` binding's `inExpr` (e.g. `let ... let z = y # ["new"]`)? 
2. Or do I traverse back up the expression (e.g. `let x = lsBinding # ["new"] ... in z`)?

While the second is the obvious approach, the problem is that `WithinField` does not currently track how many let bindings we have gone down before we learned that we need to traverse back up the let binding expressions. Since it doesn't track that, the current code will eventually update the `lsBinding` binding's value rather than append the "new" package to the value of the `x` binding. To fix this, we'd need to update `WithinField` to also track how many let bindings it has gone down, so that if we would otherwise go back "up" the expression but find that the current counter is `0`, we know we must do the update here (i.e. at the `x` binding) rather than a binding that `x` also references (i.e. the `lsBinding`).

Since I've already written a lot of tests for this and don't feel like adding another 15 or so to account for this case, I think it's simpler to say, "We don't support it for now," because of how rare this will be. If we did implement this, we'd also need to account for things like field selection. For example,
```dhall
let lsBinding = {deps = ["foo"] }
{ ...
, dependencies = let x = lsBinding.deps let y = x let z = y in z
}
```

JordanMartinez:
@f-f This is ready for a final review. Tests covering each case have been added (unless of course the case itself is just an invalid Dhall expression).

Here's a few things to consider when reviewing this again:
- Some updates aren't the best stylistically, but will still function. For example, see the let binding issue mentioned above (e.g. `dependencies = let x = lsBinding let y = x let z = y in z`). Another example happens with the `Embed` case. By dropping the `EncounteredEmbed` case and just updating that expression as locally as possible, we no longer do the update at the root. When we have a root-level `Prefer _ embedExpr recordLitExpr`, the update doesn't look as nice.
- Are there any docs that should be updated?
- Should any file names be updated?
- Should any functions be renamed?
- Should `modifyRawDhallExpression` itself be moved to a module called `Spago.Dhall.AST` and `modifyRawConfigExpression` moved to `Spago.Config`?

test/fixtures/alternative4-root-recordlit-let-success.dhall
JordanMartinez: Ok... I got curious to see how hard this would be, and was forced to do the "ExprLevel is just a stack of keys" optimization I mentioned in the original doc we covered this past Tuesday. In short, the latest commits make the amount of code I'm adding much smaller. Moreover, supporting targets is literally just adding a `"targets"` in front of the initial key stack we pass into the initial `updateExpr` call.

```diff
modifyRawDhallExpression :: HasLogFunc env => Text -> AstUpdate -> Expr -> RIO env Expr
modifyRawDhallExpression initialKey astMod originalExpr = do
-  result <- updateExpr (ExprLevel { levelKeyStack = ["dependencies" }) originalExpr
+  result <- updateExpr (ExprLevel { levelKeyStack = ["targets", "dependencies"] }) originalExpr
```

f-f: This is wonderful!

f-f:
@JordanMartinez I noticed you added some more code after your last comment of "this is now ready for review", should I have a look or do you think there are more changes on the way?

JordanMartinez:
@f-f I feel like "the guy who cried wolf" but replacing 'wolf' with 'review.' :sweat_smile: 

I'd like to update the docs, and look over the branches again to verify that each is being tested in current tests. Then it'll be ready for a review.

JordanMartinez:
47a056e adds the test that checks [this situation](https://github.com/purescript/spago/pull/815/#discussion_r705492677) and fixes the result. I tried to make the update to the expression [here](https://github.com/purescript/spago/pull/815/#discussion_r705457218) prettier, but I think it could be an invalid update. So, I've decided to use an "always correct" modification that's not as pretty (though why one would be using let bindings here would be beyond me).
```dhall
-- always correct, but not pretty
let x = [ "console", "effect", "prelude", "psci-support" ] in x # [ "newtype" ]

-- pretty, but I think it can be incorrect?
let x = [ "console", "effect", "newtype", "prelude", "psci-support" ] in x
```
So the comments on the case branches have been updated. The remaining work to do here is:
- [ ] Update the doc comments on functions to ensure they are still correct (now that I've implemented that optimization)
- [ ] Verify one last time that all branches have tests

JordanMartinez:
@f-f This is ready for a review. I think with my "optimized via levelKeyStack" implementation for `ExprLevel`, we might even be able to drop some of the tests I've written.

f-f:
Apologies for the delayed response on this one - I have tried to convince myself that this should be merged ASAP, but I failed: the reason why I'm hesitant is that this PR adds a really big chunk of code (1.8k, of which 1k of Haskell, in a codebase of 6k lines, tests included) for implementing a feature that will need to change (we don't know how dramatically, but the fact that we don't know is part of the problem) once we ship the new Config format to support the Registry.

I like this code and I'd like to have it in, but I would like to hold off until we ship the first version of the Registry, so we can consider this all together in the design of the new config.

JordanMartinez:
> I like this code and I'd like to have it in, but I would like to hold off until we ship the first version of the Registry, so we can consider this all together in the design of the new config.

That's fine by me! I think your decision is wise.

f-f:
Superseded by #849

PR #817, opened 2021-09-10
Always include psci-support when running `spago repl`

JordanMartinez:
### Description of the change

Drops the need to have `psci-support` added to one's dependencies by always adding it when one runs `spago repl`. See the last point in [this comment](https://github.com/purescript/spago/issues/681#issuecomment-909243133) for more details.

~Note: This is not meant to be merged until AFTER #815~

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+Features:
+- Drop need for `psci-support` to be defined in `dependencies` field (#817)
+
 Bugfixes:
 - Don't warn on unused deps when building --deps-only. (#794)
 
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -53,8 +53,8 @@ main = withUtf8 $ do
         -> CLI.echo spagoVersion
       Path whichPath buildOptions
         -> Path.showPaths buildOptions whichPath
-      Repl replPackageNames paths pursArgs depsOnly
-        -> Spago.Build.repl replPackageNames paths pursArgs depsOnly
+      Repl replPackageNames paths pursArgs depsOnly replPackage
+        -> Spago.Build.repl replPackageNames paths pursArgs depsOnly replPackage
       BundleApp modName tPath shouldBuild buildOptions
         -> Spago.Build.bundleApp WithMain modName tPath shouldBuild buildOptions globalUsePsa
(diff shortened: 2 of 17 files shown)


JordanMartinez:
I've fixed the merge conflict via GitHub. CI should pass, but I haven't confirmed locally.

JordanMartinez:
CI now passes.

f-f (commented):
This looks great! One thing that I think we should address before merging is the fact that, while `psci-support` is by far the most used REPL package, someone might want to use some other package. E.g. [here](https://discourse.purescript.org/t/collaborating-on-improving-the-repl-evaluator/1333/15) we discussed about having a `psci-support-aff` package to enable an asynchronous REPL.
I think the easiest way to handle this would be to add a flag (something like `--no-implicit-repl-package`) to disable this "magic" behaviour.

JordanMartinez:
> I think the easiest way to handle this would be to add a flag (something like --no-implicit-repl-package) to disable this "magic" behaviour.

Perhaps instead we should have a flag `--repl-package` that allows one to indicate which package to use for the repl? Otherwise, those who want to use `psci-support-aff` would still need to declare it in their dependencies. If the default is `psci-support`, then they could override it with `--repl-package psci-support-aff`.

Or as a different solution, we indicate in the `spago.dhall` file what the default package should be when starting the repl? Something like a `repl-package = psci-support`?

f-f:
@JordanMartinez I think we should have both! I started keeping track of the config part in #842, and we can go ahead with the flag over here :slightly_smiling_face:

JordanMartinez:
@f-f Sorry for taking so long to get to this, but this should be ready for review.

JordanMartinez:
```sh
$ stack exec bash <<EOF
spago repl --help
EOF

Usage: spago repl [-d|--dependency DEPENDENCY] [-p|--path PATH] 
                  [-u|--purs-args PURS-ARGS] [--deps-only] 
                  [--repl-package PACKAGE]
  Start a REPL

Available options:
  -d,--dependency DEPENDENCY
                           Package name to add as a dependency
  -p,--path PATH           Source path to include (in addition to paths in
                           spago.dhall)
  -u,--purs-args PURS-ARGS Arguments to pass to purs compile. Wrap in quotes.
  --deps-only              Only use sources from dependencies, skipping the
                           project sources.
  --repl-package PACKAGE   The REPL package to use for evaluating
                           expressions (default: "psci-support")
  -h,--help                Show this help text
```

f-f (approved):
This is great, thank you!

PR #818, opened 2021-09-10
Convert deps & source paths from List to Set

JordanMartinez:
### Description of the change

Converts the `dependencies` and `configSourcePaths` types for `Config` from a `List` to a `Set`. See [this comment and the comments preceding it](https://github.com/purescript/spago/issues/813#issuecomment-908369678) for more context. We shouldn't pass the same source globs into `purs compile` or `purs repl` multiple times. By storing these values as a `Set`, we can add new items without fear of duplication.

I wasn't sure whether to include this in #817

~Note: This is not meant to be merged until AFTER #815~

### Checklist:

### DIFF
--- a/src/Spago/Build.hs
+++ b/src/Spago/Build.hs
@@ -55,9 +55,9 @@ build maybePostBuild = do
   BuildOptions{..} <- view (the @BuildOptions)
   Config{..} <- view (the @Config)
   deps <- Packages.getProjectDeps
-  let partitionedGlobs@(Packages.Globs{..}) = Packages.getGlobs deps depsOnly configSourcePaths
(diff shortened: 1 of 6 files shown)


f-f:
So this can be merged on its own before #815?

JordanMartinez:
More or less. I would need to rework parts of #815, but I think that PR's commit history could be cleaned up anyway. So, I was thinking I would reimplement it in a new PR.

f-f:
Sounds good! No worries about commit history, we squash everything anyways.

CI passes and I'll now approve this, so you can merge it in if it's ready to go.

f-f (approved):

Issue #828, opened 2021-12-05
Failed to run a JS file when spago bundle-app with purescript-unordered-collections


kalxd:
I running this code on `purs 0.14.5` and `spago 0.20.3`:

```purescript
module Main where

import Prelude

import Data.HashMap (HashMap, empty, insert)
import Effect (Effect)
import Effect.Console (logShow)

type MyHash = HashMap String String

main :: Effect Unit
main = do
  let o = empty
  logShow $ insert "key" "value" o
```

it runs very well with command `spago run`. 
when I bundle them with `spago bundle-app` and run it:

```bash
➜  Sample node index.js
D:\msys64\home\XGLey\Code\Purescript\Sample\index.js:474
  Collision.prototype.lookup = function collisionLookup(Nothing, Just, keyEquals, key, keyHash, shift) {
  ^

ReferenceError: Collision is not defined
    at D:\msys64\home\XGLey\Code\Purescript\Sample\index.js:474:3
    at Object.<anonymous> (D:\msys64\home\XGLey\Code\Purescript\Sample\index.js:691:3)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
    at internal/main/run_main_module.js:17:47
```

It seems `Collision constructor` is not in the bundled sources.

klntsky:
It looks like spago calls `purs` with `-m` argument instead of `--main` [here](https://github.com/purescript/spago/blob/aa7c0de6d903262f69452663c09fad9c441af8d3/src/Spago/Purs.hs#L90).


Using direct call to purs works for me.

```
➜  purs bundle 'output/*/*.js' -m Main -o target.js && node target.js
/home/me/c/repor-1/target.js:474
  Collision.prototype.lookup = function collisionLookup(Nothing, Just, keyEquals, key, keyHash, shift) {
  ^

ReferenceError: Collision is not defined
    at PS.Data.HashMap (/home/me/c/repor-1/target.js:474:3)
    at Object.<anonymous> (/home/me/c/repor-1/target.js:691:3)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
[...]

PR #829, opened 2021-12-05
Fix argument name in purs bundle call (#828)

klntsky:
### Description of the change

Looks like the argument name was wrong (#828)

### Checklist:


klntsky:
Oops, this is probably the wrong way to fix it. I'll close my PR for now.

PR #830, opened 2021-12-07
Add sha256 checksum generation on the release workflow

### LINKED ISSUE #816: Spago should have a release checksum to avoid dependency confusion attack
One of most common way to hijack a software project is by placing compromised dependencies in place of real ones. Adding a checksum of each release asset in the build process would allow to check if downloaded file corresponds to the expected file.

[Purescript compiler](https://github.com/purescript/purescript) does this using a SHA checksum. So I can do this in my project:

```Makefile
# Makefile
vendor/purescript:
	curl -o /tmp/purescript_linux64.tar.gz -fSL https://github.com/purescript/purescript/releases/download/v${PURS_VERSION}/linux64.tar.gz
	shasum /tmp/purescript_linux64.tar.gz | grep ${PURS_HASH}
	tar -xzf /tmp/purescript_linux64.tar.gz -C vendor
	chmod +x vendor/purescript/purs
	rm /tmp/purescript_linux64.tar.gz
```

`shasum` validates the downloaded binary, so there is no chance to replace it by a compromised binary hijacking the name resolution.

klarkc:
### Description of the change

Add sha256 checksum generation on the release workflow

Resolves #816 

### Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -103,8 +103,8 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ needs.draft_release.outputs.upload_url }}
-          asset_path: ./${{ runner.os }}.sha
-          asset_name: ${{ runner.os }}.sha
+          asset_path: ./${{ runner.os }}.sha1
+          asset_name: ${{ runner.os }}.sha1
           asset_content_type: text/plain
 
   npm_publish:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -59,7 +59,12 @@ jobs:
         run: |
           brew install gnu-tar
           echo "/usr/local/opt/gnu-tar/libexec/gnubin" >> $GITHUB_PATH
-
+      # NB: We install coreutils because we need sha256sum on macOS
+      - name: Install coreutils (macOS)
+        if: runner.os == 'macOS'
+        run: |
+          brew install coreutils
+          echo "/usr/local/opt/coreutils/libexec/gnubin" >> $GITHUB_PATH
       - name: Package binary (Linux and macOS)
         if: runner.os != 'Windows'
         shell: bash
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -93,6 +98,23 @@ jobs:
           asset_name: ${{ runner.os }}.tar.gz
           asset_content_type: application/gzip
 
+      - name: Package Checksum
+        shell: bash
+        run: |
+          sha256sum ${{ runner.os }}.tar.gz | awk '{print $1}' > ${{ runner.os }}.sha
```
klarkc, line 104: This part can break in Windows (idk if these commands are available in the windows environment), I don't have a Windows machine to test that.

f-f: Yeah, I don't think we can use `sha256sum` on Windows. We should do [like the compiler does](https://github.com/purescript/purescript/blob/18e3b14c556c5bd58e391531e4f4e511f834c3ff/bundle/build.sh#L48-L57), and use `openssl` on Windows. Quoting from the link above:
```bash
# Calculate the SHA hash
if [ "$OS" = "win64" ]
then
  # msys/mingw does not include shasum. :(
  SHASUM="openssl dgst -sha1"
else
  SHASUM="shasum"
fi

$SHASUM bundle/${OS}.tar.gz > bundle/${OS}.sha
```

.github/workflows/release.yml
```diff
@@ -93,6 +92,21 @@ jobs:
...
+
+      - name: Upload Checksum Asset
+        id: upload-checksum-asset
+        uses: actions/upload-checksum-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ needs.draft_release.outputs.upload_url }}
+          asset_path: ./${{ runner.os }}.sha
+          asset_name: ${{ runner.os }}.sha
```
f-f, line 107: In the previous step we point the script to a file with a `sha1` extension, so I think we should use that here too:
```suggestion
          asset_path: ./${{ runner.os }}.sha1
          asset_name: ${{ runner.os }}.sha1
```

f-f (approved):
Thank you!

Issue #836, opened 2022-01-09
Removing the `bundle` commands


f-f:
[The community is evaluating a proposal to remove `purs bundle`](https://discourse.purescript.org/t/rfc-the-future-of-the-purescript-bundler).

This would affect our `bundle-module` and `bundle-app` commands, as we'd have to decide what to do in such case.
There are some proposals in that Discourse thread, but I'm opening this issue to track the decision here.

thomashoneyman:
Snowpack [takes an interesting approach](https://www.snowpack.dev/guides/optimize-and-bundle) in which they include `esbuild` support by default, but you can swap in other bundlers as you see fit.

sigma-andex:
I created a draft pr #839 to get the discussion started. I changed `spago run` to supoprt es modules and replaced `bundle` to use `esbuild`.

f-f:
In #862 we started using `esbuild` for bundling code when using purs 0.15, deciding to keep the `bundle` commands in spago, since they provide an easy-to-get-started bundling experience.
So I think this resolves the current ticket, and I'll close. Let's reopen it if we figure out that we want to go in another direction instead

PR #839, opened 2022-01-18
Add support of es modules branch

sigma-andex:
### Description of the change

Draft pr for #836 using esbuild as bundler to get the discussion started. 

### Checklist:

### DIFF
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -55,10 +55,10 @@ main = withUtf8 $ do
         -> Path.showPaths buildOptions whichPath
       Repl replPackageNames paths pursArgs depsOnly
         -> Spago.Build.repl replPackageNames paths pursArgs depsOnly
-      BundleApp modName tPath shouldBuild buildOptions
-        -> Spago.Build.bundleApp WithMain modName tPath shouldBuild buildOptions globalUsePsa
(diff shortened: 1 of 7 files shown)


f-f:
@sigma-andex what's the status of this? I see this is not a draft anymore, but would you like to actually have it merged right now?

I'm not sure we would merge any breaking change of this sort until changes to the bundle logic in the compiler are released, and even then we would still support the `bundle` commands as they are for earlier compilers.
My understanding is that this patch exists to make the es-module work move forward, but I'm not sure it should even be a pull request yet?

sigma-andex:
No, not yet merged of course until this is coordinated with the v0.15 release. 
But it would be great to get some feedback/review on this.

i-am-the-slime:
@f-f I think it would be nice if the maintainers (you? 😁 ) could give their input on whether this needs changes or it could be merged in concert with the 0.15 release. 
I think it makes sense to get some PRs ready and publicise this early on rather than it just happening inside the obscure es-modules-working-group org.
 
If you prefer for it to stay a draft that's fine too, of course.

sigma-andex:
@f-f regarding the backwards compatibility: I can have a look at this. I think that shouldn't be a problem if the version information is available somewhere

sigma-andex:
Closing this in favour of #862

PR #841, opened 2022-01-25
Add --offline flag

### LINKED ISSUE #575: Add `--offline` flag
Currently there's no way to ensure that `spago` does not connect to the network. Even if all dependencies are installed, `--no-install` is set, there's still traffic.

This makes `spago` unusable in pure CI environments, such as `nix`. And it would be nice if that was not the case.

JordanMartinez:
### Description of the change

Fixes #575

### Checklist:

### DIFF
--- a/src/Spago/FetchPackage.hs
+++ b/src/Spago/FetchPackage.hs
@@ -41,10 +41,6 @@ fetchPackages
 fetchPackages allDeps = do
   logDebug "Running `fetchPackages`"
 
-  internetAccess <- view (the @GlobalOffline)
-  when (internetAccess == Offline) $ do
-    die [ display Messages.fetchPackagesOffline ]
-
   PackageSet.checkPursIsUpToDate
 
   -- Ensure both local and global cache dirs are there
@@ -62,6 +58,9 @@ fetchPackages allDeps = do
   -- Note: it might be empty depending on the cacheFlag
   let nOfDeps = List.length depsToFetch
   when (nOfDeps > 0) $ do
+    internetAccess <- view (the @GlobalOffline)
+    when (internetAccess == Offline) $ do
+      die [ display Messages.fetchPackagesOffline ]
     logInfo $ "Installing " <> display nOfDeps <> " dependencies."
     metadata <- GlobalCache.getMetadata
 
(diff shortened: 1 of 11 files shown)


JordanMartinez:
I'm not really sure what should be added to the readme (if anything). It's more likely that Nix tools would update their readme to show how to use this command in their workflow.

src/Spago/FetchPackage.hs
```diff
@@ -35,12 +35,16 @@ import qualified Spago.PackageSet              as PackageSet
...
 --       * if not, run a series of git commands to get the code, and copy to local cache
 fetchPackages
-  :: (HasLogFunc env, HasJobs env, HasGlobalCache env, HasPackageSet env)
+  :: (HasLogFunc env, HasJobs env, HasGlobalCache env, HasPackageSet env, HasGlobalOffline env)
   => [(PackageName, Package)]
   -> RIO env ()
 fetchPackages allDeps = do
   logDebug "Running `fetchPackages`"
 
+  internetAccess <- view (the @GlobalOffline)
```
f-f, line 44: This would prevent anyone from getting in this code path at all - however there are some cases where we don't need to reach for the internet at all, e.g. when all packages are in cache. About 10 lines from here we bind `depsToFetch` which contains the list of things that we actually have to download (i.e. we filter out all the things in cache). We can fail then if `nOfDeps > 0` and the flag is set

f-f:
I think we could merge this already, but I'd first like that we figure out a good testing story for this, because if we can't properly test that the changes we introduce are going to continue working, then we are deemed to break them.

I don't have many ideas about this though - `nix-shell` doesn't seem to offer any switch to disable the network :thinking:

JordanMartinez:
So.... how should I make progress on this?

thomashoneyman:
Could you mock all network calls so they fail the test right away if called? You could use these mocked calls for the offline mode test only.

f-f:
Oh, it looks like we can call a process with `unshare -r -n` to prevent it from reaching the network.

So I'd say the next step on this is to fixup a test infrastructure that runs the tests for the offline feature in such container

JordanMartinez:
Hmm... I should add a second test that verifies `spago build` fails when we drop the `offline` flag but still use `unshare -r -n`...

JordanMartinez:
Dropping the `offline` flag but still running `unshare -r -n` does not produce the expected failure. I wondered as much. AFAICT, `spago build` should always work so long as the dependencies are stored in the cache folder. I think the real issue is that running Spago in a Nix environment means the `.cache` folder doesn't contain these dependencies and so triggers the network call.

f-f:
@JordanMartinez I see. The location of both Dhall and Spago folders is dependent on the value of the `XDG_CACHE_DIR` environment variable. We could try to set that do a temp directory when running these tests, so that they would start from a clean slate?

PR #843, opened 2022-01-28
Override a package in the package set with a remote one: dependencies

flip111:


f-f:
I'm a little concerned about giving out the idea that all three keys are needed in order to override something, do you think we could move this to some other place?

flip111:
the idea comes from spago itself, seems right to document it

```
[error] Failed to read the config. Error was:
[error] Error: Error while reading spago.dhall:

Explanation: a record is missing a required key.

The key missing is:

↳ dependencies

The keys in the record are:

↳ repo, version
```

f-f:
Oh, but that's the case when you add a new package (which is documented in the next section), this section deals with overriding an existing package

flip111:
Thought package already existed. Don't really understand why i had to copy paste dependencies ?

https://github.com/natefaubion/purescript-tidy-codegen/blob/main/spago.dhall#L14-L56

f-f:
It's for historical reasons - we couldn't/didn't want to read them from Bower files when creating the package set. This is going to change once the new [Registry](https://github.com/purescript/registry) is live

flip111:
ok let's close this for now, registry sounds promising !

Issue #845, opened 2022-01-30
Remove `run.js`


sigma-andex:
Remove creation of `run.js` file for running and testing in the backend. 

This is necessary for the `v0.15.0` release and resolves some bugs.

sigma-andex:
Solved by #846

PR #846, opened 2022-01-30
Remove `run.js`

sigma-andex:
### Description of the change

This removes the creation of a `run.js` file when running `spago run`. The file content is now passed as an argument to node. Solves #845 .

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 Other improvements:
-- Remove run.js in preparation of `v0.15.0` release (#845)
+- Avoid writing a JS file when executing `spago run` (#845, #846)
 
 ## [0.20.4] - 2022-01-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+Other improvements:
+- Remove run.js in preparation of `v0.15.0` release (#845)
+
 ## [0.20.4] - 2022-01-29
 
 Bugfixes:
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+Other improvements:
+- Remove run.js in preparation of `v0.15.0` release (#845)
```
f-f, line 11: The current line might be cryptic to a future reader that doesn't know about `run.js` or `0.15.0`, so I propose we explain this with more context:
```suggestion
- Avoid writing a JS file when executing `spago run` (#845, #846)
```

src/Spago/Build.hs
```diff
@@ -330,29 +330,24 @@ runBackend maybeBackend RunDirectories{ sourceDir, executeDir } moduleName maybe
...
     nodeAction outputPath' = do
-      logDebug $ "Writing " <> displayShow @Text runJsSource
-      writeTextFile runJsSource (nodeContents outputPath')
-      void $ chmod executable $ pathFromText runJsSource
       -- cd to executeDir in case it isn't the same as sourceDir
       logDebug $ "Executing from: " <> displayShow @FilePath executeDir
       Turtle.cd executeDir
       -- We build a process by hand here because we need to forward the stdin to the backend process
-      let processWithStdin = (Process.shell (Text.unpack nodeCmd)) { Process.std_in = Process.Inherit }
+      let processWithStdin = (Process.shell (Text.unpack $ nodeCmd outputPath')) { Process.std_in = Process.Inherit }
```
f-f, line 350: Now that we are not writing a file anymore we should log the command instead:
```suggestion
      logDebug $ "Running node command: `" <> nodeCmd outputPath' <> "`"
      let processWithStdin = (Process.shell (Text.unpack $ nodeCmd outputPath')) { Process.std_in = Process.Inherit }
```

f-f (approved):
Thank you!

sigma-andex:
Awesome. thx :-)

Issue #861, opened 2022-02-18
Add support for bundling es modules


sigma-andex:
As discussed in [this proposal in `purescript/purescript`](https://github.com/purescript/purescript/issues/4226) and in the working group [tooling evaluation](https://github.com/working-group-purescript-es/challenges/issues/5) `purs bundle` will be dropped in favour of `esbuild`.

PR #862, opened 2022-02-18
Add support for es modules bundling using esbuild

sigma-andex:
### Description of the change

Solves #861 

- v0.14 bundling uses cjs and `purs bundle`
- v0.15 bundling uses es modules and `esbuild`

### Checklist:

### DIFF
--- a/app/Spago.hs
+++ b/app/Spago.hs
@@ -55,10 +55,8 @@ main = withUtf8 $ do
         -> Path.showPaths buildOptions whichPath
       Repl replPackageNames paths pursArgs depsOnly replPackage
         -> Spago.Build.repl replPackageNames paths pursArgs depsOnly replPackage
-      BundleApp modName tPath platform minify shouldBuild buildOptions
-        -> Run.withPursEnv globalUsePsa $ Spago.Build.bundleApp WithMain modName tPath platform minify shouldBuild buildOptions globalUsePsa
-      BundleModule modName tPath platform minify shouldBuild buildOptions
-        -> Run.withPursEnv globalUsePsa $ Spago.Build.bundleModule modName tPath platform minify shouldBuild buildOptions globalUsePsa
+      BundleModule withMain bundleOptions buildOptions
+        -> Run.withPursEnv globalUsePsa $ Spago.Build.bundleModule withMain bundleOptions buildOptions globalUsePsa
       Script modulePath tag dependencies scriptBuildOptions
         -> Spago.Build.script modulePath tag dependencies scriptBuildOptions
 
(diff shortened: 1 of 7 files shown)


app/Spago.hs
```diff
@@ -55,10 +55,10 @@ main = withUtf8 $ do
...
       Repl replPackageNames paths pursArgs depsOnly replPackage
         -> Spago.Build.repl replPackageNames paths pursArgs depsOnly replPackage
-      BundleApp modName tPath shouldBuild buildOptions
-        -> Spago.Build.bundleApp WithMain modName tPath shouldBuild buildOptions globalUsePsa
-      BundleModule modName tPath shouldBuild buildOptions
-        -> Spago.Build.bundleModule modName tPath shouldBuild buildOptions globalUsePsa
+      BundleApp modName tPath platform minify shouldBuild buildOptions
+        -> Run.withPursEnv globalUsePsa $ Spago.Build.bundleApp WithMain modName tPath platform minify shouldBuild buildOptions globalUsePsa
+      BundleModule modName tPath platform minify shouldBuild buildOptions
+        -> Run.withPursEnv globalUsePsa $ Spago.Build.bundleModule modName tPath platform minify shouldBuild buildOptions globalUsePsa
```
f-f, line 61: At this point `bundleModule` is taking 7 arguments, which is a little much. We should make a new datatype `BundleOptions` and pass that one in

src/Spago/Purs.hs
```diff
@@ -74,24 +74,50 @@ repl sourcePaths extraArgs = do
...
+    minifyOpt = case minify of 
+      NoMinify -> ""
+      Minify -> " --minify"
+    cmd = case withMain of
+      WithMain -> 
+        "echo \"import { main } from './output/" <> moduleName <> "/index.js'\nmain()\" | "
+        <> esbuild <> " --platform=" <> platformOpt <> minifyOpt <> " --bundle "
+        <> " --outfile=" <> targetPath
+      WithoutMain -> 
+        esbuild <> " --platform=" <> platformOpt <> minifyOpt <> " --bundle "
```
f-f, line 100: I'm not sure all this logic about `esbuild` should go here, since this module is supposed to only be dealing with `purs` itself. Would it be possible to pull this logic up in the `Build` module?

sigma-andex: yeah I was also wondering where best to put that, and then decided to have it next to the purs bundle so to not have it in two different places, but I can move it to `Build.hs`

app/Spago.hs
sigma-andex: Is WithoutMain synonymous with module bundling in v0.14? Or is it possible to bundle a module with main?

f-f: It's not possible, see here:
https://github.com/purescript/spago/blob/3d1a5be3efe37e92e822d74056461c2f1f33099e/src/Spago/Build.hs#L403-L409

sigma-andex: Ok I simplified this and merged `bundleApp` and `bundleModule` functions. Moved the es bundling to `Build.hs`

sigma-andex: Also added the `BundleOptions` data type

sigma-andex:
@f-f any news on this? :-)

f-f (approved):
Looks great, thanks!

Issue #822, opened 2021-10-08
Bug with spaces in filenames


nathanielvirgo:
I just did the following:

* installed purescript and spago via npm
* created a folder `/Users/nathaniel/Code/purescript learning/test-project-1` and cd'd into it
* ran `spago init`
* ran `spago build`
* ran `spago test`

I was surprised to get the following error:

```
[info] Build succeeded.
internal/modules/cjs/loader.js:905
  throw err;
  ^

Error: Cannot find module '/Users/nathaniel/Code/purescript'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:76:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
[error] Tests failed: exit code: 1
```

but I guessed correctly that the problem was the space in the name of the enclosing folder. Changing its name from `purescript learning` to `purescript-learning` gives the expected result, i.e. it reports success.

This is only a minor inconvenience, but it seems a bit silly to fail due to spaces in a filename.

f-f:
Thanks for reporting this! It's definitely a bug, and it comes from the way that we are calling `node` (i.e. I suspect that something is not quoted [when we put together the script to call](https://github.com/purescript/spago/blob/7728ce42a1c66968cce7a72dbb6ab05ea0904d9f/src/Spago/Build.hs#L335)) in `spago run`/`spago test`

jbrains:
A note to the folks who need it. If you're bitten by this, but don't want to rename your project directory, you can run these commands:

```
# Equivalent to `spago run`
node -e "require('./output/Main/index').main()"

# Equivalent to `spago test`, as long as you use the default test module name.
# If you're a spago/purescript beginner, then you are probably using the default test module name. :)
node -e "require('./output/Test.Main/index').main()"
```

f-f:
Fixed in #846

PR #866, opened 2022-03-11
Undo removal of run.js

sigma-andex:
### Description of the change

Undo #846 since this does not work in the case of cli commands, see [this comment](https://github.com/purescript-contrib/pulp/pull/401#issuecomment-1059610898). 
Use `run.mjs` in the case of es modules since otherwise this will fail if `"type":"module"` is not set in the `package.json`.

### Checklist:


src/Spago/Build.hs
```diff
@@ -323,11 +323,14 @@ runBackend
 runBackend maybeBackend RunDirectories{ sourceDir, executeDir } moduleName maybeSuccessMessage failureMessage extraArgs = do
   logDebug $ display $ "Running with backend: " <> fromMaybe "nodejs" maybeBackend
   BuildOptions{ pursArgs } <- view (the @BuildOptions)
-  isES <- Purs.hasMinPursVersion "0.15.0"
+  isES <- Purs.hasMinPursVersion "0.15.0-alpha-01"
```
JordanMartinez, line 326: Can `Purs.hasMinPursVersion` be updated to drop the pre-release and other build metadata on the purs version before checking if it matches the minimum version? In other words, could this could be updated to read:
```
isEs <- Purs.hasMinPursVersion "0.15.0"
```

src/Spago/Build.hs
```diff
@@ -323,11 +323,14 @@ runBackend
...
   let postBuild = maybe (nodeAction isES $ Path.getOutputPath pursArgs) backendAction maybeBackend
   build (Just postBuild)
   where
     fromFilePath = Text.pack . Turtle.encodeString
+    runJsSource isES =
+      let file = if isES then "run.mjs" else "run.js"
```
JordanMartinez, line 332: Could this be updated to use `.js` for both CJS and ES modules as we did in `pulp`? It requires adding a `package.json` file to the same dir that stores the `run.js` file.

On Node.js <`13.0.0`, we'd also need to add the `--experimental-modules` flag before calling the file in the script. For example, `node --experimental-modules run.js`.

JordanMartinez (changes requested):
Just a few comments about this overall.

JordanMartinez:
Also, by reverting back to using files for `run`/`test`, we need to reopen #822. I think that issue should be tackled in a separate PR because it also affects CJS modules, too.

src/Spago/Build.hs
sigma-andex: I “downgraded” it to work with the current release. "0.15.0" > "0.15.0-alpha-01”.

src/Spago/Build.hs
sigma-andex: Is there any advantage in adding a `package.json` the `.spago` folder?

Good catch, I’m not sure though we have the node version here 🤔

sigma-andex:
> Also, by reverting back to using files for `run`/`test`, we need to reopen #822. I think that issue should be tackled in a separate PR because it also affects CJS modules, too.

yep makes sense. I wonder if there is a reason to use an absolute path, this problem shouldn’t happen with relative paths I guess

src/Spago/Build.hs
sigma-andex: Oh sorry I misread your comment. I guess so, but is this always wanted? I mean `hasMinPursVersion` is also used in other places.

JordanMartinez:
> yep makes sense. I wonder if there is a reason to use an absolute path, this problem shouldn’t happen with relative paths I guess

I'm not sure absolute paths are needed, but it seemed like it might have been needed for `pulp test` to work.

src/Spago/Build.hs
JordanMartinez: When we update the `contrib` libraries, they need to use `spago`. There may be additional alpha releases between now and then, so it might be `alpha-04` by that point. I'd prefer that we don't need to make a new `spago` release bumping the alpha number to something else each time a new PS alpha release is made.

sigma-andex: I think this shouldn't be a problem, because "0.15.0-alpha-04” should be greater than "0.15.0-alpha-01”, but I will check that this works

src/Spago/Build.hs
JordanMartinez: > Is there any advantage in adding a package.json to the .spago folder over using the extensions?

:man_shrugging: globs will always use a single extension?

src/Spago/Build.hs
sigma-andex:
```
(>=) <$> Version.semver "0.15.0-alpha-04" <*> Version.semver "0.15.0-alpha-01"
Right True
```

src/Spago/Build.hs
sigma-andex: Fixed

src/Spago/Build.hs
f-f: Is it also the case that `0.15.0 > 0.15.0-alpha-01`?

f-f:
> Also, by reverting back to using files for run/test, we need to reopen https://github.com/purescript/spago/issues/822. I think that issue should be tackled in a separate PR because it also affects CJS modules, too.

I wouldn't like to fix a bug by introducing a regression, so I think that bug should also be fixed in this PR.
To reproduce it we just need to have a project's folder name contain a space, so it's a test that we can easily add (and the fix should be very easy too I suspect)

sigma-andex:
> I wouldn't like to fix a bug by introducing a regression, so I think that bug should also be fixed in this PR.
> To reproduce it we just need to have a project's folder name contain a space, so it's a test that we can easily add (and the fix should be very easy too I suspect)

Regarding tests: I think it would be also nice to have some tests for this, since it has grown now in logic. But the problem is that we would need different purescript versions and I am not sure how we would do that. 
The alternative would be to refactor it and provide unit tests instead.. Do you have a proposal on how to provide some test for this whole bundling logic?

src/Spago/Build.hs
sigma-andex: Yes

sigma-andex:
I have added the escaping of the file name with spaces, which should fix the bug. I added also two more fixes for source maps and esm format on node when bundling modules.

As it is right now, it is ready for review, but I still don't know how to write some test cases without refactoring the whole thing. Also I don't really know how to add a test for the `space in file path` bug.

src/Spago/Build.hs
JordanMartinez: I still think the overall approach here is still problematic. This line should read:
```haskell
isES <- Purs.hasMinPursVersion "0.15.0"
```
or better yet
```hsakell
isES <- Purs.hasMinPursVersion purs0_15_0
where
-- purs0_15_0 would be defined as a constant somewhere
purs0_15_0 = SemVer 0 15 0 [] Nothing
```

The problem with the above line is that we don't indicate whether the possible pre-release and metadata stuff should be dropped in the `hasMinPurs` comparsion check. So, perhaps that function could take a second argument that indicates whether such "extra" information should be ignored in the comparison? If so, then it might read:
```
data SemVerComparison
   = IncludePreAndMeta
   | IgnorePreAndMeta

isES <- Purs.hasMinPursVersion purs0_15_0 IgnorePreAndMeta
```
and updating `hasMinPursVersion` as needed.

JordanMartinez (changes requested):
Just some general comments.

Also, I think a test can be added for #822 by changing [this line](https://github.com/purescript/spago/blob/master/test/SpagoSpec.hs#L21) to replace `spago-test` with `spago test`. However, when I did that locally, a few other tests failed, too.

src/Spago/Build.hs
sigma-andex: Regarding your first proposal:
I can do this, but then the `-alpha-0x` versions won't work.
Regarding your second proposal: I am not really sure how to do this.

thomashoneyman:
> Hi @f-f , do you have time to review this soon? It is currently blocking the `v0.15.0` ecosystem update since the `purescript-contrib` libraries depend on `spago test`. Thank you! :-)

We can also comment out spago test temporarily in CI and only rely on spago build, switching to pulp test for the time being. After all, we need to add bower and pulp to all the contrib libraries too. That way @f-f can have some breathing room to review as he’s got a ton of things on his plate.

src/Spago/Build.hs
JordanMartinez: This is what I mean: https://github.com/working-group-purescript-es/spago/pull/2

JordanMartinez:
> We can also comment out spago test temporarily in CI and only rely on spago build, switching to pulp test for the time being. After all, we need to add bower and pulp to all the contrib libraries too. That way @f-f can have some breathing room to review as he’s got a ton of things on his plate.

Works for me!

src/Spago/Build.hs
f-f: @JordanMartinez I am not sure I understand why including the prerelease segment is problematic there? This feature is indeed supported starting from the `alpha` releases, so this looks correct to me?

src/Spago/Cmd.hs
```diff
@@ -0,0 +1,26 @@
...
+import qualified Spago.Messages as Messages
+import qualified Turtle.Bytes
+import           Spago.Prelude
+import qualified Data.Text      as Text
+import qualified Data.Text.Encoding as Text.Encoding
+import qualified Data.Text.Encoding.Error as Text.Encoding
+import qualified Data.Versions  as Version
+
+-- | Get the semantic version of a command, e.g. purs --version
+getCmdVersion :: Text -> RIO env (Either Text Version.SemVer)
```
f-f, line 12: Having a whole separate module only for this is a bit noisy - we usually hold all these utilities in our custom `Spago.Prelude`, could we move this function there?

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@ Features:
 - Make `spago bundle-app` and `spago bundle-module` use esbuild and es modules for projects >= v0.15
 - Make `spago run` use es modules for projects >= v0.15 
 
+Bugfixes
+- Undo removing `run.js` as this does not work for cli programs.
+- Make spago escape filepaths so that spaces in paths are allowed
+- Fix adding experimental flag for node version 12
+- Remove support for node versions < 12 because they do not work with es modules
```
f-f, line 18:
```suggestion
- Fix `spago run` and `spago run` to accept command line arguments correctly, by writing a JS file to run (#865, #866)
- Remove support for node versions older than 12.0.0 as they do not work with es modules (#866)
```

f-f:
@sigma-andex a few more notes:
- as Jordan noted, to reproduce the "space in path" bug we need to rename the test directory to have a space in it. Let's do that and see what else breaks
- to test different purescript versions we could rework the build to be nix-based, as we're about to do in the Registry. AFAIK @thomashoneyman might have some input on this, as he's been looking at https://github.com/purescript/registry/issues/360.
  Alternatively, and in a less disruptive way, we could install purescript 0.14, run tests, then install purescript 0.15 and run tests again. We have some setup to deal with a similar things for [the `psa`-related tests](https://github.com/purescript/spago/blob/cc376c85bc6107ec38da2d0917170e046ba5a211/test/SpagoSpec.hs#L685), as we run them both with `psa` and without.

src/Spago/Cmd.hs
sigma-andex: I did this before, but then I ran into a cyclic dependency issue, that's why I created a separate module

src/Spago/Build.hs
JordanMartinez: @f-f @sigma-andex With how things are currently implemented, would all of the below expressions evaluate to true?
```
foo = do
  let
    s = Version.semver
    gt left right = (>) <$> left <*> right
  s "0.15.0-alpha-01" `gt `s "0.14.7"
  s "0.15.0" `gt` s "0.15.0-alpha-01"
  s "0.15.0" `gt` s "0.15.0-alpha-02"
  s "0.15.0" `gt` s "0.15.0-rc-01"
  s "0.15.0" `gt` s "0.15.0-rc-02"
```
If so, then I will drop this point.

sigma-andex:
```
>  s "0.15.0-alpha-01" `gt `s "0.14.7"
Right True
>  s "0.15.0" `gt` s "0.15.0-alpha-01"
Right True
>  s "0.15.0" `gt` s "0.15.0-alpha-02"
Right True
>  s "0.15.0" `gt` s "0.15.0-rc-01"
Right True
>  s "0.15.0" `gt` s "0.15.0-rc-02"
Right True
```

JordanMartinez: Welp! Ignore me! :sweat_smile:

JordanMartinez:
> Alternatively, and in a less disruptive way, we could install purescript 0.14, run tests, then install purescript 0.15 and run tests again. We have some setup to deal with a similar things for [the psa-related tests](https://github.com/purescript/spago/blob/cc376c85bc6107ec38da2d0917170e046ba5a211/test/SpagoSpec.hs#L685), as we run them both with psa and without.

This sounds like the easiest way forward.

sigma-andex:
I have added a check for v0.15 in the bundle tests. However at this point it is not really possible for me to test this, since there is not yet the 0.15 package set, or at least I don't know how to tell `spago init` use it. 
I added the space to the test folder.

test/SpagoSpec.hs
```diff
@@ -766,23 +783,36 @@ spec = around_ setup $ do
...
       spago ["init"] >>= shouldBeSuccess
       spago ["build"] >>= shouldBeSuccess
       -- Now we don't remove the output folder, but we pass the `--no-build`
       -- flag to skip rebuilding (i.e. we are counting on the previous command
       -- to have built stuff for us)
-      spago ["bundle-module", "--to", "bundle-module.js", "--no-build"] >>= shouldBeSuccess
-      checkFixture "bundle-module.js"
+      if pursVersion >= purs0_15_0 then do
+        spago ["bundle-module", "--to", "bundle-module.js", "--no-build"] >>= shouldBeSuccess
+        checkFixture "bundle-module-esm.js"
```
JordanMartinez, line 796: Whoops! Might want to bundle to `bundle-module-esm.js`, not `bundle-module.js`

sigma-andex: Good catch!

sigma-andex:
I undid adding a space to the test folder name. That broke quite a few things, which I think were more related to the test setup than to spago. I tried to add a small unit test that runs again within a temp folder with spaces in it, but I couldn't get that to work..

JordanMartinez:
When I ran the current PR's tests with a space in the `spago test` dir, I see that a file called `spago` in the `test` dir is created with the following content:
```
before test-c0eacd51de0f333c/testOutput
then test-51939f6c09d7e567/testOutput
before test-ce18bcb72081cc24/testOutput
then test-ce18bcb72081cc24/testOutput
else test-118dea0b8939b3b6/testOutput
else test-9903ece9494a9f12/testOutput
before1 test-9264a7f3865c0601/testOutput
before2 test-9264a7f3865c0601/testOutput
then1 test-9264a7f3865c0601/testOutput
then2 test-9264a7f3865c0601/testOutput
```

JordanMartinez:
@sigma-andex I'd re-add that space. Seems like you just found another bug.

JordanMartinez:
On line 487, this change makes the test pas
```diff
-       spago ["build", "--before", "echo before>> " <> ( Text.pack $ encodeString dumpFile )] >>= shouldBeSuccess
+       spago ["build", "--before", "echo before>> " <> ( Text.pack $ escapeFilePath $ encodeString dumpFile )] >>= shouldBeSuccess
```
where `escapeFilePath` is defined like so and added to the `Utils` module
```hs
import Data.Foldable (foldl')
escapeFilePath :: String -> String
escapeFilePath = reverse . foldl' escSpace ""
  where
    escSpace acc x
      | x == ' ' = x : '\\' : acc
      | otherwise = x : acc
```

sigma-andex:
@JordanMartinez I'm not sure how you did it but to me it seems that this change would need to be applied to various locations. We could do that, but I find this quite odd. I think there should rather be a separate test for this.

Regarding the 0.15 testing: Adding 0.15 to the github action will not be enough, because the tests update to the latest package set and there is not yet a 0.15 package set released.

JordanMartinez:
> I'm not sure how you did it but to me it seems that this change would need to be applied to various locations. We could do that, but I find this quite odd. I think there should rather be a separate test for this.

@sigma-andex Yeah, the change does need to be applied in various locations. To verify that the `before`, `then`, and `else` commands work properly, spago is running `spago build --before 'echo before >> spago test/testOutput'`. Because of the space, it's writing `before` to the `spago file`, not the `spago\ test/testOutput` file. So, on non-Windows OSes, we need to escape that space. Otherwise, when it attempts to read the `spago\ test/testOutput` file, the test finds that no such file exists (because we didn't write to the correct file).

We don't want to write separate tests for this as it incurs duplicating everything. If one set goes out of sync with another, it's harder to catch before it's already a bug.

> Regarding the 0.15 testing: Adding 0.15 to the github action will not be enough, because the tests update to the latest package set and there is not yet a 0.15 package set released.

Good point. I think we can workaround that if we modify how we call `spago init`. If a `packages.dhall` file already exists in the directory, I believe `spago init` will create all other files but ignore that one. I'll test this out locally and then push a change.

sigma-andex:
> We don't want to write separate tests for this as it incurs duplicating everything. If one set goes out of sync with another, it's harder to catch before it's already a bug.

I meant more like one separate test for testing the space issue so we don't have to change all the tests. But I'm fine with changing it as you did.

f-f:
> I meant more like one separate test for testing the space issue so we don't have to change all the tests

I'd agree with this. We can make a directory with a space inside the test ([see here](https://github.com/purescript/spago/blob/cc376c85bc6107ec38da2d0917170e046ba5a211/test/SpagoSpec.hs#L235)) and run a test in there so we don't have to change all the tests.

JordanMartinez:
Since spago depends so heavily on a package set that already exists in it's tests, it's hard to update this PR without modifying how `spago init` works. While the space issue can be resolved with more work, this PR might be blocked until we can release an initial `0.15.0` package set.

JordanMartinez:
> @JordanMartinez how come? It's possible to override the package set produced by `spago init`, e.g. we do it in tests [already here](https://github.com/purescript/spago/blob/cc376c85bc6107ec38da2d0917170e046ba5a211/test/SpagoSpec.hs#L637). If you have a branch somewhere you can point at the `src/packages.dhall` in there.

Because `spago init` exits with `1`:
```sh
$ purs --version
0.15.0-alpha-02
$ spago init
[error] It seems that you're using a compiler version that is not supported by package-sets at the moment.
Please install one of the following versions of the compiler and try again: 
- 0.14.7
- 0.14.6
- 0.14.5
- 0.14.4
- 0.14.3
- 0.14.2
- 0.14.1
- 0.14.0
- 0.13.8
- 0.13.6
- 0.13.5
- 0.13.4
- 0.13.3
- 0.13.2
- 0.13.0
- 0.12.5
- 0.12.4
- 0.12.3
- 0.12.2
- 0.12.1
- 0.12.0

$ echo $?
1
```

Which means every test that calls `spago ["init"] >>= shouldBeSuccess` needs to be modified somehow until we get a `0.15.0` package set. Right?

JordanMartinez:
I just pushed a few commits that only runs the `run` and `test` tests in a temporary directory called `spago test`. All others run in `spago-test`.

Since functions like `checkFixture` assume the fixtures directory is obtainable via the path `../fixtures`, this minimizes the amount of changes made to tests themselves. Moreover, tests added in the future can continue to use the same utility functions without accounting for more nested folders.

f-f:
@JordanMartinez if you create a `packages.dhall` before running `spago init` then it should accept what's there and not complain too much

JordanMartinez:
> @JordanMartinez if you create a `packages.dhall` before running `spago init` then it should accept what's there and not error out (though it looks like we don't do it in tests yet)

@f-f Correct, but where do we want to do this? Do we do this in every test? If so, updating the tests to handle this case every time `spago ["init"]` produces a large diff. Hence why I think just getting a working package set is better. If we're only doing this in the bundle, run, and test tests, then it's less of an issue. However, I don't believe every test that currently runs will pass on `v0.15.0`. So, we'll need a way to ignore such tests when testing against `v0.15.0`.

f-f:
@JordanMartinez right, I see. We can do this only for tests where we want to explicitly test 0.15 (I expect this kind of logic to stay even after 0.15 ships, but flipped to instead explicitly test some things on 0.14 while 0.15 becomes the default), by selectively running some of the tests after we install 0.15.
You can pass a pattern of tests to include to the test framework, in this way (e.g. this selects all the build tests that deal with alternate backends):
```
stack test --ta '--match "/Spago/spago build/alternate backend/"'
```

We can restructure the test tree to isolate the 0.15 tests and make them easy to run by themselves.

JordanMartinez:
Looks like adding a `packages.dhall` file doesn't quite work if it doesn't have a recognizable tag:
```
$ mkdir tryagain
$ cd tryagain/
$ cp ../test/fixtures/packages-prepare-0-15.dhall packages.dhall
$ dhall freeze packages.dhall
$ cat packages.dhall | tail -n 5
let upstream =
      https://raw.githubusercontent.com/purescript/package-sets/prepare-0.15/src/packages.dhall
        sha256:e94fe7acb228cfcfd0acae6d1e931ad8ddf264e372ac14c7f0cf66733862350c

in  upstream
$ spago init
[warn] Found a "packages.dhall" file, skipping copy. Run `spago init --force` if you wish to overwrite it.
[error] Cannot find a package set import in your "packages.dhall".
```

f-f:
@JordanMartinez that looks like a bug! We should not try so hard to crash if the compiler version doesn't look right :slightly_frowning_face:

JordanMartinez:
Bug or not, that can get fixed later as I provided a workaround for now. However, any idea what's going on with CI?

I'm actually not sure how to run the `v0.15.0` tests on Windows since `pulp` does that in PowerShell, not bash.

JordanMartinez:
Also, correct me if I'm wrong, but wasn't #868 supposed to prevent errors like the following?
```
purs: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.27' not found (required by purs)
```

f-f:
Now that https://github.com/purescript/purescript/pull/4228 was merged we can cut a new alpha release of `purs` and hopefully this should fix the glibc failure

JordanMartinez:
> Now that [purescript/purescript#4228](https://github.com/purescript/purescript/pull/4228) was merged we can cut a new alpha release of `purs` and hopefully this should fix the glibc failure

I'll work on that. However, the Windows build in CI will still fail due to how I configured it. `pulp`'s CI tests work on Bash, not PowerShell, so I'm not sure how to translate that to here.

JordanMartinez:
Not sure if [2e79469](https://github.com/purescript/spago/pull/866/commits/2e794699367057e74b6997751285b3e7fab561ec) will make CI pass, but I wanted to push the changes here for now since it's late. Once the alpha-04 release is made, I think this will build if CI is restarted. If someone gets to that before I do, cool. If not, I'll work on it tomorrow.

JordanMartinez:
@sigma-andex I've fixed a few CI tests, but I'm not sure how to proceed with this one:
```
  1) Spago.purs-0.15, spago bundle-app, Spago should bundle successfully
       expected: "(() => {\n  // output/Effect.Console/foreign.js\n  var log = function(s) {\n    return function() {\n      console.log(s);\n    };\n  };\n\n  // output/Main/index.js\n  var main = /* @__PURE__ */ log(\"\\u{1F35D}\");\n\n  // index.js\n  main();\n})();\n"
        but got: "(() => {\n  // output/Effect.Console/foreign.js\n  var log = function(s) {\n    return function() {\n      console.log(s);\n    };\n  };\n\n  // output/Main/index.js\n  var main = /* @__PURE__ */ log(\"\\u{1F35D}\");\n\n  // <stdin>\n  main();\n})();\n"
        
To rerun use: --match "/Spago/purs-0.15/spago bundle-app/Spago should bundle successfully/"
```
Does it matter that the comment is the only diff? I assume not and we could just update the expected result to match the actual one, right?
```diff
// index.js
// <stdin>
```

JordanMartinez:
Huh... 
Mac build failure:
```
/Users/runner/work/_temp/ce481860-090b-4be9-8704-db6d579aca74.sh:
  line 2: /Users/runner/bin/purescript-v0.15.0-alpha-04/purs: cannot execute binary file
```
Not sure about the Mac build, but the Windows build is what I experienced in `pulp`. Windows build failure:
```
internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                                  ^
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'd:'
    at new NodeError (internal/errors.js:322:7)
    at Loader.defaultResolve [as _resolve] (internal/modules/esm/resolve.js:859:11)
    at Loader.resolve (internal/modules/esm/loader.js:89:40)
    at Loader.getModuleJob (internal/modules/esm/loader.js:242:28)
    at ModuleWrap.<anonymous> (internal/modules/esm/module_job.js:76:40)
    at link (internal/modules/esm/module_job.js:75:36) {
      code: 'ERR_UNSUPPORTED_ESM_URL_SCHEME'
    }
[error] Tests failed: exit code: 1\n")
```

JordanMartinez:
Looks like there's a related 'space in file name' bug with spago test on windows. After running the `spago bundle-app --to bundle-app-esm.js`, there's no corresponding `bundle-app-esm.js` file:
```
"STDOUT: output\npackages.dhall\nspago.dhall\nsrc\ntest\n"
"STDERR: "
```

JordanMartinez:
Anyone with a Windows build able to debug CI?

JordanMartinez:
Looks like `echo` might have issues?
```
  test\SpagoSpec.hs:729:13: 
  1) Spago.purs-0.15, spago bundle-app, Spago should bundle successfully
       expected: "(() => {\n  // output/Effect.Console/foreign.js\n  var log = function(s) {\n    return function() {\n      console.log(s);\n    };\n  };\n\n  // output/Main/index.js\n  var main = /* @__PURE__ */ log(\"\\u{1F35D}\");\n\n  // <stdin>\n  main();\n})();\n"
        but got: "(() => {\n  // <stdin>\n  \"import { main } from './output/Main/index.js'; main();\";\n})();\n"

  To rerun use: --match "/Spago/purs-0.15/spago bundle-app/Spago should bundle successfully/"
```

JordanMartinez:
I got access to a Windows machine. I found that this change makes the tests pass on Windows, but then makes it fail on *nix:
```diff
- "echo \"import { main } from './output/" <> moduleName <> "/index.js'; main();\" | "
+ "echo   import { main } from './output/" <> moduleName <> "/index.js'; main();   | "
```

sigma-andex:
Ok no idea. In think esbuild is for some reason not resolving the import on Windows...

src/Spago/Build.hs
```diff
@@ -376,23 +412,29 @@ runBackend maybeBackend RunDirectories{ sourceDir, executeDir } moduleName maybe
...
+      Browser -> " --platform=browser"
+      Node -> " --platform=node"
     minifyOpt = case minify of
       NoMinify -> ""
       Minify -> " --minify"
+    srcMapOpt = case srcMap of
+      WithSrcMap -> " --sourcemap"
+      WithoutSrcMap -> ""
+    echoLine = case System.Info.os of
+      "mingw3" -> "echo import { main } from './output/" <> moduleName <> "/index.js'; main(); | "
```
sd-yip, line 429:
```suggestion
      "mingw32" -> "echo import { main } from './output/" <> moduleName <> "/index.js'; main(); | "
```

sigma-andex: Geez I’m blind. Thank you very much!

JordanMartinez (approved):
LGTM!

sigma-andex:
@JordanMartinez do you know why two steps are still not running?

f-f:
@sigma-andex they are the same steps that are running now, but with the old name that is expected by GitHub to consider CI "passing" - so we just need to configure the new ones to be expected instead

I'll have a review at this later today or tomorrow

JordanMartinez:
@sigma-andex Here's another solution to that IIFE vs ESM issue you brought up earlier. In other words, we can first output the wrapper to the target file path and then use that as both the input and output of `esbuild`, so long as we use the `--allow-overwrite` flag:
```console
$ cat foo.js 
export function main() {
  console.log("this works");
}
$ cat bar.js 
import { main } from "./foo.js"

main();
$ esbuild bar.js --bundle --outfile=bar.js --allow-overwrite

  bar.js  105b 

⚡ Done in 9ms
$ cat bar.js 
(() => {
  // foo.js
  function main() {
    console.log("this works");
  }

  // bar.js
  main();
})();
```

sigma-andex:
sorry was a bit late yesterday, but I think I understand your approach now. use the same file-based approach that we used for `spago run`, which avoids the windows problematics, and instead of the separate `run.js` file you just do in-file replacement  using the target outfile. However I think there is a problem when you have a different target path, for example like `../out.js` or `./dist/out.js`, then the `import` path would be incorrect in that file. So I think either the import path would need to be updated, or the much simpler alternative would be to either create a `.spago/bundle.js` or reuse the `.spago/run.js`. I vote for going the simple way here.

JordanMartinez:
> However I think there is a problem when you have a different target path, for example like `../out.js` or `./dist/out.js`, then the `import` path would be incorrect in that file. So I think either the import path would need to be updated, or the much simpler alternative would be to either create a `.spago/bundle.js` or reuse the `.spago/run.js`. I vote for going the simple way here.

Ah! Good catch! Perhaps we should write the 'echo' content to `.spago/bundle-app.js`, then run `esbuild` on that file, and then move the file to the target path's location? Regardless, we'll want to write a test for that.

JordanMartinez:
Coming back to this, I'm actually wondering why I didn't just continue using the `echoLine` approach and add `--format=esm` to the esbulid command for the `bundle-app` path. Using `echo` avoids the potential issue of spaces in file names.

sigma-andex:
The iife vs esm format is independent of whether there is a space or not in the filename. But I'm fine with just using esm everywhere. As you said if someone needs iife they can just do another in-place bundling using iife. And if a lot of people demand it we can still add another cli option for it. So to me this PR looks now like it's good to go.

.github/workflows/build.yml
```diff
@@ -112,6 +120,24 @@ jobs:
...
+
+      - name: Download PureScript 0.15.0
+        shell: bash
+        env:
+          DL_FILE: ${{ matrix.purs_0_15_download_file }}
+          BIN_FILE: ${{ matrix.purs_0_15_binary_file }}
+        run: |
+          DIR="$HOME/bin/purescript-$PS_0_15_0_VERSION"
+          mkdir -p "$DIR"
+          curl -o "$DIR/purescript.tar.gz" -L "https://github.com/purescript/purescript/releases/download/$PS_0_15_0_VERSION/$DL_FILE.tar.gz"
```
f-f, line 135: I guess we have to do this because the alphas are not on NPM, right? What if we'd publish those (but not make them install by default of course) on NPM as well?

src/Spago/Build.hs
```diff
@@ -323,15 +342,22 @@ runBackend
 runBackend maybeBackend RunDirectories{ sourceDir, executeDir } moduleName maybeSuccessMessage failureMessage extraArgs = do
   logDebug $ display $ "Running with backend: " <> fromMaybe "nodejs" maybeBackend
   BuildOptions{ pursArgs } <- view (the @BuildOptions)
-  isES <- Purs.hasMinPursVersion "0.15.0"
-  let postBuild = maybe (nodeAction isES $ Path.getOutputPath pursArgs) backendAction maybeBackend
-  build (Just postBuild)
+  isES <- Purs.hasMinPursVersion "0.15.0-alpha-01"
+  nodeEsSupport <- hasNodeEsSupport
+  case (isES, nodeEsSupport) of
+    (True, Unsupported nv) -> die [ "Unsupported Node.js version: " <> display (Version.prettySemVer nv), "Required Node.js version >=12." ]
```
f-f, line 348: Usecase: we might be running an alternate backend and not have node installed in the first place. Here we'd kill the program, while we were supposed to keep running!
So this check should go inside the `nodeAction`.

src/Spago/Build.hs
```diff
@@ -376,25 +412,29 @@ runBackend maybeBackend RunDirectories{ sourceDir, executeDir } moduleName maybe
...
+    esbuildBase = esbuild <> platformOpt <> minifyOpt <> srcMapOpt <> " --format=esm --bundle --outfile=" <> targetPath
     cmd = case withMain of
-      WithMain ->
-        "echo \"import { main } from './output/" <> moduleName <> "/index.js'\nmain()\" | "
-        <> esbuild <> " --platform=" <> platformOpt <> minifyOpt <> " --bundle "
-        <> " --outfile=" <> targetPath
+      WithMain -> do
+        let
+          echoLine = case SysInfo.os of
+            "mingw32" -> "echo import { main } from './output/" <> moduleName <> "/index.js'; main(); | "
```
f-f, line 433: Why are we running `echo` instead of writing to a file? I thought this change was about running from a file again?

test/SpagoSpec.hs
```diff
@@ -5,831 +5,880 @@ import qualified Data.Text          as Text
...
+  let
+    purs0_15_0TestMsg = "purs-0.15"
+    whenUsingEsModulesFixPackagesDhall = when usingEsModules $ do
+      -- The prepare-0.15 package set's contents can change.
+      -- So, we copy an unfrozen one into the directory
+      -- and then freeze it before running `spago init`.
+      cp "../fixtures/packages-prepare-0-15.dhall" "packages.dhall"
+      dhall ["freeze", "packages.dhall"] >>= shouldBeSuccess
+
+  around_ (setup "spago-test") $ do
```
f-f, line 42: Adding these few lines here in the beginning is causing a huuge diff, which makes it really hard to review.
Can we do something (e.g. extract this in a function) to try keeping the same indentation as before? We can eventually fix this indentation in a subsequent change.

test/fixtures/packages-prepare-0-15.dhall
```diff
@@ -0,0 +1,104 @@
+{-
```
f-f, line 1: Can we remove this big comment?

sd-yip:
@sigma-andex @JordanMartinez @f-f
Hi. Sorry to interrupt. The Turtle library accepts an extra argument as the standard input that might be useful in this case.
Required changes for using such a feature are described in this PR. I would appreciate it if you could take a look. Thanks!
https://github.com/working-group-purescript-es/spago/pull/3

sigma-andex:
@sd-yip sorry I haven't seen your PR before. Looks good to me! @JordanMartinez any comments from your side or can we merge this?

.github/workflows/build.yml
JordanMartinez: We haven't done this in the past. I'd bring it up with the core team.

src/Spago/Build.hs
JordanMartinez: Not quite.

When `spago run` or `spago test` is called, then we do need to write the wrapper to a file, so that we call `node run.js` rather than `node --eval "..."`. CLI programs will drop the first two args (path to Node executable, path to entry file) to get the CLI args to the program. When using a file (i.e. `node run.js`), the CLI program will get its args correctly. When using a script (i.e. `node --eval ...`), Node only prefixes the CLI args with one arg (path to Node executable) but not a second arg. Thus, the CLI program will drop its first arg, thinking its the path to the script. So, we have to use a file there.

The above situation is for a different command: `spago bundle-app`. We're trying to bundle the outputted JavaScript code via `esbuild`. However, let's say we run `esbuild --bundle output/Main/index.js --outfile=index.js` and then `node index.js`. When we run `index.js`, nothing will happen because the `main` function was never called (e.g. `main()`). Thus, the above wrapper code is needed to call `main`. Since we're bundling the echo line, which imports `output/Main/index.js`, the resulting `index.js` file will actually run instead of sitting there looking pretty.

The reason why we are using an echo line here rather than writing to a file directly is to avoid the potential 'space in file name' issue that caused problems for `spago run` and other related file system issues. However, the echo line is really just a bad way of passing in the desired `stdin` directly to the `esbuild` command via the Process API, which is what https://github.com/working-group-purescript-es/spago/pull/3 fixes.

JordanMartinez:
> @sd-yip sorry I haven't seen your PR before. Looks good to me! @JordanMartinez any comments from your side or can we merge @sd-yip 's change ?

Other than 'thanks!', nope! :smile:

test/SpagoSpec.hs
JordanMartinez: If you look at the commit history, this diff isn't bad.
- https://github.com/purescript/spago/pull/866/commits/b014d28ec4691df6136e37636d7d89599e521e17
- https://github.com/purescript/spago/pull/866/commits/ea010a5a010ebc4067d6409419c3db6f7e28297b
- https://github.com/purescript/spago/pull/866/commits/918ba6b2918069d10161be7ad6d66423467b04f5
- https://github.com/purescript/spago/pull/866/commits/79766363289149bad13e3f0626df111e4f09225e

If you're looking at the entire diff, then yes, I agree. However, I'm not entirely sure how this should be fixed. Could you provide more guidance?

.github/workflows/build.yml
```diff
@@ -8,18 +8,26 @@ on:
...
 env:
   STACK_VERSION: '2.7.3'
+  PS_0_15_0_VERSION: 'v0.15.0-alpha-04'
 
 jobs:
   build:
     runs-on: ${{ matrix.os }}
     container: ${{ matrix.image }}
     strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
JordanMartinez: Yeah, this was used primarily for debugging.

test/fixtures/packages-prepare-0-15.dhall
JordanMartinez: Fixed in latest commits.

src/Spago/Build.hs
JordanMartinez: Fixed in latest commits. However, I'd guess the fix grants too many capabilities to `build`'s `postBuild` action. I'm not familiar with all the different capabilities, so I've just gone with a simple fix for now.

.github/workflows/build.yml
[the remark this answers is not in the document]
JordanMartinez: Fixed in latest commits.

src/Spago/Build.hs
f-f: That's alright, thanks!

test/SpagoSpec.hs
f-f: Something like this: (disclaimer: I haven't tried compiling it)
```haskell
-- move these to the top level
purs0_15_0TestMsg = "purs-0.15"

fixPackagesDhall usingEsModules = when usingEsModules $ do
    -- The prepare-0.15 package set's contents can change.
    -- So, we copy an unfrozen one into the directory
    -- and then freeze it before running `spago init`.
    cp "../fixtures/packages-prepare-0-15.dhall" "packages.dhall"
    dhall ["freeze", "packages.dhall"] >>= shouldBeSuccess

getUsingEsModules = runIO $ do
    pursVersion <- Cmd.getCmdVersion "purs" >>= either die pure
    esmVersion <- either (const $ die "Failed to parse `0.15.0` static version") pure $ Version.semver "0.15.0-alpha-02"
    pure $ pursVersion >= esmVersion

-- then do a bind-desugar, so that we can keep the `around` on the same line
-- and we don't have to indent everything else:
spec = getUsingEsModules >>= \usingEsModules -> around_ (setup "spago-test") $ do
```

JordanMartinez: That almost works, but then you run into the test setup that needs the space in `spago-test`: https://github.com/purescript/spago/pull/866/files#diff-1be330c42c3288a92800364a45bc6722483cb2700996458be112b464f0fde5d2R775

So, you get code like:
```
spec = getUsingEsModules >>= \usingEsModules -> around_ (setup "spago-test") $ do
  describe "..."
  describe "..."
  
  around_ (setup "spago test") $ do
    describe "..."
```
which is why I indented it.

If you break it into two parts, then this is actually doable

```
spec = do
  usingEsModules <- getUsingEsModules
  noSpacesInParentDir usingEsModules
  spacesInParentDir usingEsModules

noSpacesInParentDir usingEsModules = around_ (setup "spago-test") $ do
  describe "..."
  describe "..."
  
spacesInParentDir usingEsModules =  around_ (setup "spago test") $ do
    describe "..."
```

Since your concern is the diff, I'm going to try that out.

f-f: It might be worth trying a different approach (though I might be missing something): since all we care about is having a space in the path, why not just create a folder (with a space) and do the thing in there for the few tests that we care about, instead of parameterizing the whole setup?

JordanMartinez: Ok. The diff is definitely much cleaner now.

JordanMartinez: > It might be worth trying a different approach (though I might be missing something): since all we care about is having a space in the path, why not just create a folder (with a space) and do the thing in there for the few tests that we care about, instead of parameterizing the whole setup?

Because your `checkFixture` function assumes that the `fixtures` directory is always reachable via a `../fixtures` path. I suppose you could do a recursive ascent that always checks for a `fixture` directory at the current level and, if it's not there, to go up one level before calling itself again.

JordanMartinez: Ok. That's about as clean of a diff as I can get.

.github/workflows/build.yml
JordanMartinez: Could you clarify why that's an issue?

f-f:
Thanks @JordanMartinez for improving the diff! I'll have another look now :slightly_smiling_face:

.github/workflows/build.yml
f-f: It's churn, it's error prone, and I assume it's likely something that I'll have to take care of myself.
So trying to minimize the churn on this right now is less work for future me, unless you or someone else involved in this PR would like to do the update once 0.15 is out.
My opinion is that getting the release out on npm is strictly less work, but I might be missing something.

f-f (approved):
Looks great now! Thanks everyone :clap: 

I think this is good to merge once we figure out the NPM stuff.

.github/workflows/build.yml
JordanMartinez: Sorry for taking a few days before responding to this. I've been focusing on the release script.

> My opinion is that getting the release out on npm is strictly less work, but I might be missing something.

After thinking about this for a while, I think I agree with you. Since CI currently builds, I'd say the above statement is not necessarily true _in the short term_. However, I do think it's true in the _long-term_ when updating tools to work on a new breaking change compiler release. For example, it took me a while both in `pulp` and here to figure out what commands I have to run to correctly install the binary in CI. Delegating that to `npm` is a lot easier rather than figuring this out every time we do a breaking change PS release since debugging CI is a pain. Moreover, by having pre-releases available via `npm`, more people are likely to try it out and find potential problems with it before the core team does.

The only concern I have with this is if a release is done improperly and the 'next' release points to the pre-release rather than the last stable one. I don't have much experience with publishing to NPM, so how would we do that?

> It's churn, it's error prone, and I assume it's likely something that I'll have to take care of myself.
> So trying to minimize the churn on this right now is less work for future me, unless you or someone else involved in this PR would like to do the update once 0.15 is out.

Even if we did publish the pre-releases to NPM, there's still a lot more work than just what this PR entails before spago (and all of its CI) is fully updated to 0.15.0. Since you're already busy with other things, I can see how this just feels like another TODO in an already busy life. @sigma-andex, would you be willing to work with me to get this done?

Once `0.15.0` is actually released, I think there are a couple of changes we would need to make:

First, we would replace the custom download script with just `npm install -g purescript@0.15.0`. In other words, this...:
```
- name: Download PureScript 0.15.0
  shell: bash
  env:
    DL_FILE: ${{ matrix.purs_download_file }}
    BIN_FILE: ${{ matrix.purs_binary_file }}
  run: |
    DIR="$HOME/bin/purescript-$PS_0_15_0_VERSION"
    mkdir -p "$DIR"
    curl -o "$DIR/purescript.tar.gz" -L "https://github.com/purescript/purescript/releases/download/$PS_0_15_0_VERSION/$DL_FILE.tar.gz"
    tar -xvf "$DIR/purescript.tar.gz" -C "$DIR" --strip-components 1 "purescript/$BIN_FILE"
    chmod a+x "$DIR/$BIN_FILE"

- name: Run tests (PureScript >= 0.15.0)
  shell: bash
  run: |
    PATH="$HOME/bin/purescript-$PS_0_15_0_VERSION:$PATH"
    ./scripts/fix-home stack test --ta "--match purs-0.15"
```
would be updated to this:
```
- name: Download PureScript 0.15.0
  shell: bash
  run: npm install -g purescript@0.15.0

- name: Run tests (PureScript >= 0.15.0)
  shell: bash
  run: ./scripts/fix-home stack test --ta "--match purs-0.15"
```

We'd also drop the `env.PS_0_15_0_VERSION` and `jobs.build.strategy.matrix.include.[purs_download_file/purs_binary_file]` stuff.

Second, there are a lot of tests not being run with the `0.15.0` alpha version because we don't have a released 0.15.0 package set yet. So, the tests may also need to be updated.

Third, some packages (e.g. `simple-json`) used throughout the test may take a while to get into the package set. So, even if we have a released 0.15.0 package set, either someone would need to update that in a PR or change the tests to stop using those packages. 

Fourth, there may be some tests we want to continue running when using `0.14.0`  but not `0.15.0` or vice versa (just hypothesizing here).

JordanMartinez: > I don't have much experience with publishing to NPM, so how would we do that?

Answering my own question, it looks like we follow [these instructions](https://gist.github.com/schmich/d04acc4b02b45e489f329cfdf3280a3f).

JordanMartinez: I've opened https://github.com/purescript/purescript/issues/4294 to discuss the publish to NPM idea.

sigma-andex: @JordanMartinez sorry I somehow missed this threat here. Sure, can help you with this task. Let me know how I can help you.

JordanMartinez: I've updated the code to use `npm i -g purescript@next`. Just waiting on CI to build now.

JordanMartinez:
Ok. CI builds! :tada: @f-f Since you previously approved this and the last thing to resolve--the NPM installation--has now been resolved, I'm going to merge this.

JordanMartinez:
Since the most recent commit gets rid of the metadata needed to install and runs `purs` via the tar download approach, CI now expects `build (macOS-latest, macos, purs)` and `build (windows-latest, win64, purs.exe)` builds to occur, which no longer exist. @f-f Not sure how that was changed last time, but just FYI.

thomashoneyman (approved):

f-f (approved):
Wonderful, we're good to go! Thanks everyone for contributing to this :blush:

f-f:
I'll cut a release later today so we get to use this ASAP

PR #870, opened 2022-03-29
Use purs v0.15 logic for versions > 0.15.0-alpha-01

### LINKED ISSUE #869: spago run with purs v0.15.0-alpha-02
I can't get `spago run` or `spago test` working with `purs` v0.15.0-alpha-02. I think it's because on this line `isES` is `false` for the string `"0.15.0-alpha-02"`.

https://github.com/purescript/spago/blob/cc376c85bc6107ec38da2d0917170e046ba5a211/src/Spago/Build.hs#L326

I think this is because the second hyphen makes the semver string illegal. https://hackage.haskell.org/package/versions-5.0.3/docs/Data-Versions.html#t:SemVer

jamesdbrock:
Otherwise `spago run` will use CommonJS instead of ES modules for `purs` v0.15.0-alpha-01, and users will see an error like

```
node:internal/modules/cjs/loader:1146
      throw err;
      ^

Error [ERR_REQUIRE_ESM]: require() of ES Module output/Test.Main/index.js from [eval] not supported.
Instead change the require of index.js in/[eval] to a dynamic import() which is available in all CommonJS modules.
    at [eval]:1:1
    at Script.runInThisContext (node:vm:129:12)
    at Object.runInThisContext (node:vm:305:38)
    at [eval]-wrapper:6:22 {
  code: 'ERR_REQUIRE_ESM'
}

Node.js v17.1.0
```

Resolves #869


sigma-andex:
This is already being fixed here: https://github.com/purescript/spago/pull/866/files

Issue #834, opened 2021-12-27
Attempt retry using exponential backoff when downloading fails


mikesol:
For the past 24-hours or so, GitHub downloads have been crapping out at a rate of 1/100. I suggested on the Discord server that automatic retry (ie 3 retries) with exponential backoff could help get the ball over the net when this happens and @f-f recommended opening an issue. Hopefully I'll be able to implement this soon!

f-f:
Thanks for reporting! I thought we didn't have this, but we do:
https://github.com/purescript/spago/blob/aa7c0de6d903262f69452663c09fad9c441af8d3/src/Spago/GlobalCache.hs#L168-L169

..however it's only for downloading tarballs. In the case of `git` errors there's no retry, and it should be patched somewhere here:
https://github.com/purescript/spago/blob/aa7c0de6d903262f69452663c09fad9c441af8d3/src/Spago/FetchPackage.hs#L144-L158

f-f:
Fixed in #873

PR #873, opened 2022-04-09
Retry git clone (#834)

joe-op:
Retry git clone in case there are intermittent connection failures, based on the suggested solution in https://github.com/purescript/spago/issues/834.

I came across this issue and tested the change during a time when Github connections were occasionally failing and it successfully retried.  I also tested against a local server and spago tried to connect 3 times with this change and 1 without.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Features:
 - Make `spago bundle-app` and `spago bundle-module` use esbuild and es modules for projects >= v0.15
 - Make `spago run` use es modules for projects >= v0.15 
 - Support Glibc versions >= `2.24`
+- Retry git clone up to two times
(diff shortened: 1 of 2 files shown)


f-f (approved):
Looks great, thanks!

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Features:
 - Make `spago bundle-app` and `spago bundle-module` use esbuild and es modules for projects >= v0.15
 - Make `spago run` use es modules for projects >= v0.15 
 - Support Glibc versions >= `2.24`
+- Retry git clone up to two times
```
f-f, line 14:
```suggestion
- Retry git clone up to two times (#834, #873)
```

PR #880, opened 2022-05-01
Use iife when bundling browser apps

sigma-andex:
### Description of the change

TODO

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+Other improvements:
+- Use `iife` format when bundling browser apps (#880)
+
 ## [0.20.8] - 2022-04-27
 
 Features:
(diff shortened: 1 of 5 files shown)


CarstenKoenig:
Not sure how other folks like to use this but I like to add the [global name option](https://esbuild.github.io/api/#global-name) to wrap the exports into an object.
I think this is very close to the way purescript was bundled before and it's really helpful for me when I like to call `main` with arguments (a bit like Elm does with flags) or when I want to expose other functions/entry points into the app.

Maybe we could add a flag to use `--global-name=PS` or even let users choose the name?

sigma-andex:
We had some discussions about the esbuild in the working group and I think the consent was that we didn't want to wrap the whole `esbuild` params list but just the basic . I think a flag to add any additional esbuild params like `--purs-args` would make sense, but I guess that deserves a different issue/pr then.

CarstenKoenig:
Ok I guess users can opt to directly use esbuild in this case - still very much appreciate this PR as it would get rid of an bug/error I had when using `spago bundle-app` without `--minify` when something exports `document` :heart:

f-f (approved):

PR #882, opened 2022-05-05
example GitHub workflow for for building Spago statically

cdepillabout:
### Description of the change

wip

### Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,16 @@ env:
   STACK_VERSION: '2.7.3'
 
 jobs:
+  static:
+    name: statically-linked exe / ubuntu-latest
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: cachix/install-nix-action@v17
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
+    - run: nix-build
+
   build:
     runs-on: ${{ matrix.os }}
     container: ${{ matrix.image }}
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,122 @@
+
+# This builds a statically-linked `spago` executable.
+# It currently only works on Linux.
+
+let
+  nixpkgs-src = builtins.fetchTarball {
+    # This is one of the last commits from https://github.com/NixOS/nixpkgs/pull/169513.
+    # This commit is used because it is for the `haskell-updates` branch.
+    # Hydra builds a GHC for static linking on this branch, so we can get
+    # most dependencies from the NixOS cache.
+    url = "https://github.com/NixOS/nixpkgs/archive/999a0e87f9b1792e229c53602e80bda6dd52c105.tar.gz";
+    sha256 = "1p0vgmdz884gld4p2zcdy7g0w7xznc4kxj80wf5pk2qdxicbnscy";
+  };
+
+  spago-overlay = final: prev: {
+
+    myhaskell = prev.haskell // {
+      packages = prev.haskell.packages // {
+        native-bignum = prev.haskell.packages.native-bignum // {
+          ghc902 = prev.haskell.packages.native-bignum.ghc902.override (oldAttrs: {
+            overrides =
+              final.lib.composeExtensions
+                (oldAttrs.overrides or (_: _: {}))
+                (hfinal: hprev: {
+
+                  aeson = hfinal.aeson_1_5_6_0;
+
+                  # bower-json needs aeson_1_5_6_0 and is marked broken without it.
+                  bower-json = final.haskell.lib.doDistribute (final.haskell.lib.markUnbroken hprev.bower-json);
+
+                  spago =
+                    let
+                      filesToIgnore = [
+                        "default.nix"
+                        ".git"
+                        ".github"
+                        "Makefile"
+                        "result"
+                        ".stack-work"
+                        "stack.yaml"
+                      ];
+
+                      src = builtins.path {
+                        name = "spago-src";
+                        path = ./.;
+                        filter = path: type:
+                          with final.lib;
+                          ! elem (baseNameOf path) filesToIgnore &&
+                          ! any (flip hasPrefix (baseNameOf path)) [ "dist" ".ghc" ];
+                      };
+
+
+                      docsSearchApp_0_0_10 = final.fetchurl {
+                        url = "https://github.com/purescript/purescript-docs-search/releases/download/v0.0.10/docs-search-app.js";
+                        sha256 = "0m5ah29x290r0zk19hx2wix2djy7bs4plh9kvjz6bs9r45x25pa5";
+                      };
+
+                      docsSearchApp_0_0_11 = final.fetchurl {
+                        url = "https://github.com/purescript/purescript-docs-search/releases/download/v0.0.11/docs-search-app.js";
+                        sha256 = "17qngsdxfg96cka1cgrl3zdrpal8ll6vyhhnazqm4hwj16ywjm02";
+                      };
+
+                      purescriptDocsSearch_0_0_10 = final.fetchurl {
+                        url = "https://github.com/purescript/purescript-docs-search/releases/download/v0.0.10/purescript-docs-search";
+                        sha256 = "0wc1zyhli4m2yykc6i0crm048gyizxh7b81n8xc4yb7ibjqwhyj3";
+                      };
+
+                      purescriptDocsSearch_0_0_11 = final.fetchurl {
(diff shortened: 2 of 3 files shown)


cdepillabout:
@f-f This PR is a rough proof-of-concept of adding a GitHub Action job for building Spago statically.

Currently, this PR just adds `static` job to the `build.yaml` GitHub Action.  I'd like you to take a look at it, and if it generally looks alright, then I'd go though and add more documentation, as well as adding some code to the `release.yaml` GitHub Action to actually upload the static binary to the GitHub release.

One thing to notice is that building the statically-linked `spago` binary takes about 20m without using any sort of cache, so that seems basically in-line with the Windows an OSX builds.  You mentioned in https://github.com/NixOS/nixpkgs/pull/171604#issuecomment-1118363612 about potentially having adding a cachix cache.  That's something we could do to speed up the build, but I see it more as a potential future improvement, and not something that is strictly necessary to add in this PR.

.github/workflows/build.yml
```diff
@@ -10,6 +10,16 @@ env:
...
 jobs:
+  static:
+    name: statically-linked exe / ubuntu-latest
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: cachix/install-nix-action@v17
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
+    - run: nix-build
```
cdepillabout, line 21: One potential problem with this is that there is a non-trivial amount of Nix code in `default.nix` to be able to get Spago to compile with Nix.  For instance, overrides to the `aeson` version, unbreaking `bower-json`, the workaround for the `docsSearchApp` thing, etc.  It would be unfortunate for a non-Nix user to send a PR to `spago`, have the Nix build break, and then not know how to fix it.

I'd of course be able to help with stuff like this, but I'm not sure a random contributor would know to ping me.

f-f:
@cdepillabout this looks amazing! ❤️

I have only one question: what would we do when upgrading Haskell dependencies?
My understanding is that the static building infra from Nixpkgs is not always reliable (or at least it has not been so far, but I'd love to hear that we can trust it nowadays), so bumping the Nixpkgs commit to get the latest Haskell dependencies could be risky?
But then I think that if we'd not do it we'd have to rebuild a whole lot of dependencies? (lengthening the build considerably, at least without cache)

default.nix
```diff
@@ -0,0 +1,122 @@
...
+
+let
+  nixpkgs-src = builtins.fetchTarball {
+    # This is one of the last commits from https://github.com/NixOS/nixpkgs/pull/169513.
+    # This commit is used because it is for the `haskell-updates` branch.
+    # Hydra builds a GHC for static linking on this branch, so we can get
+    # most dependencies from the NixOS cache.
+    url = "https://github.com/NixOS/nixpkgs/archive/999a0e87f9b1792e229c53602e80bda6dd52c105.tar.gz";
+    sha256 = "1p0vgmdz884gld4p2zcdy7g0w7xznc4kxj80wf5pk2qdxicbnscy";
+  };
```
cdepillabout, line 13: One problem with building from Nixpkgs like this is that we get the Haskell package versions in whatever commit of Nixpkgs we use.  So with the above commit, we are building with mainly LTS-19, while the `stack.yaml` in this repo is still LTS-18.

While this discrepancy doesn't seem to cause any big problems now, it is at least unfortunate that we're not using the same exact package versions.

([haskell.nix](https://github.com/input-output-hk/haskell.nix) has tools for generating a Haskell package set using versions from a `stack.yaml` and Stackage resolver, but haskell.nix comes with its own set of problems.)

default.nix
```diff
@@ -0,0 +1,122 @@
...
+                                "$sourceRoot/templates/docs-search-app-0.0.10.js" \
+                                "$sourceRoot/templates/purescript-docs-search-0.0.10" \
+                                "$sourceRoot/templates/docs-search-app-0.0.11.js" \
+                                "$sourceRoot/templates/purescript-docs-search-0.0.11"
+                            '';
+                          });
+                    in
+                    final.haskell.lib.compose.generateOptparseApplicativeCompletion
+                      "spago"
+                      spagoDrvFromCabal2nix;
```
cdepillabout, line 106: These overrides are basically taken directly from Nixpkgs.

cdepillabout:
@f-f

> I have only one question: what would we do when upgrading Haskell dependencies? My understanding is that the static building infra from Nixpkgs is not always reliable (or at least it has not been so far, but I'd love to hear that we can trust it nowadays), so bumping the Nixpkgs commit to get the latest Haskell dependencies could be risky? But then I think that if we'd not do it we'd have to rebuild a whole lot of dependencies?

In my opinion, this is basically the biggest problem.

A  few months ago, there was a PR that got static-linking working pretty reliably in Nixpkgs for Haskell executables: https://github.com/NixOS/nixpkgs/pull/162374.  With this PR, most Haskell executables can be built statically (at least with GHC-9).  It seems like there are now enough people interesting in static-linking and cross-compilation in Nixpkgs that the Musl-related infrastructure is pretty solid.  

We also have some checks in our CI to make sure that statically-linking Haskell executables is working. This gives us Musl-built GHCs for static linking from the NixOS cache, so end users (or the `spago` CI) don't have to build GHC themselves.  In my opinion, moving forward, relying on the Haskell static-linking facility in Nixpkgs isn't a crazy idea.

However, the bigger problem is what do to when upgrading Haskell dependencies.  The Haskell packages in Nixpkgs are generally set as the latest LTS release.  There are times when this matches the LTS release that Spago is using, but sometimes Spago lags behind.  In the past, this hasn't seemed to prove a big problem maintaining Spago in Nixpkgs, but like I said above, it would be unfortunate for other contributors to be faced with these kinds of errors.

My suggestion would be to move forward with this PR as-is, but keep an eye out for these types of problems in the future.  If the Nix-related stuff seems to be turning away too many contributors (or taking up too much of our time fixing things), then we could always consider removing it.

default.nix
```diff
@@ -0,0 +1,122 @@
...
+                      purescriptDocsSearch_0_0_11 = final.fetchurl {
+                        url = "https://github.com/purescript/purescript-docs-search/releases/download/v0.0.11/purescript-docs-search";
+                        sha256 = "1hjdprm990vyxz86fgq14ajn0lkams7i00h8k2i2g1a0hjdwppq6";
+                      };
+
+                      spagoDrvFromCabal2nix =
+                        final.haskell.lib.overrideCabal
+                          (hfinal.callCabal2nix "spago" src {})
+                          (drv: {
+                            doCheck = false;
```
cdepillabout, line 77: Since `spago` uses the network during tests, I've had to disable the tests here.

We end up not really testing this Nix-built `spago` binary, which is unfortunate.  We could potentially add some simple tests to the `build.yaml` CI file, like running `spago version`, `spago init`, `spago build`, etc.

default.nix
cdepillabout: That's a good point, and a good reason for us to make sure we use the same package versions.

> Would it make sense to generate the package versions in the stack.yaml from the ones in Nixpkgs? (and of course check that they match in CI)

I think this would make sense, although I don't know of any sort of tool that is able to generate a `stack.yaml` from a given Nixpkgs commit.  I'm not sure how difficult it would be to write something like this.

For now, it would probably be the easiest to just manually keep the Nixpkgs Haskell package set and the resolver from the `stack.yaml` file the same.  This PR uses Nixpkgs at commit https://github.com/NixOS/nixpkgs/commit/999a0e87f9b1792e229c53602e80bda6dd52c105, which is included in PR https://github.com/NixOS/nixpkgs/pull/169513.  At the top of that PR, you can see there is a commit bumping the Nixpkgs Haskell stuff to LTS-19.4.

So maybe I could take a stab at bumping spago to LTS-19.4 (possibly in this PR, or a separate PR).  Then we'd be using the same LTS version and GHC version in both `stack.yaml` and the Nixpkgs Haskell package set.

> This would mean depending on whatever versions of packages coming from Nixpkgs, which is not the best (because we introduce a dependency) but it's pretty alright in the end.

I agree this is somewhat unfortunate.  Ideally there would be a tool that could generate a Nixpkgs Haskell package set from a `stack.yaml` file, but we don't currently have something like that.  (There used to be a tool https://github.com/input-output-hk/stack2nix that would do this, but it hasn't been maintained over the last few years.)

> I assume we would/could tolerate some overrides in case we need them for the build to work.

Yeah, the Nixpkgs Haskell side is somewhat flexible, and it should be possible to add any overrides that we may need.  I'd of course be happy to help out with this.

f-f:
I'll close this as we're about to switch to a PureScript-based Spago.

PR #884, opened 2022-05-07
Fix for non-zero Windows npm/npx exit codes

### LINKED ISSUE #883: exit code always 0 on Windows running with NPM
with npx or npm, anytime I run spago it retuns exit code 0 regardless of build status.  I narrowed down the problem to the `spago` node file:
```javascript
#!/usr/bin/env node
// This file is only used on Windows. In *nix NPM installations it's replaced
// with the actual spago binary.

const cp = require("child_process");

// Ignore SIGINT (Ctrl-C) because we don't want to terminate this process from
// it (#483). Spago.exe will handle it for us
process.on('SIGINT', () => {});

const spago = cp.spawn(__dirname + '/spago.exe', process.argv.slice(2), {stdio: 'inherit'});
spago.on('error', (err) => {
    console.log("Downloading the spago binary failed. Please try reinstalling the spago npm package.");
});
```
If I add to the end of the file 
```javascript
spago.on('exit', (code) => {
    process.exit(code);
});
```
it seems to fix the problem.  Would you accept a PR to make that change?

ntwilson:
### Description of the change

Adjust the node.js script for launching spago when running on Windows so that exit codes are forwarded.  Fixes #883.

I would have liked to implement better testing by e.g., adjusting the `spago` function in test/Utils.hs to use `npx spago` or just invoke `node` directly pointed to the `spago` file - hopefully to observe a test failure with the old code and success with the new code.  But I wasn't able to understand how the test setup worked to try that out. 

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+Bugfixes:
+- Windows: forward the proper exit code when running spago via NPM (#883)
+
 ## [0.20.9] - 2022-05-03
 
 Bugfixes:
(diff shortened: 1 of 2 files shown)


f-f (approved):
Thank you!

Issue #887, opened 2022-05-21
aeson 2.0 compatibility


maralorn:
As of a few days ago, there is a new bower-json version which supports aeson 2.0.

Now to support aeson 2.0 there are only the build errors in spago itself left. I get the following build error, when trying to build with aeson  2.0, but it might be masking more:

```
src/Spago/Bower.hs:112:33: error:
    • Couldn't match type: Data.Aeson.KeyMap.KeyMap Value
                     with: HashMap k0 a0
      Expected: HashMap k0 a0
        Actual: Object
    • In the second argument of ‘HashMap.member’, namely ‘info’
      In the expression: HashMap.member "version" info
      In a stmt of a 'do' block:
        if HashMap.member "version" info then
            pure $ Bower.VersionRange $ "^" <> version
        else
            pure $ Bower.VersionRange $ repo <> "#" <> version
    |
112 |     if HashMap.member "version" info
    |                                 ^^^^
```

f-f:
@maralorn I'd happily merge a patch with the updates if you'd like to have a go at this!
Otherwise I'll take care of it, but it might take a while before I get to this.

maralorn:
I am not in hurry and don’t use spago myself. It just came up during NixOS maintenance.

f-f:
Fixed in #901

PR #889, opened 2022-06-06
Bower.hs compatibility with Aeson 2

VlkrS:
### Description of the change

Replace Data.HashMap.Strict with Data.Aeson.KeyMap
See also issue #887


f-f:
@VlkrS could you adjust the build plan in `stack.yaml` to pull in the new Aeson? The current one is using `aeson-1.5.6.0`

f-f:
Fixed in #901

PR #914, opened 2022-10-07
Integrate PSA into codebase

### LINKED ISSUE #904: Integrate psa into Spago
Since the upcoming Spago will be in PureScript, we now have the occasion of integrating `psa` directly [as a library](https://github.com/natefaubion/purescript-psa-utils), which is strictly better than the current situation of shelling out to an external executable.

JordanMartinez:
### Description of the change

Fixes #904. Currently a WIP.

### Checklist:

### DIFF
--- a/spaghetto/bin/src/Flags.purs
+++ b/spaghetto/bin/src/Flags.purs
@@ -4,12 +4,108 @@ import Spago.Prelude
 
 import ArgParse.Basic (ArgParser)
 import ArgParse.Basic as ArgParser
+import Data.Set as Set
+import Data.String as String
+import Psa (ErrorCode, StatVerbosity(..))
 
 selectedPackage ∷ ArgParser (Maybe String)
 selectedPackage =
   ArgParser.argument [ "--package", "-p" ]
     "Select the local project to build"
     # ArgParser.optional
+
+psaCensorWarnings ∷ ArgParser Boolean
+psaCensorWarnings =
+  ArgParser.flag [ "--censor-warnings" ]
+    "Censor all warnings"
+    # ArgParser.boolean
+
+psaCensorLib ∷ ArgParser Boolean
+psaCensorLib =
+  ArgParser.flag [ "--censor-lib" ]
+    "Censor warnings from library sources"
+    # ArgParser.boolean
+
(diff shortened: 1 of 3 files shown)


JordanMartinez:
Feel free to provide any feedback while I work on this. For example:
- should `flag` args always have a `ArgParser.boolean x` to be explicit about what the default is? Or do we defer to the defaults provided by `ArgParser` (currently `false`)?
- where should I place constant values (e.g. `.psa-stash`)?
- is prefixing all `psa` args with `psa` ok? Or should that be handled differently?

f-f:
Thanks for picking this up!

> should flag args always have a `ArgParser.boolean x` to be explicit about what the default is? 

I don't have an opinion on this and providing defaults might be sensible for certain things and less sensible for others, but in any case keep in mind that for some/all the flags we should also offer the possibility to configure them from the config file. See e.g. how this is done for `bundle`'s arguments:

https://github.com/purescript/spago/blob/2cb7edfe12c6751e94d1ad6f193b1b4f9e112398/spaghetto/bin/src/Main.purs#L246-L254

> where should I place constant values (e.g. `.psa-stash`)?

Are you looking for a place to stash files? The `.spago` folder is there for that, feel free to make a subdirectory `.spago/psa` if you need to store more stuff.

> is prefixing all `psa` args with `psa` ok? Or should that be handled differently?

Definitely ok to have this prefix in the source, and we might want to even prefix some/all the flags with `psa`, so that it's clear that they control psa's behaviour

JordanMartinez:
> Are you looking for a place to stash files? The .spago folder is there for that, feel free to make a subdirectory .spago/psa if you need to store more stuff.

How about we have a special `spago-specific` directory that's easily accessible (in that it appear above all the other dependency folders)? For example `.spago/_internal/psa/<stash file>`. This could also store the `.spago/_internal/run.js` file. Otherwise, if I want to look at the stash file, I may have to scroll through a large number of dependency folders before I see the `psa` folder.

f-f:
> I may have to scroll through a large number of dependency folders before I see the psa folder.

I've been very annoyed by this in the past too, which is why spaghetto stores packages in `.spago/packages` 🙂

JordanMartinez:
Succeeded by #955

PR #924, opened 2022-11-26
Version parsing fixes

MonoidMusician:
Quick fixes to get it working for me again:
- My version of git, 2.36.1, ironically does not have the `-v` flag
- Drop the purs version string after a space, to accomodate dev builds which have a version string like `0.15.6 [development build; commit: 8da7e96005f717f03d6eee3c12b1f1416659a919]`

### DIFF
--- a/spaghetto/src/Spago/Purs.purs
+++ b/spaghetto/src/Spago/Purs.purs
@@ -38,6 +38,7 @@ getPurs =
     Left err -> do
       logDebug $ show err
       die [ "Failed to find purs. Have you installed it, and is it in your PATH?" ]
+    -- Drop the stuff after a space: dev builds look like this: 0.15.6 [development build; commit: 8da7e96005f717f03d6eee3c12b1f1416659a919]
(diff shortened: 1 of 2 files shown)


f-f (approved):
Thanks!

spaghetto/src/Spago/Purs.purs
```diff
@@ -35,7 +38,7 @@ getPurs =
     Left err -> do
       logDebug $ show err
       die [ "Failed to find purs. Have you installed it, and is it in your PATH?" ]
-    Right r -> case Version.parseVersion Version.Lenient r.stdout of
+    Right r -> case Version.parseVersion Version.Lenient (fromMaybe "" (Array.head (String.split (String.Pattern " ") r.stdout))) of
```
f-f, line 41:
```suggestion
    -- Drop the stuff after a space: dev builds look like this: 0.15.6 [development build; commit: 8da7e96005f717f03d6eee3c12b1f1416659a919]
    Right r -> case Version.parseVersion Version.Lenient (fromMaybe "" (Array.head (String.split (String.Pattern " ") r.stdout))) of
```

PR #929, opened 2023-01-11
Upgrade spago@next to the latest registry

f-f:


spaghetto/src/Spago/Config.purs
```diff
@@ -29,54 +30,55 @@ module Spago.Config
...
 import Spago.Prelude
 
 import Affjax.Node as Http
 import Affjax.ResponseFormat as Response
 import Affjax.StatusCode (StatusCode(..))
-import Data.Argonaut.Core as Core
 import Data.Array as Array
-import Data.Bifunctor (lmap)
+import Data.Codec.Argonaut as CA
+import Data.Codec.Argonaut as Codec
```
thomashoneyman, line 43: You could remove this and only have the `CA` import.

spaghetto/src/Spago/Config.purs
```diff
@@ -87,35 +89,80 @@ type PackageConfig =
   , publish :: Maybe PublishConfig
   }
 
+packageConfigCodec :: JsonCodec PackageConfig
+packageConfigCodec = CAR.object "PackageConfig"
```
thomashoneyman, line 93: You don't need this to be in any explicit order, do you? I'd just keep in mind that the record codecs (as with the argonaut ones) are alphabetically ordered by field name, so if you're expecting a specific field order you'll need to encode that manually (see `Registry.Manifest`).

spaghetto/src/Spago/Config.purs
```diff
@@ -248,78 +307,61 @@ instance Semigroup Dependencies where
...
+          Nothing -> Left name
+          Just r -> Right (Map.singleton name r)
+      )
+      $ Map.toUnfoldable deps :: Array _
+
+  from :: Array (Either PackageName (Map PackageName Range)) -> Dependencies
+  from = Dependencies <<< Map.fromFoldable <<< map
+    ( case _ of
+        Left name -> Tuple name Nothing
+        Right m -> rmap Just $ unsafeFromJust (List.head (Map.toUnfoldable m))
```
thomashoneyman, line 336:
```suggestion
        Right m -> map Just $ unsafeFromJust (List.head (Map.toUnfoldable m))
```

thomashoneyman (approved):
Looks pretty straightforward! 👍

spaghetto/src/Spago/Config.purs
f-f: I always considered the Functor instance of Tuple a footgun, and prefer to use the more explicit `rmap`

spaghetto/src/Spago/Config.purs
f-f: Ah, good point. I don't think I personally care to order fields, do you think the config should be sorted?

spaghetto/src/Spago/Config.purs
f-f: Good catch!

spaghetto/src/Spago/Config.purs
thomashoneyman: If this is something the user is going to write then ideally we’d preserve their ordering in the source — if not, then no I don’t think it matters

f-f: Ah, I'm afraid that for this codecs that are dealing with Yaml we'd be in the hands of the underlying Yaml library, so it would not matter the order that we set here. I'd need to double check.

Are we able to specify a codec that preserves users' insertion order?

thomashoneyman: We could write one that encodes an `Array (Tuple String Json)` into an object ordered the same as an array, so if you store the user's field order somewhere this codec would preserve that order.

thomashoneyman: Or we could always sort it, for the time being, in an ordering that would be typical for a config.

f-f: It looks like the yaml library that we use does not sort things by default, [and that can be configured](https://eemeli.org/yaml/#schema-options). I had my configs sorted though, probably by the editor. I am afraid that the same thing would happen to users, rendering our nice sorting useless and even counterproductive. So let's just dump things as is for now, and look into this sorting at a later stage.

PR #931, opened 2023-01-13
Remove bin/index.js since it is no longer usable

MonoidMusician:
bundle.js calls main directly and does not export it


f-f:
Ah brilliant. I can't recall the last time I used this one, I usually use `./bin/index.dev.js`

PR #932, opened 2023-01-13
Fix NodeJS version parsing

MonoidMusician:
Quick fix for NodeJS version parsing by stripping the `v` prefix. I verified that this runs locally.

The full LenientVersion parsing in the registry does a few more things, but it's in the app directory:

https://github.com/purescript/registry-dev/blob/c0f9f65371a7f8baf0335272c14164d318b6f37f/app/src/App/Legacy/LenientVersion.purs#L61-L69

### DIFF
--- a/spaghetto/src/Spago/Command/Run.purs
+++ b/spaghetto/src/Spago/Command/Run.purs
@@ -45,7 +45,7 @@ nodeVersion =
     Left err -> do
       logDebug $ show err
       die [ "Failed to find node. Have you installed it, and is it in your PATH?" ]
-    Right r -> case Version.parse r.stdout of
+    Right r -> case Version.parse $ fromMaybe r.stdout (String.stripPrefix (String.Pattern "v") r.stdout) of
       Left _err -> die $ "Failed to parse NodeJS version. Was: " <> r.stdout
       Right v ->
         if Version.major v >= 13 then


f-f:
Would it be worth porting the Lenient stuff to the lib? cc @thomashoneyman

thomashoneyman:
We’ve only been using it for legacy work — where we have to parse git tags — so I was thinking we might eventually remove it. But I think they’re fairly simple modules so we could migrate them to 

Registry.Version.Lenient
Registry.Range.Lenient

We could also rename the “foreign” package to “core” and use that as a sort of unofficial dumping ground of stuff we want to share but won’t release as an official library (foreign bindings, lenient parsing, the legacy data types)

f-f:
Oh yeah it's even better if we can move these to the foreign package! Then we don't have to worry about making breaking changes to it.

thomashoneyman:
I can take care of that — but I'll rename it to `core/Foreign`, `core/Legacy`, etc., as it will no longer be strictly foreign bindings.

f-f:
Sure, I'll get to that tomorrow morning

f-f:
This was fixed in https://github.com/purescript/spago/commit/e6e3f0ea200a4ea2ac162669e127e9fa2a7cc83c

PR #934, opened 2023-01-20
Parse legacy package set version with leading v

nwolverson:
Quick fix to unbreak parsing of legacy package sets, where metadata version is eg. v0.15.7

### Description of the change

TODO

### Checklist:


f-f:
This is the same issue that we hit in https://github.com/purescript/spago/pull/932, I'll apply the general fix (i.e. using lenient parsing)

f-f:
This was fixed properly in https://github.com/purescript/spago/commit/e6e3f0ea200a4ea2ac162669e127e9fa2a7cc83c

PR #936, opened 2023-01-23
Change config key set -> package_set to be more explicit

MonoidMusician:
As discussed in https://github.com/purescript/registry-dev/issues/584#issuecomment-1384627876

I searched through the source for `set`, can you think of any other places that need to change?


f-f (approved):
Looks great, thank you!

PR #939, opened 2023-02-01
Replace JS `execa` library with PS port

JordanMartinez:
### Description of the change

Drops the JS `execa` library in favor of the port I did of it and all of its dependencies in `purescript-node-execa`.

A few questions/comments I have:
- I noticed in `Cmd.purs` that you pipe'd the child process' `stdin` into the parent process' `stdin`. I'm guessing that was a copy-paste mistake and that you meant to do the opposite direction?
- I'd like to make another release of `node-execa` that takes adds another util `stdin.pipeParentProcess` for piping the parent process stdin into the child process' stdin. That'll clean this up a bit
- I went ahead and implemented this, but I'm sure I'm breaking some of your guidelines for where to store things (e.g. in `Cmd.purs` as opposed to elsewhere). Let me know what I need to change.

### Checklist:


JordanMartinez:
Note: I haven't tested this locally. These are simply the changes needed to get the code to compile and drop the JS deps.

f-f:
Lovely, thank you for doing this!

I would like to keep the changes local to the `Cmd` module as much as we can, for a few reasons:
- I am quite pleased with the current declarative interface to spawn processes, where I can say how I would like the pipes to be set up, and do not have to worry about the side effects to make it happen. While I do not necessarily want to push that to your library, I am of the opinion that such API can prevent bugs when refactoring
- If there is ever the need to swap the underlying library again, I am sure that all the changes are localised to the `Cmd` module, making the job much easier

f-f:
> I noticed in `Cmd.purs` that you pipe'd the child process' stdin into the parent process' stdin. I'm guessing that was a copy-paste mistake and that you meant to do the opposite direction?

Well spotted 😄 I now wonder how that did not cause any bugs? I really need to set up tests on this thing 😂

spaghetto/spago.yaml
```diff
@@ -63,3 +64,6 @@ workspace:
       git: https://github.com/purescript/registry-dev.git
       ref: 8113deecae25410ba62c31a5842c6ebbf1c12461
       subdir: foreign
+    node-execa: 1.1.0
+    node-event-emitter: 1.0.1
+    node-human-signals: 1.0.0
```
JordanMartinez, line 69: `node-execa` is now in the package set, but I'm not sure how to indicate that here...

spaghetto/src/Spago/Cmd.js
```diff
@@ -1,48 +0,0 @@
...
-  cp.stderr.pipe(process.stderr);
-}
-
-export function joinImpl(cp, onError, onSuccess) {
-  return cp.then((data) => onSuccess(data))
-    .catch((err) => {
-      // WTFFFF, this is because on ENOENT we don't get an exitCode
-      if (err.exitCode === undefined) {
-        err.exitCode = -1;
-      }
```
JordanMartinez, line 39: What specifically triggered this case? I need to see if `node-execa` handles it properly.

JordanMartinez:
K, this just updates `Cmd.purs` and fixes the `exitCode` type change. If I can't drop the `Maybe`, I may need to make additional changes in `node-execa`.

spaghetto/spago.yaml
f-f: You should bump the package set to one that contains the library, in line 55.
Latest is [`11.10.0`](https://github.com/purescript/registry/blob/main/package-sets/11.10.0.json), and contains all the libraries added here.
It has `dodo-printer` too, so you can remove line 57 as well.

spaghetto/src/Spago/Cmd.js
f-f: This is because your implementation has a `Maybe ExitCode`, where you get `Nothing` if the executable you're trying to call does not exist.
The idea is that I always wanted an exit code returned to avoid having the application layer deal with a `Nothing`, but I think it's fine to have a `Maybe` in there as long as it's documented.

spaghetto/src/Spago/Cmd.js
```diff
@@ -1,48 +0,0 @@
-import { execa, execaCommand } from "execa";
-
-const mkOptions = (opts) => {
-  let options = { preferLocal: true, shell: opts.shell };
```
f-f, line 4: Here we always set the `preferLocal: true` option, we should port that to the new code as well

f-f (approved):
This is looking great now! 👏 
I left a few minor comments, but it's otherwise good to go.

spaghetto/src/Spago/Cmd.js
JordanMartinez: `preferLocal: true` [is already the default](https://github.com/JordanMartinez/purescript-node-execa/blob/master/src/Node/Library/Execa.purs#L182)

JordanMartinez: See https://github.com/JordanMartinez/purescript-node-execa/blob/master/src/Node/Library/Execa.purs#L89-L93 for why I use `Just { Nothing, Nothing }`. When unspecified, `cwd` and `execPath` default to `cwd` and the current process' `execPath`.

spaghetto/src/Spago/Cmd.js
JordanMartinez: Where should this be documented?

spaghetto/src/Spago/Cmd.js
f-f: Ah, great then!

spaghetto/src/Spago/Cmd.js
f-f: In `execa`, no need to document it here

f-f (approved):
This is great, thank you!

PR #947, opened 2023-02-10
spago@next: ls command

andys8:
### Description of the change

The draft is supposed to replicate existing behavior of spago in spago@next. In particular `ls deps` and `ls packages` with the flags `-t` and `-j`.

The behavior is not there yet. This is a draft to get help for some of the open questions, marked as comments.

### DIFF
--- a/spaghetto/bin/src/Flags.purs
+++ b/spaghetto/bin/src/Flags.purs
@@ -75,12 +75,17 @@ json =
     # ArgParser.boolean
     # ArgParser.default false
 
-transitive ∷ ArgParser Boolean
+-- TODO: Discuss if we want to use sum types for flags or go with boolean
+-- Depending on the decision, revert or change the json flag, too
+data IncludeTransitive = IncludeTransitive | NoIncludeTransitive
+
+transitive ∷ ArgParser IncludeTransitive
 transitive =
   ArgParser.flag [ "--transitive", "-t" ]
     "Include transitive dependencies"
     # ArgParser.boolean
     # ArgParser.default false
+    # map (if _ then IncludeTransitive else NoIncludeTransitive)
 
 pedanticPackages ∷ ArgParser Boolean
 pedanticPackages =
--- a/spaghetto/bin/src/Flags.purs
+++ b/spaghetto/bin/src/Flags.purs
@@ -75,17 +75,12 @@ json =
     # ArgParser.boolean
     # ArgParser.default false
 
--- TODO: Discuss if we want to use sum types for flags or go with boolean
--- Depending on the decision, revert or change the json flag, too
-data IncludeTransitive = IncludeTransitive | NoIncludeTransitive
-
-transitive ∷ ArgParser IncludeTransitive
+transitive ∷ ArgParser Boolean
 transitive =
   ArgParser.flag [ "--transitive", "-t" ]
     "Include transitive dependencies"
     # ArgParser.boolean
     # ArgParser.default false
-    # map (if _ then IncludeTransitive else NoIncludeTransitive)
 
 pedanticPackages ∷ ArgParser Boolean
 pedanticPackages =
--- a/spaghetto/bin/src/Flags.purs
+++ b/spaghetto/bin/src/Flags.purs
@@ -70,8 +70,15 @@ noColor =
 
 json ∷ ArgParser Boolean
 json =
-  ArgParser.flag [ "--json" ]
+  ArgParser.flag [ "--json", "-j" ]
     "Format the output as JSON"
+    # ArgParser.boolean
+    # ArgParser.default false
+
+transitive ∷ ArgParser Boolean
+transitive =
+  ArgParser.flag [ "--transitive", "-t" ]
+    "Include transitive dependencies"
     # ArgParser.boolean
     # ArgParser.default false
 
--- a/spaghetto/bin/src/Flags.purs
+++ b/spaghetto/bin/src/Flags.purs
@@ -70,14 +70,14 @@ noColor =
 
 json ∷ ArgParser Boolean
 json =
-  ArgParser.flag [ "--json", "-j" ]
+  ArgParser.flag [ "--json" ]
     "Format the output as JSON"
     # ArgParser.boolean
     # ArgParser.default false
 
 transitive ∷ ArgParser Boolean
 transitive =
-  ArgParser.flag [ "--transitive", "-t" ]
+  ArgParser.flag [ "--transitive" ]
     "Include transitive dependencies"
     # ArgParser.boolean
     # ArgParser.default false
--- a/spaghetto/bin/src/Main.purs
+++ b/spaghetto/bin/src/Main.purs
@@ -28,7 +28,7 @@ import Spago.Command.Build as Build
 import Spago.Command.Bundle as Bundle
 import Spago.Command.Fetch as Fetch
 import Spago.Command.Init as Init
-import Spago.Command.Ls (IncludeTransitive(..))
+import Spago.Command.Ls (IncludeTransitive(..), JsonFlag(..))
 import Spago.Command.Ls as Ls
 import Spago.Command.Registry as Registry
 import Spago.Command.Run as Run
@@ -426,12 +426,13 @@ main =
             runSpago buildEnv (Build.run options)
             testEnv <- runSpago env (mkTestEnv args)
             runSpago testEnv Test.run
-          Ls _args -> do
+          Ls args -> do
             let fetchArgs = { packages: mempty, selectedPackage: Nothing, ensureRanges: false }
             { env, fetchOpts } <- mkFetchEnv fetchArgs
             -- TODO: --no-fetch flag
             dependencies <- runSpago env (Fetch.run fetchOpts)
-            runSpago { dependencies } (Ls.listPackages IncludeTransitive "")
+            lsEnv <- runSpago env (mkLsEnv args dependencies)
+            runSpago lsEnv (Ls.listPackageSet JsonOutputNo)
 
 mkLogOptions :: GlobalArgs -> Aff LogOptions
 mkLogOptions { noColor, quiet, verbose } = do
@@ -692,6 +693,11 @@ mkRegistryEnv = do
     , logOptions
     , git
     }
+
+mkLsEnv :: forall a b. LsArgs -> Map PackageName Package -> Spago (Fetch.FetchEnv a) Ls.LsEnv
+mkLsEnv lsArgs dependencies = do
+  { logOptions, workspace, git } <- ask
+  pure { logOptions, dependencies }
 
 shouldFetchRegistryRepos :: forall a. Spago (LogEnv a) Boolean
 shouldFetchRegistryRepos = do
--- a/spaghetto/bin/src/Main.purs
+++ b/spaghetto/bin/src/Main.purs
@@ -21,16 +21,14 @@ import Registry.Constants as Registry.Constants
 import Registry.ManifestIndex as ManifestIndex
 import Registry.Metadata as Metadata
 import Registry.PackageName as PackageName
-import Spago.Bin.Flags (IncludeTransitive)
-import Spago.Bin.Flags as Flag
 import Spago.Bin.Flags as Flags
 import Spago.BuildInfo as BuildInfo
 import Spago.Command.Build as Build
 import Spago.Command.Bundle as Bundle
 import Spago.Command.Fetch as Fetch
 import Spago.Command.Init as Init
-import Spago.Command.Ls (JsonFlag(..))
 import Spago.Command.Ls as Ls
+import Spago.Command.Ls (LsDepsArgs, LsPackagesArgs)
 import Spago.Command.Registry as Registry
 import Spago.Command.Run as Run
 import Spago.Command.Sources as Sources
@@ -134,15 +132,6 @@ type BundleArgs =
   , pedanticPackages :: Boolean
   , type :: Maybe String
   , ensureRanges :: Boolean
-  }
-
-type LsPackagesArgs =
-  { json :: Boolean
-  }
-
-type LsDepsArgs =
-  { json :: Boolean
-  , transitive :: IncludeTransitive
   }
 
 data SpagoCmd a = SpagoCmd GlobalArgs (Command a)
@@ -272,8 +261,8 @@ installArgsParser =
     , pursArgs: Flags.pursArgs
     , backendArgs: Flags.backendArgs
     , output: Flags.output
-    , pedanticPackages: Flag.pedanticPackages
-    , ensureRanges: Flag.ensureRanges
+    , pedanticPackages: Flags.pedanticPackages
+    , ensureRanges: Flags.ensureRanges
     }
 
 buildArgsParser :: ArgParser (BuildArgs ())
@@ -328,7 +317,7 @@ bundleArgsParser =
     , backendArgs: Flags.backendArgs
     , output: Flags.output
     , pedanticPackages: Flags.pedanticPackages
-    , ensureRanges: Flag.ensureRanges
+    , ensureRanges: Flags.ensureRanges
     }
 
 registrySearchArgsParser :: ArgParser RegistrySearchArgs
@@ -450,16 +439,14 @@ main =
             -- TODO: --no-fetch flag
             dependencies <- runSpago env (Fetch.run fetchOpts)
             lsEnv <- runSpago env (mkLsEnv dependencies)
-            let j = if args.json then JsonOutputYes else JsonOutputNo
-            runSpago lsEnv (Ls.listPackageSet j)
-          LsDeps args@{ transitive } -> do
+            runSpago lsEnv (Ls.listPackageSet args)
+          LsDeps args -> do
             let fetchArgs = { packages: mempty, selectedPackage: Nothing, ensureRanges: false }
             { env, fetchOpts } <- mkFetchEnv fetchArgs
             -- TODO: --no-fetch flag
             dependencies <- runSpago env (Fetch.run fetchOpts)
             lsEnv <- runSpago env (mkLsEnv dependencies)
-            let j = if args.json then JsonOutputYes else JsonOutputNo
-            runSpago lsEnv (Ls.listPackages transitive j)
+            runSpago lsEnv (Ls.listPackages args)
 
 mkLogOptions :: GlobalArgs -> Aff LogOptions
 mkLogOptions { noColor, quiet, verbose } = do
--- a/spaghetto/bin/src/Main.purs
+++ b/spaghetto/bin/src/Main.purs
@@ -135,9 +135,12 @@ type BundleArgs =
   , ensureRanges :: Boolean
   }
 
-type LsArgs =
+type LsPackagesArgs =
   { json :: Boolean
-  -- TODO: Transitive only makes sense for deps, bot for packages
+  }
+
+type LsDepsArgs =
+  { json :: Boolean
   , transitive :: Boolean
   }
 
@@ -155,7 +158,8 @@ data Command a
   | Sources SourcesArgs
   | RegistrySearch RegistrySearchArgs
   | RegistryInfo RegistryInfoArgs
-  | Ls LsArgs
+  | LsDeps LsDepsArgs
+  | LsPackages LsPackagesArgs
 
 argParser :: ArgParser (SpagoCmd ())
 argParser =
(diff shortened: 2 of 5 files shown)


spaghetto/bin/src/Flags.purs
```diff
@@ -70,11 +70,23 @@ noColor =
 
 json ∷ ArgParser Boolean
 json =
-  ArgParser.flag [ "--json" ]
+  ArgParser.flag [ "--json", "-j" ]
     "Format the output as JSON"
     # ArgParser.boolean
     # ArgParser.default false
 
+-- TODO: Discuss if we want to use sum types for flags or go with boolean
```
andys8, line 78: :question: Sum type or boolean parser flags?

spaghetto/bin/src/Flags.purs
```diff
@@ -70,11 +70,23 @@ noColor =
...
+-- Depending on the decision, revert or change the json flag, too
+data IncludeTransitive = IncludeTransitive | NoIncludeTransitive
+
+transitive ∷ ArgParser IncludeTransitive
+transitive =
+  ArgParser.flag [ "--transitive", "-t" ]
+    "Include transitive dependencies"
+    # ArgParser.boolean
+    # ArgParser.default false
+    # map (if _ then IncludeTransitive else NoIncludeTransitive)
```
andys8, line 88: :question: Sum type or boolean flags in the parser?

spaghetto/bin/src/Main.purs
```diff
@@ -411,6 +444,22 @@ main =
             runSpago buildEnv (Build.run options)
             testEnv <- runSpago env (mkTestEnv args)
             runSpago testEnv Test.run
+          LsPackages args -> do
+            let fetchArgs = { packages: mempty, selectedPackage: Nothing, ensureRanges: false }
+            { env, fetchOpts } <- mkFetchEnv fetchArgs
+            -- TODO: --no-fetch flag
+            dependencies <- runSpago env (Fetch.run fetchOpts)
```
andys8, line 451: :question: Currently downloads packages. There was a comment regarding `no-fetch`, but maybe in this case (`ls packages`), this is the wrong way to gather `dependencies` to begin with? There is also a different comment, that all packages seem to be "local", and these could be related.

spaghetto/src/Spago/Command/Ls.purs
```diff
@@ -1,3 +1,120 @@
...
+  JsonOutputNo -> OutputLines <<< formatPackageNamesText
+  where
+  formatPackageNamesJson :: Array (PackageName /\ Package) -> Array String
+  formatPackageNamesJson pkgs = (stringifyJson jsonPackageOutputCodec <<< asJson <$> pkgs)
+    where
+    asJson (name /\ package) =
+      { packageName: PackageName.print name
+      -- TODO: Do we want to stay backward compatible and have a tagged remote and local type?
+      -- Or do we create a codec for package (with 4 variants)?
+      , repo: "TODO"
```
andys8, line 80: :question: In general I'm asking myself, if we should:
* map to `Remote | Local` like it was before (https://github.com/andys8/spago/blob/44793b36ffdb38d1bf43b4ca77a7d4ab513e9632/src/Spago/Types.hs#L27-L35)
* map to `Package`, and introduce a json coded (https://github.com/andys8/spago/blob/44793b36ffdb38d1bf43b4ca77a7d4ab513e9632/spaghetto/src/Spago/Config.purs#L293-L297)

spaghetto/src/Spago/Command/Ls.purs
```diff
@@ -1,3 +1,120 @@
...
+
+  -- | Format all the package names from the configuration
+  formatPackageNamesText :: Array (PackageName /\ Package) -> Array String
+  formatPackageNamesText pkgs =
+    let
+
+      -- TODO: Currently prints git/remote packages as local packages
+      showLocation :: PackageName -> Package -> String
+      showLocation _ (GitPackage gitPackage) = "Remote " <> surroundQuote gitPackage.git
+      showLocation packageName package = "Local " <> surroundQuote (getPackageLocation packageName package)
```
andys8, line 92: :bug: Here the behavior is definitely off right now. 

> Local ".spago/packages/strings-6.0.1

Printing `Local` but what I want is the git package. See other comment.

spaghetto/bin/src/Main.purs
f-f: The old spago could put together a build plan without downloading things, because all the dependencies had to be specified in the package set.
This is not the case anymore in the new spago, as we do not force users to type down all the dependencies beforehand, but we allow for them to be read from the config files contained in the remote repos.
This means that in the general case there's no way to put together a build plan before downloading all the dependencies, as some of them might be remote git repos.
It is possible in some specific cases, e.g. when all the dependencies of your project come from the registry. In that case we can look at the registry index for dependency information.

What this means for the above code is that I think we are good to proceed as is: it makes sense that the only way to get a build plan is to call `Fetch.run`, and we'll eventually introduce a `--no-fetch` or `--offline` flag to be handled by the fetching code (so this TODO can stay in place too)

spaghetto/src/Spago/Command/Ls.purs
```diff
@@ -1,3 +1,120 @@
...
+  , version: CA.string
+  }
+
+listPackageSet :: JsonFlag -> Spago LsEnv Unit
+listPackageSet jsonFlag = do
+  logDebug "Running `listPackageSet`"
+  { workspace } <- ask
+  output $ formatPackageNames jsonFlag (Map.toUnfoldable workspace.packageSet)
+
+listPackages :: IncludeTransitive -> JsonFlag -> Spago LsEnv Unit
```
f-f, line 50: Let's pass in the json flag as a boolean too. In general we pass a record of options to commands, so you can squash these parameters in a single record.

spaghetto/src/Spago/Command/Ls.purs
f-f: Keeping backwards compatibility is not a priority here (since there are so many breaking changes anyways), and since the new types are so different from the old ones I would say it's better to come up with something else rather than try to fit in the old format.

So let's discard the previous `Local`/`Remote`, and make a codec for `Package`

spaghetto/src/Spago/Command/Ls.purs
```diff
@@ -1,3 +1,120 @@
...
+          directDependencies = unions $ ((toDependencyNames <<< _.dependencies <<< _.package) <$> workspacePackages)
+        in
+          filterKeys (_ `elem` directDependencies) dependencies
+  case packagesToList of
+    [] -> logWarn "There are no dependencies listed in your spago.dhall"
+    _ -> output $ formatPackageNames jsonFlag packagesToList
+
+formatPackageNames :: forall a. JsonFlag -> Array (PackageName /\ Package) -> OutputFormat a
+formatPackageNames = case _ of
+  JsonOutputYes -> OutputLines <<< formatPackageNamesJson
```
f-f, line 70: We have `OutputJson` for these cases (the old format used json-lines, but we should just use standard JSON everywhere)

spaghetto/src/Spago/Command/Ls.purs
```diff
@@ -1,3 +1,120 @@
...
+  logDebug "Running `listPackages`"
+  { dependencies, workspace } <- ask
+  let
+    packagesToList = Map.toUnfoldable case packagesFilter of
+      IncludeTransitive -> dependencies
+      NoIncludeTransitive ->
+        let
+          workspacePackages = Config.getWorkspacePackages workspace.packageSet
+          toDependencyNames (Dependencies deps) = Map.keys deps
+          directDependencies = unions $ ((toDependencyNames <<< _.dependencies <<< _.package) <$> workspacePackages)
```
f-f, line 61: You'll want to check if any of the workspace packages was selected, and only collect the dependencies from all of them if none was selected. E.g. see here:
https://github.com/purescript/spago/blob/243099f29ff5aec77ef236ee07f5213667cc0719/spaghetto/src/Spago/Command/Build.purs#L53-L56

spaghetto/src/Spago/Command/Ls.purs
f-f: Yeah, this is because `getPackageLocation` is not we need for this, since it's for getting the path of the downloaded package on disk.

I guess the textual representation needs a little more thought than the JSON version (since for that one it's enough to add a tag to denote the kind of package, and then just dump the representation of each package in JSON).

The current format includes package name, version and location.
We could keep the same format and have the different kind of packages map to that sensibly:
| Package type  | Version | Location |
| ------------- | ------- | -------- |
| registry      | version | ""       |
| git           | ref     | repo     |
| local         | "local" | path     |
| workspace     | "workspace" | path |

andys8:
Thanks for the very detailed, good review :)

spaghetto/bin/src/Flags.purs
andys8: Okay, 50d81ddbfb53a39e2f3914c98ea6c480adce2578

spaghetto/bin/src/Flags.purs
```diff
@@ -70,11 +70,23 @@ noColor =
 
 json ∷ ArgParser Boolean
 json =
-  ArgParser.flag [ "--json" ]
+  ArgParser.flag [ "--json", "-j" ]
```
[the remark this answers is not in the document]
andys8: Sure :) 977282400483a152c932f4c85c60bd0e5a254c4d

spaghetto/bin/src/Main.purs
andys8: Question answered

spaghetto/src/Spago/Command/Ls.purs
andys8: Done in 50d81ddbfb53a39e2f3914c98ea6c480adce2578

spaghetto/src/Spago/Command/Ls.purs
andys8: Tried the table (based on dodo-printer) in 415149d6958a929a22a95ea7b85a1d1545e4b04a

spaghetto/src/Spago/Command/Ls.purs
andys8: Okay, if it doesn't need to be the same I can use `OutputJson` and `OutputTable`

spaghetto/src/Spago/Command/Ls.purs
andys8: Went with the coded: https://github.com/andys8/spago/blob/415149d6958a929a22a95ea7b85a1d1545e4b04a/spaghetto/src/Spago/Config.purs#L318-L332

Only part missing / not sure if necessary / how to handle, is `YamlDoc Config` in `WorkspacePackage`. I feel like printing this in dependencies isn't what we would want.

https://github.com/andys8/spago/blob/415149d6958a929a22a95ea7b85a1d1545e4b04a/spaghetto/src/Spago/Config.purs#L303-L310

spaghetto/src/Spago/Command/Ls.purs
andys8: Like this 312de0db42f05fae647304188d9027523f45cd23?

spaghetto/src/Spago/Config.purs
```diff
@@ -290,12 +292,43 @@ type WorkspacePackage =
...
+  }
+
+yamlDocCodec :: JsonCodec (YamlDoc Config)
+yamlDocCodec = CA.codec' decode encode
+  where
+  -- TODO: implementation of encode
+  encode _x = CA.encode CA.null unit
+
+  -- TODO: implementation of decode if needed
+  decode _json = Left MissingValue
```
andys8, line 310: :information_source: As I wrote somewhere else, my assumption is that encoding this - at least in the dependencies - would make sense. Open to other suggestions here (noop, different type).

spaghetto/src/Spago/Log.purs
```diff
@@ -123,14 +129,76 @@ data OutputFormat a
...
+  OutputTable { titles, rows } ->
+    Log.print Log.plainText (Log.twoSpaces)
+      $ Box.toDoc
+      $ logTable { headers: textBox <$> titles, rows: (map textBox) <$> rows }
+
+textBox :: forall a. String -> DocBox a
+textBox = print Box.docBox twoSpaces <<< Dodo.text
+
+-- TODO: Change visual appearance
+logTable
```
andys8, line 151: Table implementation was missing. I took the implementation mostly from the comment. There is still room for improvement, but looks okay. Don't want to blow up the PR with more  stuff, but needed some output, to be able to use `OutputTable`.
https://github.com/natefaubion/purescript-dodo-printer/blob/master/test/snapshots/DodoBox.purs

andys8: ## Example Output

image

spaghetto/src/Spago/Command/Ls.purs
andys8: See https://github.com/purescript/spago/pull/947#discussion_r1111105804

spaghetto/src/Spago/Command/Ls.purs
andys8: ## Example JSON Output

```json
{
  "aff": "7.1.0",
  "aff-promise": "4.0.0",
  "affjax": "13.0.0",
  "affjax-node": "1.0.0",
  "argonaut-core": "7.0.0",
  "argparse-basic": "2.0.0",
  "arrays": "7.2.0",
  "bifunctors": "6.0.0",
  "codec-argonaut": "10.0.0",
  "console": "6.0.0",
  "control": "6.0.0",
  "datetime": "6.1.0",
  "dodo-printer": "2.2.1",
  "effect": "4.0.0",
  "either": "6.1.0",
  "exceptions": "6.0.0",
  "filterable": "5.0.0",
  "foldable-traversable": "6.0.0",
  "functions": "6.0.0",
  "http-methods": "6.0.0",
  "identity": "6.0.0",
  "integers": "6.0.0",
  "lists": "7.0.0",
  "maybe": "6.0.0",
  "newtype": "5.0.0",
  "node-buffer": "8.0.0",
  "node-execa": "1.1.0",
  "node-fs": "8.1.1",
  "node-fs-aff": "9.1.0",
  "node-path": "5.0.0",
  "node-process": "10.0.0",
  "now": "6.0.0",
  "nullable": "6.0.0",
  "ordered-collections": "3.0.0",
  "parallel": "6.0.0",
  "partial": "4.0.0",
  "prelude": "6.0.1",
  "profunctor": "6.0.0",
  "record": "4.0.0",
  "refs": "6.0.0",
  "registry-foreign": {
    "git": "https://github.com/purescript/registry-dev.git",
    "ref": "8113deecae25410ba62c31a5842c6ebbf1c12461",
    "subdir": "foreign"
  },
  "registry-lib": {
    "git": "https://github.com/purescript/registry-dev.git",
    "ref": "8113deecae25410ba62c31a5842c6ebbf1c12461",
    "subdir": "lib"
  },
  "spago": {
    "doc": null,
    "hasTests": true,
    "package": {
      "dependencies": [
        "aff",
        "aff-promise",
        "affjax",
        "affjax-node",
        "argonaut-core",
        "arrays",
        "bifunctors",
        "codec-argonaut",
        "console",
        "control",
        "datetime",
        "dodo-printer",
        "effect",
        "either",
        "exceptions",
        "filterable",
        "foldable-traversable",
        "functions",
        "http-methods",
        "identity",
        "integers",
        "lists",
        "maybe",
        "newtype",
        "node-buffer",
        "node-execa",
        "node-fs",
        "node-fs-aff",
        "node-path",
        "node-process",
        "now",
        "nullable",
        "ordered-collections",
        "parallel",
        "partial",
        "prelude",
        "profunctor",
        "record",
        "refs",
        "registry-foreign",
        "registry-lib",
        "strings",
        "transformers",
        "tuples",
        "unsafe-coerce"
      ],
      "name": "spago"
    },
    "path": "./"
  },
  "strings": "6.0.1",
  "transformers": "6.0.0",
  "tuples": "7.0.0",
  "unsafe-coerce": "6.0.0"
}
```

spaghetto/src/Spago/Config.purs
```diff
@@ -290,12 +292,43 @@ type WorkspacePackage =
...
   | LocalPackage LocalPackage
   | WorkspacePackage WorkspacePackage
 
+packageCodec :: JsonCodec Package
+packageCodec = CA.codec' decode encode
+  where
+  encode (RegistryVersion x) = CA.encode Version.codec x
+  encode (GitPackage x) = CA.encode gitPackageCodec x
+  encode (LocalPackage x) = CA.encode localPackageCodec x
+  encode (WorkspacePackage x) = CA.encode workspacePackageCodec x
```
f-f, line 324: A general note first: the main advantage of using codecs is that we can have different JSON encodings for the same type - we are defining this one (and other related ones) specifically for the `ls` command, so we should have all the new codecs in that module, so that we are free to have different implementations here (if we ever want a generic codec for the config)

Specifically about this codec: in my previous review I suggested that we "just dump the json using the existing codecs", but after trying it out I feel that we should be a little more structured: if you're parsing this data from any language (or even just `jq`), you'll need a way to discern the type of the inner object.
This is commonly done by wrapping the cases of the sum into records with `type` and `value` keys.
So in our case we'd have (I'm putting here just the encode as an example):
```suggestion
packageCodec = CAR.object "Package" { type: CA.string, value: innerCodec }
  where
  innerCodec = CA.codec' decode encode
  
  registryVersionCodec = CAR.object "RegistryVersion" { version: Version.codec }
  
  encode (RegistryVersion x) = CA.encode registryVersionCodec x
  encode (GitPackage x) = CA.encode gitPackageCodec x
  encode (LocalPackage x) = CA.encode localPackageCodec x
  encode (WorkspacePackage x) = CA.encode workspacePackageCodec x
```

Note that I wrapped the `RegistryVersion` one so that all the `value`s are records (for consistency). 
I would also suggest inlining `workspacePackageCodec` and other codecs that we need inside this `where` clause, since they will only be used here.

spaghetto/src/Spago/Log.purs
f-f: Yes! Beautiful! I meant to implement this with Dodo, but got busy with other things, so I just left the link in there. Very happy to see this implemented! 👏 
Now we can use this to fix the output for the `spago registry` commands, but not in this PR 😄

spaghetto/src/Spago/Command/Ls.purs
```diff
@@ -1,3 +1,83 @@
...
+    [] -> logWarn "There are no dependencies listed in your spago.dhall"
+    _ -> output $ formatPackageNames json packages
+
+formatPackageNames :: Boolean -> Array (PackageName /\ Package) -> OutputFormat (Map PackageName Package)
+formatPackageNames json pkgs =
+  if json then OutputJson (packageMap packageCodec) (Map.fromFoldable pkgs)
+  else formatPackagesTable
+  where
+  formatPackagesTable = OutputTable
+    { titles: [ "Package Type", "Version", "Location" ]
```
f-f, line 62: Typo:
```suggestion
    { titles: [ "Package", "Version", "Location" ]
```

spaghetto/src/Spago/Command/Ls.purs
```diff
@@ -1,3 +1,83 @@
...
+  { dependencies, workspace } <- ask
+  let
+    directDependencies = filterKeys (_ `elem` direct) dependencies
+      where
+      toDependencyNames (Dependencies deps) = Map.keys deps
+      packagesToNames = toDependencyNames <<< _.dependencies <<< _.package
+      direct = unions (packagesToNames <$> workspacePackages)
+      workspacePackages = case workspace.selected of
+        Just p -> [ p ]
+        Nothing -> Config.getWorkspacePackages workspace.packageSet
```
f-f, line 49: Then this gets easier once we have only one package (it can be inlined more, but this is just to give an idea):
```suggestion
  { dependencies, workspace, selected } <- ask
  let
    directDependencies = filterKeys (_ `elem` direct) dependencies
      where
      toDependencyNames (Dependencies deps) = Map.keys deps
      packagesToNames = toDependencyNames <<< _.dependencies <<< _.package
      direct = packagesToNames selected
```

spaghetto/src/Spago/Log.purs
```diff
@@ -123,14 +129,76 @@ data OutputFormat a
...
-indent2 = Log.indent <<< Log.indent
+  OutputTable { titles, rows } ->
+    Log.print Log.plainText (Log.twoSpaces)
+      $ Box.toDoc
+      $ logTable { headers: textBox <$> titles, rows: (map textBox) <$> rows }
+
+textBox :: forall a. String -> DocBox a
+textBox = print Box.docBox twoSpaces <<< Dodo.text
+
+-- TODO: Change visual appearance
```
f-f, line 150: I am fine with this visual appearance, we can remove this comment 🙂

f-f (commented):
This is great! A few minor things I noticed after trying it out, but almost there

PR #955, opened 2023-03-30
Integrate PSA into spago

### LINKED ISSUE #904: Integrate psa into Spago

JordanMartinez:
### Description of the change

Fixes #904.

- [ ] for later PRs, determine what additional features to support and which get `spago.yml` support (see https://github.com/natefaubion/purescript-psa-utils/pulls and https://github.com/natefaubion/purescript-psa/issues)

Summary of PR:
1. `psa` and `psa-utils` were vendored into this codebase and modified/updated in the following ways:
    - using `Dodo` as its printer rather than its custom code. Moreover, a brighter red and yellow are used for warnings and errors than what was used previously.
    - using `codec-argonaut` as its Json encoder/decoder
    - removing unused/unexported declarations
    - replacing some declarations with pre-existing ones (e.g. `replicate` with `power`
    - cleaning up the imports
    - renaming `PsaOptions` to `PsaOutputOptions` because `*Options` in this codebase typically refers to values used in misc commands. 
    - renaming `output` to `buildOutput` because `output` is already exported from `Spago.Prelude`
    - combining `stash` and `stashFile` into one argument
    - renaming the `ansi` arg to `color` (since Spago already has a similarly-named flag)
    - dropping the `isLib` arg as libraries can be determined via Spago
    - changing the Json printer to print to `stdout` rather than `stderr`. (I'm not sure if this is correct to do, but I'll clarify why in the PR via a code comment).
2. `psa` was integrated into Spago in the following way:
    - logic in `psa`'s `main` function was extracted out into `Psa.psaCompile`, which wraps `Purs.compile`
    - various `logDebug` statements were added to make it easier to see where problems arise (if any) in the future
    - Of the args that `psa` takes, I decided to keep all of them configurable via `spago.yml` except for:
        - `json-errors`: I assume this is only useful if being called by an IDE or if someone wants to run their own custom printer
        - `ansi`: this doesn't seem like something one should configure via `spago.yml`, only via CLI
        - `libraryDirs`: with Spago in control, we know what the library directories are.
    - Stashes are stored in one of three places:
        - when `spago build -b pkgName` is used: `.spago/stashes/defaults/pkgName.stash`
        - when `spago build` is used: `.spago/stashes/defaults/entire-workspace.stash`
        - when `spago build --psa-stash-file "foo"` is used: `.spago/stashes/custom/foo.stash`
        - when `spago build --psa-stash-file "../foo"` is used: `.spago/stashes/custom/foo.stash`

### Checklist:


JordanMartinez:
CI will fail as the current code does not compile. If you can help me figure out what I need to fix to make this compile, then I can see whether this code works.

f-f:
Thanks for the update! I'll check out the branch in the morning and see how I can help.
(btw CI doesn't know anything about spaghetti yet, no purescript is being built yet, and there are no tests for it anyways. It's on my to-do list to split off the Haskell implementation so we can have a simpler CI setup. Maybe tomorrow is a good day for that)

spaghetto/bin/src/Main.purs
```diff
@@ -403,56 +454,68 @@ main =
...
             env <- mkRegistryEnv
             void $ runSpago env (Registry.info args)
-          Install args@{ packages, selectedPackage, ensureRanges } -> do
+          Install args@{ packages, selectedPackage, ensureRanges, jsonErrors } -> do
             { env, fetchOpts } <- mkFetchEnv { packages, selectedPackage, ensureRanges }
             -- TODO: --no-fetch flag
             dependencies <- runSpago env (Fetch.run fetchOpts)
             env' <- runSpago env (mkBuildEnv args dependencies)
-            let options = { depsOnly: true, pursArgs: List.toUnfoldable args.pursArgs }
+            let options = { depsOnly: true, pursArgs: List.toUnfoldable args.pursArgs, jsonErrors }
```
JordanMartinez, line 462: Should we drop `json-errors` from `InstallArgs` and just set this to `false`? Likewise for Bundle, Run, and Test?

spaghetto/bin/src/Flags.purs
```diff
@@ -4,13 +4,87 @@ import Spago.Prelude
...
+    "Censor warnings from project sources"
+    # ArgParser.boolean
+    # ArgParser.optional
+
+psaShowSource ∷ ArgParser (Maybe ShowSourceCode)
+psaShowSource =
+  NoSourceCode
+    <$ ArgParser.flag [ "--psa-no-source" ]
+      "Disable original source code printing"
+    # ArgParser.optional
```
JordanMartinez, line 53: I used an ADT here rather than a Boolean as it seemed a bit easier to read because the existence of this flag disables the 'show source code' feature.

spaghetto/spago.yaml
```diff
@@ -45,11 +45,12 @@ package:
     - transformers
     - tuples
     - unsafe-coerce
+    - versions
```
JordanMartinez, line 48: I think this can be removed. Originally, I thought I would need it because `psa` would output things to `stderr` or `stdout` depending on the `purs` version. Since spago only supports `0.15.4` or higher, that question is irrelevant now.

spaghetto/src/Spago/Psa.purs
```diff
@@ -0,0 +1,205 @@
...
+type PsaArgs =
+  { libraryDirs :: Array String
+  , jsonErrors :: Boolean
+  , color :: Boolean
+  }
+
+defaultParseOptions :: PsaOptions
+defaultParseOptions =
+  { showSource: Core.ShowSourceCode
+  , stashFile: Nothing -- ".psa-stash"
```
JordanMartinez, line 45: This comment can be removed.

spaghetto/src/Spago/Psa.purs
```diff
@@ -0,0 +1,205 @@
...
+          loadLinesImpl = if showSource == Core.ShowSourceCode then loadLines files else loadNothing
+          filenames = insertFilenames (insertFilenames Set.empty out.errors) out.warnings
+        merged <- mergeWarnings filenames stashData.date stashData.stash out.warnings
+        for_ stashFile \f -> writeStashFile f merged
+
+        out' <- buildOutput loadLinesImpl outputOptions out { warnings = merged }
+
+        liftEffect $ if psaArgs.jsonErrors then JsonPrinter.print out' else DefaultPrinter.print outputOptions out'
+
+        pure $ FO.isEmpty out'.stats.allErrors
```
JordanMartinez, line 114: The logic here might be wrong regarding `jsonErrors`. The current code will output the Psa-modified errors rather than what is directly received from `purs`. Should it? Otherwise, we could rewrite this to:
```purs
if psaArgs.jsonErrors then
   Console.log pursJsonErrors
   die ...
else ...
  DefaultPrinter.print ...
```
and drop the JsonPrinter completely.

spaghetto/src/Spago/Psa.purs
```diff
@@ -0,0 +1,205 @@
...
+  isEmptySpan filename pos =
+    filename == ""
+      || pos.startLine
+      == 0
+      && pos.endLine
+      == 0
+      && pos.startColumn
+      == 0
+      && pos.endColumn
+      == 0
```
JordanMartinez, line 135: This could be better formatted.

spaghetto/src/Spago/Psa/Printer/Default.purs
```diff
@@ -0,0 +1,275 @@
...
+  renderStats' = case options.statVerbosity of
+    Core.NoStats -> mempty
+    Core.CompactStats -> renderStats
+    Core.VerboseStats -> renderVerboseStats
+
+renderWarning :: Int -> Int -> PsaAnnotedError -> D.Doc Ansi.GraphicsParam
+renderWarning = renderWrapper Ansi.BrightYellow
+
+renderError :: Int -> Int -> PsaAnnotedError -> D.Doc Ansi.GraphicsParam
+renderError = renderWrapper Ansi.BrightRed
```
JordanMartinez, line 64: Should we add a dark/light theme option that changes whether normal Red/Yellow or bright Red/Yellow are used? This works for my terminal colors but may not for others.

spaghetto/src/Spago/Psa/Types.purs
```diff
@@ -0,0 +1,141 @@
...
+import Data.Codec.Argonaut.Compat as CACompat
+import Data.Maybe (Maybe(..))
+import Data.Set (Set)
+import Data.Tuple (Tuple(..))
+import Spago.Core.Config as Core
+
+type ErrorCode = String
+type ModuleName = String
+type Filename = String
+type Lines = Array String
```
JordanMartinez, line 37: I'm wondering if these type aliases should be dropped.

spaghetto/src/Spago/Purs.purs
```diff
@@ -42,16 +42,13 @@ getPurs =
...
   { purs } <- ask
   let args = [ "compile" ] <> pursArgs <> Set.toUnfoldable globs
   logDebug [ "Running command:", "purs " <> String.joinWith " " args ]
-  Cmd.exec purs.cmd args Cmd.defaultExecOptions >>= case _ of
-    Right _r -> logSuccess "Build succeeded."
-    Left err -> do
-      logDebug $ show err
-      die [ "Failed to build." ]
+  Cmd.exec purs.cmd args $ Cmd.defaultExecOptions
+    { pipeStdout = false }
```
JordanMartinez, line 51: `pipeStdout = false` is needed to prevent the JSON errors from `purs` being outputted to the console.

spaghetto/README.md
```diff
@@ -46,6 +46,48 @@ package:
     #   2. the package, using a specified vesrion range
     - packageName: ">=1.1.1 <2.0.0"
 
+# optional
+  build:
+    # optional
+    psaOptions:
```
f-f, line 52: I would not nest this under `psaOptions`, they can all be "build options"

spaghetto/README.md
```diff
@@ -46,6 +46,48 @@ package:
...
+      # optional, Boolean, censor all warnings
+      censorWarnings: false
+      # optional, Boolean, censor warnings from library sources
+      censorLib: false
```
f-f, line 56: I feel like `censorLibWarnings` is a more consistent name?

spaghetto/README.md
```diff
@@ -46,6 +46,48 @@ package:
...
+      # optional, Boolean, censor warnings from project sources
+      censorSrc: false
```
f-f, line 58: And this as `censorProjectWarnings`

spaghetto/README.md
```diff
@@ -46,6 +46,48 @@ package:
...
+      # the pretty-printed compiler warnings/errors
+      # and whether to store them in a non-default location
+      stashFile:
+        # one of 3 values
+        #    1. Disable it
+        false
+        #    2. Enable it using default location (.spago/default/)
+        true
+        #    3. Enable it using custom location (.spago/custom)
+
```
f-f, line 90: Is there a value missing here?

spaghetto/README.md
```diff
@@ -148,4 +190,9 @@ workspace:
     output: "output"
     # optional, Boolean, fail the build if `spago.yml` has redundant/missing packages
     pedantic_packages: false
+    # optional, PsaOptions
+    # These options only take affect when
+    # a package is not selected (e.g. `spago build` vs `spago build -p pkgName`)
```
f-f, line 195: I'm feeling weird about having a `build` config in the `package` section - we already have one here in the `workspace` section, and it contains some options already.

I think we should not have it in the `package` section at all, since a "build configuration" feels like a property of the overall build - I guess one could have different psa settings for different packages in the same workspace, but that sounds contrived to me: the usual situation in my experience is that one wants to censor some warning globally.

spaghetto/bin/src/Flags.purs
```diff
@@ -4,13 +4,87 @@ import Spago.Prelude
...
+psaStrict ∷ ArgParser (Maybe Boolean)
+psaStrict =
+  ArgParser.flag [ "--strict" ]
+    "Promotes project sources' warnings to errors"
+    # ArgParser.boolean
+    # ArgParser.optional
+
+psaCensorWarnings ∷ ArgParser (Maybe Boolean)
+psaCensorWarnings =
+  ArgParser.flag [ "--psa-censor-warnings" ]
```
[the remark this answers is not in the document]
f-f: Yeah I agree that we should not use the `psa` prefix for these flags. I think `--censor-warnings` works, but we I agree that these flags could have something like a common prefix to signal that they are configuring the same group of features. Maybe `--censor-build-warnings`?

spaghetto/bin/src/Flags.purs
f-f: That's alright - the codebases uses booleans all over to avoid conversions, but this works too

spaghetto/bin/src/Flags.purs
```diff
@@ -4,13 +4,87 @@ import Spago.Prelude
...
+psaStatVerbosity = ArgParser.optional $ ArgParser.choose "StatVerbosity"
+  [ Core.VerboseStats <$ ArgParser.flag [ "--psa-verbose-stats" ] "Show counts for each warning type"
+  , Core.NoStats <$ ArgParser.flag [ "--psa-censor-stats" ] "Censor warning/error summary"
+  ]
+
+psaStashFile ∷ ArgParser (Maybe (Either Boolean String))
+psaStashFile = ArgParser.optional $ ArgParser.choose "stash"
+  [ ArgParser.argument [ "--psa-stash-file" ] "Enable persistent warnings using a specific stash file"
+      # ArgParser.unformat "FILE" (Right <<< Right <<< Paths.mkLocalCachesCustomStashFile <<< Path.basename)
+  , Left true <$ ArgParser.flag [ "--psa-stash" ] "Enable persistent warnings using default stash file location"
```
f-f, line 79: What's the value of making this location configurable?

spaghetto/bin/src/Flags.purs
```diff
@@ -4,13 +4,87 @@ import Spago.Prelude
...
+
+psaStatVerbosity :: ArgParser (Maybe Core.StatVerbosity)
+psaStatVerbosity = ArgParser.optional $ ArgParser.choose "StatVerbosity"
+  [ Core.VerboseStats <$ ArgParser.flag [ "--psa-verbose-stats" ] "Show counts for each warning type"
+  , Core.NoStats <$ ArgParser.flag [ "--psa-censor-stats" ] "Censor warning/error summary"
+  ]
+
+psaStashFile ∷ ArgParser (Maybe (Either Boolean String))
+psaStashFile = ArgParser.optional $ ArgParser.choose "stash"
+  [ ArgParser.argument [ "--psa-stash-file" ] "Enable persistent warnings using a specific stash file"
```
f-f, line 77: The name of this flag is a bit confusing, I'd prefer something more direct (I mean, it doesn't matter if these are persisted to a file or anywhere else) like `--persist-warnings`

spaghetto/bin/src/Main.purs
```diff
@@ -68,6 +72,20 @@ type InstallArgs =
   , output :: Maybe String
   , pedanticPackages :: Boolean
   , ensureRanges :: Boolean
+  , jsonErrors :: Boolean
+  , psaArgs :: PsaArgs
+  }
+
+type PsaArgs =
```
f-f, line 79: There already is a `PsaOptions` type in the Psa source so there's no need to bundle these together here, and I think it'd make more sense to group by functionality, e.g. by calling this something like `BuildFlags`, and adding here all the other build-related things (as grouped in the configuration file) like `output`, `pedanticPackages`, `jsonErrors`, etc

spaghetto/bin/src/Main.purs
```diff
@@ -133,10 +157,13 @@ type BundleArgs =
   , pedanticPackages :: Boolean
   , type :: Maybe String
   , ensureRanges :: Boolean
+  , jsonErrors :: Boolean
+  , psaArgs :: PsaArgs
   }
 
 type PublishArgs =
   { selectedPackage :: Maybe String
+  , psaArgs :: PsaArgs
```
f-f, line 166: Maybe the publishing build should not be configurable? I mean, the compiler is going to build a few times in different configurations, so not sure the user should look at that at all - we can probably skip these flags (the idea is that every flag adds cognitive weight to folks using a command, and this is important for `publish`) and give sensible defaults when we build the environment

spaghetto/bin/src/Main.purs
f-f: Agreed. As noted above, we should generally try to minimise the amount of flags exposed in any command, since they all add a burden when trying to use a specific command.

spaghetto/src/Spago/Command/Build.purs
```diff
@@ -69,19 +75,51 @@ run opts = do
       Nothing -> args
       Just output -> args <> [ "--output", output ]
 
+  -- find the output flag and die if it's there - Spago handles it
```
f-f, line 78:
```suggestion
  -- find the json-errors flag and die if it's there - Spago handles it
```

spaghetto/src/Spago/Config.purs
```diff
@@ -322,7 +325,10 @@ data WithTestGlobs
   | OnlyTestGlobs
 
 sourceGlob :: WithTestGlobs -> PackageName -> Package -> Array String
-sourceGlob withTestGlobs name package = map (\p -> Path.concat [ getPackageLocation name package, p ])
+sourceGlob withTestGlobs name package = map snd $ sourceGlob' withTestGlobs name package
+
+sourceGlob' :: WithTestGlobs -> PackageName -> Package -> Array (Tuple String String)
```
f-f, line 330: Do we need this change at all? Every `WorkspacePackage` contains a `path`, so we could figure the source location from there:
https://github.com/purescript/spago/blob/b92239d18971779966c15460e03166acbc9c042a/spaghetto/src/Spago/Config.purs#L109-L110

spaghetto/src/Spago/Psa.purs
```diff
@@ -0,0 +1,197 @@
...
+      Right success -> { output: success.stdout, exitCode: Just success.exitCode, err: Nothing }
+  arrErrorsIsEmpty <- forWithIndex (Str.split (Str.Pattern "\n") result'.output) \idx err ->
+    case jsonParser err >>= CA.decode psaResultCodec >>> lmap CA.printJsonDecodeError of
+      Left decodeErrMsg -> do
+        logDebug $ Array.intercalate "\n"
+          [ "Failed to decode PsaResult at index '" <> show idx <> "': " <> decodeErrMsg
+          , "Json was: " <> err
+          ]
+        -- Note to reviewer:
+        -- Should a stash decode error cause a non-zero exit?
```
f-f, line 100: The rule of thumb is "we crash if we can't figure out how to proceed". This seems an entirely recoverable situation, in the sense that we can in fact just move on without a stash file.
I would probably log a warning though (instead of a debug log), so we get bug reports - because a decoding failure is 95% likely to be our bug.

spaghetto/src/Spago/Psa.purs
f-f: I am not sure how psa is changing the errors here?
But the thing is that editors might get confused with a stash file, so maybe we should output all the json errors anyways. @nwolverson likely has more useful input on this

spaghetto/src/Spago/Psa.purs
```diff
@@ -0,0 +1,197 @@
...
+    die [ "Failed to build." ]
+
+  where
+  insertFilenames = foldr \x s -> maybe s (flip Set.insert s) x.filename
+  loadNothing _ _ = pure Nothing
+
+  isEmptySpan filename pos =
+    filename == "" || pos.startLine == 0 && pos.endLine == 0 && pos.startColumn == 0 && pos.endColumn == 0
+
+  -- TODO: Handle exceptions
```
f-f, line 129: Should we handle these exceptions? 😄

spaghetto/src/Spago/Psa/Printer/Default.purs
f-f: I can try this out on a few terminals and report. But I'd say that if the colors are wrong it's likely that we'll get bug reports quickly, as this is going to be the default behaviour 🙂

spaghetto/src/Spago/Psa/Printer/Json.purs
```diff
@@ -0,0 +1,22 @@
...
+import Prelude
+
+import Data.Argonaut.Core (stringify)
+import Data.Codec.Argonaut as CA
+import Effect (Effect)
+import Effect.Console as Console
+import Spago.Psa.Output (Output)
+import Spago.Psa.Types (psaResultCodec)
+
+print :: Output -> Effect Unit
```
f-f, line 14: This is a tiny module, I wonder if we could collapse it into some other module?

spaghetto/src/Spago/Psa/Types.purs
f-f: I think they are generally useful, let's keep them.

spaghetto/src/Spago/Purs.purs
f-f: I would expect this to take the output pipe and do things with the stream of the output so that things happen in real time rather than being buffered.
Is this how the current psa works?

f-f (commented):
Thanks for putting it together Jordan!
This looks great and while I left a few comments, they are mostly about UX, and overall the patch is basically ready to go

f-f:
I was wondering if @natefaubion is ok with us forking psa in here - I suggested this to Jordan as the feeling that I got from the issue trackers is that the codebase is mostly in a frozen state, and I wanted to move quickly here.
We have a whole set of changes happening here, so I'd like to make sure if this is the preferred setup for everyone, or would Nate prefer that we upstream things and possibly take over maintainership of the original repo?
My main worry is about folks being confused about and mention of `psa` here (even though we're trying to avoid them) possibly having a mismatched behaviour with the original `psa` tool.

f-f:
Thanks Nate! If you have no preference on this then my personal preference is that we stick to this fork here so we can iterate quickly on the new Spago. Once things get more stable we can think about upstreaming improvements to your repo.

spaghetto/src/Spago/Psa.purs
JordanMartinez: Possibly later, but that's what was in the original PSA code :smile:

spaghetto/README.md
JordanMartinez: Oh good point. If we drop the "psa" idea and just see this as a build option, then it can go up one more level.

spaghetto/README.md
JordanMartinez: A few comments about this:
- I only added a separate build option in the workspace spot because I wasn't sure how to deal with building the entire workspace (e.g. when no package is selected). Originally, it was just for each package directly.
- I can imagine some codebases wanting `--strict` to be required (e.g. core libraries) for their source globs but not for their test globs. Right now, Spago compiles both source and test code in the same run, which I find problematic because sometimes the src code is first updated and then the test code. Using the same set of PSA options for both globs is problematic. I imagine that will only be more so across packages.
- If the workspace option was provided, there are two ways to interpret it.
    1. What I've done in this PR: the workspace options are used when no packages are selected.
    2. What I'm hesitant to do because of readability: the workspace options could act as a "defaults" when everything else is otherwise unspecified. PSA options precedence would then be `CLI arg > package option > workspace option > spago PSA defaults`.
- Given a package `foo` which stores its source code in `foo/src/**/*.purs`, I may not want to put my test code in `foo/test/**/*.purs`. As an example, consider https://github.com/thomashoneyman/purescript-halogen-hooks/issues/24. The test code may be its own package suitable for downstream consumers. In order for Halogen Hooks to support that, it would need to define 2-3 packages:
    - src code for halogen-hooks
    - src code for halogen-hooks-test
    - test code that uses halogen-hooks-test to test halogen-hooks. I'm not sure if this would go in the `halogen-hooks-test`'s `test/**/*.purs` source globs or be its own not-to-be-published package that depends on the other two.

spaghetto/bin/src/Flags.purs
JordanMartinez: That can work, but this doesn't just persist warnings. It also persists compiler errors. We could do `--persist`, but then it's not clear what's being saved.

spaghetto/bin/src/Main.purs
JordanMartinez: If `spago build` normally uses `--strict`, I'd expect `spago publish` to operate likewise (e.g. `--strict` will fail publishing if warnings exist). I think the stash file should be used as otherwise some compiler warnings that existed in a previous `build` may not be reported by `purs` a second time.

spaghetto/bin/src/Flags.purs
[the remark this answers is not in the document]
JordanMartinez: I decided to drop `censor-lib` and `censor-src` and move them all into `censor-build-warnings`. Technically, it's three different flags for 2 values, so using a single flag makes this easier to track.

spaghetto/src/Spago/Purs.purs
JordanMartinez: Just to clarify, when `purs` outputs its errors, it does so to `stdout`. When it outputs "Compiling ..." messages, it does so to `stderr`. So, the above code is saying:
1. pipe `stderr` to the parent's `stderr`, so that we see "Compiling ..." messages real-time.
2. don't pipe `stdout` to the parent's `stdout`, so that we don't see `{"warnings":[],"errors":[]}` in the console.

Are you not seeing a real-time output of the "compiling" messages?

JordanMartinez:
As I've spent time on this, I'm actually wondering why this work isn't being done in the compiler itself as that would remove the underlying reason for why `psa` even exists. And then, this work wouldn't even be necessary and the IDE wouldn't need to depend on `psa` either. Thoughts?

f-f:
@JordanMartinez indeed it would be nice if the compiler did all of this and `psa` would not have reason to exist - however the issue is not straightforward, so (as the linked thread shows) I am of the opinion that we'd be better off getting this to work in Spago first, and then eventually deprecate this functionality once the relevant things are merged in the compiler.

f-f (approved):
I'll merge this now - thanks Jordan!

PR #958, opened 2023-05-01
Add support for esbuild --external flag.

jsparkes:
See https://esbuild.github.io/api/#external.

The --external flag allows users to exclude modules from the build.  The case I ran into was importing "fs-extra" into a foreign js file for work on psvm-ps.  It referenced the module "electron" even though I was bundling for node.  Now I can add "--external electron" to the arguments to "spago bundle-app" and it wil bundle correctly.  It is possible to exclude multiple modules by using a comma separated list as the argument.  I tried to split using using a regular expression, but couldn't figure out how to include Text.Regex in Haskell.  I left that code in comments for someone who isn't a beginner.  It currently works, but only splits on commas, not whitespace.

I made some sort of mistake, and a diff for Config.purs is included.  I think it changed in master, but somehow got duplicated here.

Writing tests is beyond my current Haskell ability.  A test  should check that the `--external module1,module2` should expand into `--external:module1 --external:module2` in the esbuild args.

TODO

### Checklist:


f-f:
Hi @jsparkes - sorry for delay but we've been going through a rewrite. The old Haskell codebase is now living a https://github.com/purescript/spago-legacy, and it's basically frozen. The new codebase lives here and it's written in PureScript.

Please try it out and see if it solves your issue (can install it with `npm install spago@next`), and let's discuss further if it does not

Issue #917, opened 2022-10-18
Support spago.lock files


thomashoneyman:
The registry-compliant Spago specifies dependencies as either entries in a package set or via explicit version bounds of the form `>=X.Y.Z <X.Y.Z`, where libraries published to the registry _must_ use version bounds. This means that it's no longer possible to tell the exact versions that a project is using from its package set alone.

Lockfiles solve this problem by listing the exact resolutions a project is currently using. They'll also typically list other metadata sufficient for the package manager to produce a build from the lockfile alone, such as:

* the dependency version (a registry version, a ref, ...)
* the location of the package source (registry packages can omit this, but we'd need this if the package is overridden, such as to a fork)
* the hash of the package contents (commonly called "integrity", and used to ensure the code downloaded exactly matches what the lockfile expects.)

There may be other fields we choose to support, if they are useful to Spago. Selfishly, I'm most interested in the lockfile storing package hashes (which we can get from the registry, or compute in the case of an override) because this is necessary for Nix tooling to be able to support Spago projects without generating additional files. However, lockfiles are useful beyond just making it easier for Nix tools to work.

f-f:
Lockfiles are not necessary for the situations where one uses only a package set and fixed versions (which is what Spago has been dealing with this whole time), but become unavoidable when using a solver, as one would want to save the plan and not recompute it every time.

thomashoneyman:
Right. Although you may also want lockfiles for the integrity checks as well even if only using package sets (some of this was previously done with Dhall). But with this in mind it makes sense for lock files to be optional if you are using package sets.

thomashoneyman:
And the hashing of the package set works even if you have overrides? It’s not just a hash of the upstream package set?

f-f:
It is a hash of the package set, the overrides are up to the users, which are instructed to not point at branches (because they might move), but only to commits or tags.

That's not a nixified build and things might not be reproducible in certain corner cases, but it's been good enough for users so far, which is why I'm comfortable with having an optional lockfile for package-set builds.

thomashoneyman:
In #962 we implemented initial support for writing Spago lockfiles, including:

* The lockfile format and tests
* Reading and writing lockfiles from disk
* Determining whether to write a lockfile based on a user's config and whether `lock: true` is set

We've got a number of things still remaining to do, including:

**CLI commands**
- [ ] Add a --generate-lockfile that pulls all git dependencies, regenerates lockfile
[...]

PR #959, opened 2023-05-13
Support Spago lockfiles

thomashoneyman:
### Description of the change

This PR implements the first part of #917. It:

* adds lockfile support to the Spago config (a `lock` option) which controls a `spago.lock` file in JSON
* specifies a lockfile format for workspace, registry, and remote git packages suitable for fetching
* threads the lockfile through the `BuildOptions` record

I haven't attempted to use or generate lockfiles yet because we should discuss the implementation.

## Proposed lockfile format

I have created a lockfile in JSON that has the below form:

```jsonc
  {
    // The lockfile version, in case we need to make changes to the format in the future
    "lockfile": "1.0.0",
    // A list of all packages in the workspace (including test dependencies) at specific resolved versions
    "packages": {
      "console": "git:https://github.com/purescript/purescript-console.git#3b83d7b792d03872afeea5e62b4f686ab0f09842",
      "effect": "registry:4.0.0#sha256-eBtZu+HZcMa5HilvI6kaDyVX3ji8p0W9MGKy2K4T6+M=",
      "my-library": "workspace:./my-library",
      "prelude": "git:https://github.com/purescript/purescript-libraries.git#3b83d7b792d03872afeea5e62b4f686ab0f09842?subdir=prelude"
    }
  }
```

Specifically, this format supports three types of packages, where each lock entry is prefixed by its type (`registry:` for registry packages, `git:` for remote git packages, and `workspace:` for local packages that point at a path). Then:

* Registry packages use the format `version#hash` where the hash comes from the registry metadata entry
* Remote git packages use the format `url#commit-sha` plus an optional `?subdir=` specifier if a subdirectory is needed
* Workspace packages simply record the path to the package

In https://github.com/thomashoneyman/purescript-nix-example I am using a variant of this lockfile to successfully produce a Nix build from a Spago workspace.

## Proposed lockfile usage

Lockfiles are meant to make builds (more) reproducible by locking down exact versions for dependencies that would otherwise change depending on what day you installed them. For example, if you specify the dependency `console: >=1.0.0 <2.0.0` then the exact version installed will change anytime a new version comes out in that range; if you specify (in pseudocode) the dependency  `console: github.com/purescript/purescript-console@master` then the exact version installed will depend on the tip of the master branch that day.

**Determining whether or not to use a lockfile**

In Spago's case, if you use a package set and have no additional dependencies then you will already essentially have locked dependencies. To prevent package-set-only users from having to generate a lockfile, I have added a `lock: <true|false>` config item to the workspace. If present and true, then a lockfile will be required. If present and false, then a lockfile will never be generated or used. If missing, then a lockfile will only be used if you specify dependencies with a version range or specify additional dependencies at non-exact versions.

For example, if you have a config with no `lock` setting, and you're using a package set, but then write `console: >=1.0.0 <2.0.0` as a dependency, then a lockfile will be generated that specifies the version seen in your package set (such as 1.0.1). Or, if you're using a package set but then specify an extra package `console: github.com/purescript/purescript-console@master` then a lockfile will be generated that specifies the commit sha at the time you install.

**Using the lockfile**

The lockfile should be used any time Spago needs to install packages for your workspace. This happens, for example, when you fetch/install, build, bundle, or run your project. In each case Spago needs to ensure all your dependencies are installed. For each of these commands, _if_ a lockfile should be used, then Spago should either:

1. Parse and use the existing lockfile to determine exact dependency versions, or if it is missing then
2. Resolve dependencies and then generate a lockfile with the resolutions

If Spago is using a lockfile then it will skip any dependency resolution and just verify that there is nothing in the config's list of dependencies that isn't also in the lockfile, as that would indicate an error.

**Additional commands**

We may wish to add a new command and a new flag:

  * `spago generate-lockfile` could take a spago config and generate / update a lockfile. (For reference, there is a [generate-lockfile](https://doc.rust-lang.org/cargo/commands/cargo-generate-lockfile.html) for Cargo). We could then re-use this code in other commands like `Fetch.purs` so we have a standard process for generating / updating a lockfile when necessary
  * `spago install` or any other command that manipulates the installed dependencies could take a flag `--locked` or `--frozen` to guarantee they use the lockfile versions without attempting to update the lockfile, or they could take a flag `--update` to make updating the lockfile opt-in — you only get new versions of packages if you pass `--update`.

## Discussion Questions

**1. Lockfile format**
I have opted for JSON simply because it's so easy to work with, but since the Spago config files are in yaml I would understand wanting to do lockfiles in yaml as well. The format would then be, for example:

```yaml
lockfile: 1.0.0
packages:
  - effect: registry:4.0.0#sha256-eBtZu+HZcMa5HilvI6kaDyVX3ji8p0W9MGKy2K4T6+M=
```

**2. Lock entry format**
I have opted to make lock entries strings so that the lockfile doesn't become enormous with unnecessarily repeated fields. However, you may prefer a more verbose format, such as the one seen in the purescript-nix example:
https://github.com/thomashoneyman/purescript-nix-example/blob/d5742c141c8a59781eb3be1dbd71f9bdf051cbd2/spago.lock#L1C1-L24

**3. Local packages**
I am not sure what should be done about "local" packages, ie. packages that can exist outside the workspace. They don't make sense in a lockfile because they're a mutable part of the build that depends on the state of your machine, separate from the workspace itself.


f-f:
Thanks for putting this together @thomashoneyman! 

I haven't looked at the code yet and I'll only answer the general questions here:
1. Lockfile format: being the config in YAML, it's definitely less confusing to have the lockfiles be YAML too. (see e.g. [Stack lockfiles](https://docs.haskellstack.org/en/stable/lock_files/)). Is Nix ok with parsing YAML?
2. Lock entry format: I do prefer the more structured approach from your link, that'd be my preference
3. Local packages: I think the two options that we have are: (1) refuse to generate a lockfile, display a warning, etc, or (2) include their paths in the lockfile anyways.
  Note that one of the usecases for local packages is to be able to include shared code in a repo that is building with more than one backend (you can't include the shared code in both builds, as the package discovery mechanism only traverses subtrees), so I'd be leaning towards option 2 here.

A few more passing comments:
- > The lockfile version, in case we need to make changes to the format in the future
  
  We shold not need to store the version here, we'll be backwards compatible with old lock files anyways, until we decide to break compatibility. This is how it will work for the Config file too, which in fact does not have a version field.
- > if you're using a package set but then specify an extra package [...] then a lockfile will be generated that specifies the commit sha at the time you install
  
  In this case I'd prefer failing the build to force the user to use a commit hash, rather than generate a lockfile
- > We may wish to add a new command and a new flag
  
  Definitely yes for the new command, but we can skip the flag and always edit the lockfile too, as doing it only sometimes feels like a footgun.

thomashoneyman:
>  Is Nix ok with parsing YAML?

Yes (by converting to JSON first).

> Local packages: I think the two options that we have are: (1) refuse to generate a lockfile, display a warning, etc, or (2) include their paths in the lockfile anyways.

To be clear, if the package is within the repository then it's OK. As far as the lockfile is concerned, a "local" package and a "workspace" package are the same thing — a path to a directory in the repository. The only issue is when the package is outside the repository altogether. If Spago allows that, then I think we should refuse to generate a lockfile.

> [Regarding lockfile versioning]: We should not need to store the version here, we'll be backwards compatible with old lock files anyway

OK, I'll do my best to keep in mind backwards compatibility.

> [Regarding the use of extra_packages]: In this case I'd prefer failing the build to force the user to use a commit hash, rather than generate a lockfile

I'm thinking of when users want to specify a branch or tag in their extra packages. Surely we shouldn't force them to only use a commit hash? After all, this workflow is pretty common.

However, maybe we can make the heuristic simpler for when to generate a lockfile — if you use a package set then we don't generate one by default, unless you also specify ranges in your dependencies. And we don't worry about the extra_packages.

f-f:
> If Spago allows that, then I think we should refuse to generate a lockfile.

That makes sense, but what would we do if a lockfile is there and we do not update it because of this? Just accept that the lockfile is out of date?

> OK, I'll do my best to keep in mind backwards compatibility.

To be clear, we don't need to do anything right now, but when we do change the format of the lockfile then we should also support the old formats.

> However, maybe we can make the heuristic simpler for when to generate a lockfile — if you use a package set then we don't generate one by default, unless you also specify ranges in your dependencies. And we don't worry about the extra_packages.

Yeah, I think this makes most sense to start with, we can always refine it later!

thomashoneyman:
I've changed the lockfile format to YAML in the example repo — here's how it looks right now:

https://github.com/thomashoneyman/purescript-nix-example/blob/8b48c706ba70385dda87539876b9a5151f596df5/spago.lock#L1-L16

Since the lockfile will use only simple YAML we can do a little YAML parser written in pure nix and skip import-from-derivation:
https://github.com/thomashoneyman/purescript-nix-example/blob/main/nix/from-yaml.nix

I think this keeps things pretty simple and fast.

thomashoneyman:
> If Spago allows that, then I think we should refuse to generate a lockfile.
> > That makes sense, but what would we do if a lockfile is there and we do not update it because of this? Just accept that the lockfile is out of date?

Perhaps if this is encountered then Spago warns? "<dependency> is a local path outside the repository, lockfile will be ignored"?

PR #962, opened 2023-05-26
Implement first support for writing lockfiles

f-f:
Setting `lock: true` in the config will write out a lockfile which _should_ be usable by Nix.

Reading lockfiles and doing interesting things with it on the Spago side will come in future PRs.


thomashoneyman:
Supersedes #959

thomashoneyman (approved):

PR #963, opened 2023-05-29
Proposal: Provide a Nix flake for Spago in Nix

thomashoneyman:
As per a few discussions we've had lately in the #spago channel of the https://purescript.org/chat, this PR demonstrates building Spago in this repository via the [`spago-nix`](https://github.com/thomashoneyman/spago-nix) project. This avoids, for example, the many horrors of node2nix, napalm, npmlock2nix, buildNpmPackage, etc., and includes conveniences for the PureScript portion of building the Spago bundle.

This PR demonstrates building Spago itself with Nix and exporting itself from a Nix flake so that users downstream can e.g. run:

```
nix run github:purescript/spago
```

and have access to Spago (or use it in their Nix builds). At the moment this is for demonstration purposes only — I don't intend this to be merged until we can discuss the implications.


spaghetto/spago.lock
```diff
@@ -1,7 +1,7 @@
 workspace:
   packages:
     spago:
-      path: ./
```
thomashoneyman, line 4: This is problematic for spago-nix, but perhaps we could work around it.

spaghetto/shell.nix
```diff
@@ -27,5 +27,6 @@ mkShell {
     easy-ps.purescript-language-server
     easy-ps.purs-tidy
     nodejs
+    esbuild
```
thomashoneyman, line 30: I moved this to the shell instead of npm dependencies, since it's not really needed there.

flake.nix
```diff
@@ -0,0 +1,100 @@
...
+
+    mkFlakeApps = pkgs: packages: {
+      default = {
+        type = "app";
+        program = "${
+          pkgs.writeShellScriptBin "spago-run" ''
+            echo "$@"
+            ${pkgs.nodejs}/bin/node --eval 'require("${packages.default}/bundle.js").main()' "$@"
+          ''
+        }/bin/spago-run";
```
thomashoneyman, line 75: This doesn't work right now — perhaps you know how to pass through arguments, @f-f, or how to change the esbuild command?

thomashoneyman:
An alternate implementation can be seen here:
https://github.com/thomashoneyman/purescript-nix/blob/main/spago/mkSpago.nix

PR #972, opened 2023-07-15
Add include and exclude to publish config

i-am-the-slime:
This change adds config keys  
 -  `include` for specifying an array of globs matching files that shall be published in addition to the default (anything matching in `/src/**.purs`)
 -  `exclude` for finally specifying an array of globs to remove from the union of what's matched by the default and `include`

It's the first step to supporting this in the registry. Once this is merged, we'll add support and validation for this in the `registry-dev` repo and then finally finish with a PR to `spago/app`.

See #967 for discussion leading to this change.

### DIFF
--- a/spaghetto/core/src/Config.purs
+++ b/spaghetto/core/src/Config.purs
@@ -94,6 +94,8 @@ type PublishConfig =
   { version :: Version
   , license :: License
   , location :: Maybe Location
+  , include :: Maybe (Array FilePath)
+  , exclude :: Maybe (Array FilePath)
   }
 
 publishConfigCodec :: JsonCodec PublishConfig
@@ -101,6 +103,8 @@ publishConfigCodec = CAR.object "PublishConfig"
   { version: Version.codec
   , license: License.codec
   , location: CAR.optional Location.codec
+  , include: CAR.optional (CA.array CA.string)
+  , exclude: CAR.optional (CA.array CA.string)
   }
 
 type RunConfig =


spaghetto/core/src/Config.purs
```diff
@@ -94,13 +94,17 @@ type PublishConfig =
   { version :: Version
   , license :: License
   , location :: Maybe Location
+  , include :: Maybe (Array FilePath)
+  , exclude :: Maybe (Array FilePath)
```
thomashoneyman, line 98: These are really globs, right? Should we newtype and sanitize this the way it’s done in the registry?

https://github.com/purescript/registry-dev/blob/6c1209c9eaad10269a2e33445f9beb61b7ac0d12/lib/src/Internal/Path.purs#L29

f-f: This part of the PR is about specifying things in the config file. I am not sure we'd want to parse/validate things here already, all this stuff is mostly about reading things in.

Even in the registry the `sanitizePath` function that you link is only used at glob-matching time, inside `Registry.Foreign.FastGlob.match'`.
We use that same function to match globs here, so if I understand correctly we're already getting the sanitization for free.

f-f (approved):

spaghetto/core/src/Config.purs
i-am-the-slime: Oh so it's impossible to add sources from a sibling project in the same project? I guess we can always relax that if the need ever arises.

thomashoneyman: It should be impossible to add sources from a sibling project as far as the registry (publishing) is concerned, though I suppose it’s fine for Spago

f-f: I don't think it should be fine for Spago either, the proper mechanism for adding sibling projects is through a package dependency.

PR #993, opened 2023-09-11
bump `text` lower bound, support GHC 9.4

peterbecich:
### Description of the change

- GHC 9.4
- bump Stack resolver
- bump `text` lower bound
  - succeeds: `cabal build all --constraint 'text >= 2.0' --enable-tests`
  - new Stack resolver has `text` 2.0: https://www.stackage.org/lts-21.11
  - removing the lower bound on `text` entirely would also be fine, but might as well keep it to stay consistent with the Stack resolver  

I have not run the entire test suite locally, let's see if it succeeds in CI

Bumping the lower bound on `text` is helping me to get Spago on NixPkgs building: https://github.com/NixOS/nixpkgs/issues/253198
It appears `text` 1.* is no longer available in NixPkgs: https://search.nixos.org/packages?channel=unstable&show=haskellPackages.text_2_0_2&from=0&size=50&sort=relevance&type=packages&query=haskell+text

Thanks
  
### Checklist:

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,0 +1,5 @@
+packages:
+  .
+
+allow-newer:
+  aeson-better-errors:aeson
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,11 +1,11 @@
-resolver: lts-20.26
+resolver: lts-21.11
 packages:
   - .
 extra-deps:
-  - Cabal-3.6.3.0
   - semver-range-0.2.8
-  - fsnotify-0.4.1.0
-  - versions-6.0.1
+  - async-pool-0.9.2
+  - bower-json-1.1.0.0
+  - aeson-better-errors-0.9.1.1
 allow-newer: true
 nix:
   packages: [zlib]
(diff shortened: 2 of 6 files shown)


stack.yaml
```diff
@@ -1,20 +1,20 @@
...
 nix:
   packages: [zlib]
 flags:
   aeson-pretty:
     lib-only: true
   # This is so we don't depend on libtinfo
-  haskeline:
-    terminfo: false
-    examples: false
+  # haskeline:
```
peterbecich, line 16: The new Stack resolver does not recognize this.

I don't know what the implications are of this change

cabal.project
```diff
@@ -0,0 +1,5 @@
+packages:
+  .
+
+allow-newer:
+  aeson-better-errors:aeson
```
peterbecich, line 5: can be removed if this is merged and new version released: https://github.com/hdgarrood/aeson-better-errors/pull/24

peterbecich:
Looks like there are some test failures

peterbecich:
Yes, you're correct, thanks @f-f

PR #994, opened 2023-09-14
Run tests in cwd

### LINKED ISSUE #984: `spago test` runs script in package's directory, not workspace directory
See https://github.com/JordanMartinez/spago-next-bug for a reproducible example, which surprised me.

I think there are three responses to this issue.:
1. always run `spago test` in the workspace directory
1. always run `spago test` in the package directory
1. always run `spago test` in the directory in the current working directory

Most of the time, if I'm running a test, I execute the `spago test` command when my current working directory is the workspace directory.

JordanMartinez:
### Description of the change

Fixes #984.

### Checklist:

### DIFF
--- a/test-fixtures/spago-subpackage-test-cwd.purs
+++ b/test-fixtures/spago-subpackage-test-cwd.purs
@@ -4,11 +4,8 @@ import Prelude
 
 import Effect (Effect)
 import Effect.Console (log)
-import Node.Path as Path
 import Node.Process as Process
 
 main :: Effect Unit
 main = do
-  pwd <- Process.cwd
-  pwd' <- Path.resolve [] pwd
-  log pwd'
+  Process.cwd >>= log
--- a/test-fixtures/spago-subpackage-test-cwd.purs
+++ b/test-fixtures/spago-subpackage-test-cwd.purs
@@ -4,8 +4,11 @@ import Prelude
 
 import Effect (Effect)
 import Effect.Console (log)
+import Node.Path as Path
 import Node.Process as Process
 
 main :: Effect Unit
 main = do
-  Process.cwd >>= log
+  pwd <- Process.cwd
+  pwd' <- Path.resolve [] pwd
+  log pwd'
--- a/test-fixtures/spago-subpackage-test-cwd.purs
+++ b/test-fixtures/spago-subpackage-test-cwd.purs
@@ -12,6 +12,5 @@ main :: Effect Unit
 main = launchAff_ do
   readTextFile UTF8 filePath >>= log
 
--- DO NOT EDIT. This value is replaced in the corresponding spago test
 filePath :: String
-filePath = "FILE_PATH"
+filePath = "foo.txt"
--- a/test-fixtures/spago-subpackage-test-cwd.purs
+++ b/test-fixtures/spago-subpackage-test-cwd.purs
@@ -1,0 +1,11 @@
+module Subpackage.Test.Main where
+
+import Prelude
+
+import Effect (Effect)
+import Effect.Console (log)
+import Node.Process as Process
+
+main :: Effect Unit
+main = do
+  Process.cwd >>= log
--- a/test-fixtures/spago-subpackage-test-cwd.purs
+++ b/test-fixtures/spago-subpackage-test-cwd.purs
@@ -3,9 +3,15 @@ module Subpackage.Test.Main where
 import Prelude
 
 import Effect (Effect)
-import Effect.Console (log)
-import Node.Process as Process
+import Effect.Aff (launchAff_)
+import Effect.Class.Console (log)
+import Node.Encoding (Encoding(..))
+import Node.FS.Aff (readTextFile)
 
 main :: Effect Unit
-main = do
-  Process.cwd >>= log
+main = launchAff_ do
+  readTextFile UTF8 filePath >>= log
+
+-- DO NOT EDIT. This value is replaced in the corresponding spago test
+filePath :: String
+filePath = "FILE_PATH"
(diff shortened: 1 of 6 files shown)


JordanMartinez:
There are tests unrelated to this PR that currently fail locally. So, the third commit above proves that this change works. Once proven by CI, I'll revert the third commit.

f-f:
@JordanMartinez the CI for the PureScript side of things is not running yet - I am fixing it at the moment, so that we'll be able to get meaningful reporting here on GH

JordanMartinez:
Oh! That would make sense.

f-f:
We have PureScript CI now! Could you rebase on the latest commit? I had to shuffle files around unfortunately, so it might be easier to just copy the patch to a new branch instead of trying to merge.

JordanMartinez:
Sure. I'll get to that when I can.

test/Prelude.purs
```diff
@@ -66,8 +68,15 @@ checkFixture filepath fixturePath = do
...
-checkResultAndOutputs maybeOutFixture maybeErrFixture resultFn execResult = do
+plusDependencies :: Array String -> Config -> Config
+plusDependencies deps config = config
+  { package = config.package <#> \p -> p { dependencies = p.dependencies <> (Dependencies $ Map.fromFoldable $ map mkDep deps) }
+  }
+  where
+  mkDep p = Tuple (unsafeFromRight $ PackageName.parse p) Nothing
+
+checkResultAndOutputs' :: Maybe String -> Maybe FilePath -> (Either ExecError ExecResult -> Boolean) -> Either ExecError ExecResult -> Aff _
+checkResultAndOutputs' maybeOutStr maybeErrStr resultFn execResult = do
```
JordanMartinez, line 79: Thoughts on renaming this to something else? The `'` isn't much of a difference and the type signature is the same as `checkResultAndOutputs`. So, I see this as an easy gotcha when writing tests in the future.

f-f: Good point. How's `checkResultAndOutputsStr`?

JordanMartinez: Done.

f-f:
What if you write a file in the root of the repo and try to read its content from the script in the test? Then we don't need to verify what the cwd is, just that we were able to read the file from the right place.

f-f:
It looks like some macOS tests are just timing out, let's try to bump the general test timeout from 30s to maybe 60? (It's in `test/Spago.purs`)

test-fixtures/spago-subpackage-test-cwd.purs
```diff
@@ -0,0 +1,17 @@
...
+import Node.Encoding (Encoding(..))
+import Node.FS.Aff (readTextFile)
+
+main :: Effect Unit
+main = launchAff_ do
+  readTextFile UTF8 filePath >>= log
+
+-- DO NOT EDIT. This value is replaced in the corresponding spago test
+filePath :: String
+filePath = "FILE_PATH"
```
f-f, line 17:
```suggestion
filePath = "./foo.txt"
```

I think we don't need to do the whole replacing thing, but can just use a relative path?

test/Prelude.purs
```diff
@@ -66,8 +68,15 @@ checkFixture filepath fixturePath = do
...
-checkResultAndOutputs :: Maybe FilePath -> Maybe FilePath -> (Either ExecError ExecResult -> Boolean) -> Either ExecError ExecResult -> Aff _
-checkResultAndOutputs maybeOutFixture maybeErrFixture resultFn execResult = do
+plusDependencies :: Array String -> Config -> Config
+plusDependencies deps config = config
+  { package = config.package <#> \p -> p { dependencies = p.dependencies <> (Dependencies $ Map.fromFoldable $ map mkDep deps) }
+  }
+  where
+  mkDep p = Tuple (unsafeFromRight $ PackageName.parse p) Nothing
+
+checkResultAndOutputsStr :: Maybe String -> Maybe FilePath -> (Either ExecError ExecResult -> Boolean) -> Either ExecError ExecResult -> Aff _
```
f-f, line 78:
```suggestion
checkResultAndOutputsStr :: Maybe String -> Maybe String -> (Either ExecError ExecResult -> Boolean) -> Either ExecError ExecResult -> Aff _
```

test-fixtures/spago-subpackage-test-cwd.purs
thomashoneyman: Looks like spago has historically had windows CI, so should this be (“.” <> Path.sep <> “file.txt”) instead?

f-f: It might also work to just use `foo.txt`

JordanMartinez:
> It looks like some macOS tests are just timing out, let's try to bump the general test timeout from 30s to maybe 60? (It's in `test/Spago.purs`)

I'm pretty sure it's node-execa that's at fault. In the latest release where I updated it to the new node libraries, I came across this issue thanks to Verity: https://github.com/JordanMartinez/purescript-node-execa/commit/fda206466d694fd3bb042b0f6cb758a0aea9adae

That version of the dependency used here might still have that code.

f-f:
> I'm pretty sure it's node-execa that's at fault

I would try to extend the timeout first - tests are passing on my mac.

If it turns out it's not that then we can consider upgrading to the latest set (probably in a separate PR)

JordanMartinez:
> > I'm pretty sure it's node-execa that's at fault
> 
> I would try to extend the timeout first - tests are passing on my mac.
> 
> If it turns out it's not that then we can consider upgrading to the latest set (probably in a separate PR)

Oh right. Tests pass on Linux, so it can't be that...

f-f (approved):
Great, thanks!

PR #1009, opened 2023-09-21
Properly propagate `--strict` flag

### LINKED ISSUE #982: `psa` related options don't work yet
Here's a concrete example that shows that the `--strict` flag does not work yet:

```
$spago --version
0.93.9
$psa --version
0.8.2
$spago init
Initializing a new project...
Found PureScript 0.15.10, will use package set 32.3.0
Set up a new Spago project.
Try running `spago run`
```

*src/Main.purs:*
```
module Main where

import Prelude

import Effect (Effect)
import Effect.Console (log)

main :: Effect Unit
main = do
  let main = 1
  log "🍝"
```

```
spago build --strict

[...]
           Src   Lib   All
Warnings     2     0     2
Errors       0     0     0

✅ Build succeeded.
```

```
$ls .spago/
BuildInfo.purs	packages
```

JordanMartinez:
### Description of the change

Fixes #982

### Checklist:

### DIFF
--- a/bin/src/Flags.purs
+++ b/bin/src/Flags.purs
@@ -19,8 +19,8 @@ strict ∷ ArgParser (Maybe Boolean)
 strict =
   ArgParser.flag [ "--strict" ]
     "Promotes project sources' warnings to errors"
-    # ArgParser.boolean
-    # ArgParser.optional
+    $> Just true
+    # ArgParser.default Nothing
 
 censorBuildWarnings ∷ ArgParser (Maybe Core.CensorBuildWarnings)
 censorBuildWarnings =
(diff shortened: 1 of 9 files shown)


bin/src/Flags.purs
```diff
@@ -19,8 +19,8 @@ strict ∷ ArgParser (Maybe Boolean)
 strict =
   ArgParser.flag [ "--strict" ]
     "Promotes project sources' warnings to errors"
-    # ArgParser.boolean
-    # ArgParser.optional
```
JordanMartinez, line 23: This always returns `Just false` when the flag is not provided because the combo of `boolean` and then `optional` is:
```
default Nothing $ map Just $ default false $ true <$ flag
-- which simplifies to
map Just $ default false $ true <$ flag
-- which simplifies to
default (Just false) $ Just true <$ flag
```

Thus, the `Just` value would always override the workspace one.

f-f (approved):
This is great, thank you!

PR #1011, opened 2023-09-22
Add --test-deps flag to install

### LINKED ISSUE #983: `spago install --test-deps`
spago@next configured with `pedantic_packages: true` errors when a package imports a transitive dependency.
In a normal package it logs the suggestion like:

`spago install -p my-package console`

Running the command works. However, if the same happens with a missing dependency in the tests of a package, the following command is suggested:
`spago install --test-deps -p my-package console`

This command seems to be not implemented yet.

JordanMartinez:
### Description of the change

Fixes #983.

### Checklist:

### DIFF
--- a/src/Spago/Config.js
+++ b/src/Spago/Config.js
@@ -9,6 +9,13 @@ export function addPackagesToConfigImpl(doc, isTest, newPkgs) {
     ? doc.get("package").get("test").get("dependencies")
     : doc.get("package").get("dependencies");
 
+  // Stringify this collection as 
+  //  - dep1 
+  //  - dep2
+  // rather than
+  //  [ dep1, dep2 ]
+  deps.flow = false;
+
   // Gather all deps, old and new, in a new set
   let depsSet = new Set(deps.toJSON());
   for (const pkg of newPkgs) {
--- a/src/Spago/Config.js
+++ b/src/Spago/Config.js
@@ -4,8 +4,10 @@ export function updatePackageSetHashInConfigImpl(doc, sha) {
   doc.get("workspace").get("package_set").set("hash", sha);
 }
 
-export function addPackagesToConfigImpl(doc, newPkgs) {
-  const deps = doc.get("package").get("dependencies");
+export function addPackagesToConfigImpl(doc, isTest, newPkgs) {
+  const deps = isTest
+    ? doc.get("package").get("test").get("dependencies")
+    : doc.get("package").get("dependencies");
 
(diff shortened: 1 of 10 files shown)


test-fixtures/spago-install-test-deps-success.yaml
```diff
@@ -0,0 +1,13 @@
+package:
+  dependencies:
+    - console
+    - effect
+    - prelude
+  name: aaa
+  test:
+    dependencies: [ foreign ]
```
JordanMartinez, line 8: I'm not sure how to get this to format better, but presumably because the original content was `[]`, that's why we get `[ foreign ]` rather than
```
dependencies:
  - foreign
```

Thoughts on how to fix that?

f-f: I had no idea so I checked in the node repl:
```node
> yaml = require('yaml');
> let d1 = yaml.parseDocument("- foreign")
> let d2 = yaml.parseDocument("[foreign]")
> d1.contents
YAMLSeq {
  items: [
    Scalar {
      value: 'foreign',
      range: [Array],
      source: 'foreign',
      type: 'PLAIN'
    }
  ],
  range: [ 0, 9, 9 ]
}
> d2.contents
YAMLSeq {
  items: [
    Scalar {
      value: 'foreign',
      range: [Array],
      source: 'foreign',
      type: 'PLAIN'
    }
  ],
  flow: true,
  range: [ 0, 9, 9 ]
}
```

It looks like should set the [flow style](https://www.yaml.info/learn/flowstyle.html) to false whenever we modify collections

src/Spago/Config.js
```diff
@@ -4,8 +4,10 @@ export function updatePackageSetHashInConfigImpl(doc, sha) {
   doc.get("workspace").get("package_set").set("hash", sha);
 }
 
-export function addPackagesToConfigImpl(doc, newPkgs) {
-  const deps = doc.get("package").get("dependencies");
+export function addPackagesToConfigImpl(doc, isTest, newPkgs) {
+  const deps = isTest
+    ? doc.get("package").get("test").get("dependencies")
```
f-f, line 9: The `test` section is optional, so we should also have a test where the package does not have one, and see what happens (we'll presumably have to fix this implementation)

f-f (approved):
This is great, thank you!

JordanMartinez:
Did you forget to merge this? Or was it something else?

f-f:
I left it open so you could merge, but I can also merge it of course 😄

PR #1012, opened 2023-09-22
Run all tests, not just find flags

JordanMartinez:
### Description of the change

TODO

### Checklist:

### DIFF
--- a/test/Prelude.purs
+++ b/test/Prelude.purs
@@ -86,7 +86,7 @@ checkResultAndOutputsStr maybeOutStr maybeErrStr resultFn execResult = do
       Left err -> err.stderr
       Right res -> res.stderr
 
-  when false do
+  when true do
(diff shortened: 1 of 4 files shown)


JordanMartinez:
The `run > works at all` test keeps timing out in various PRs. I'm wondering what's going on there...

f-f:
I wonder if we should update to the latest package set since there might be execa or node bugs in this one?

JordanMartinez:
Quite possibly. But that requires updating the Registry as well. I already started looking down this path.

f-f:
We are already pointing to the latest commit of the registry-dev repo if that's what you are referring to

f-f:
I hate this windows build, rerunning..

f-f:
@JordanMartinez is this PR still relevant?

PR #1013, opened 2023-09-22
Allow git branches with file-system-invalid chars

### LINKED ISSUE #1004: Make sure that "interesting" branch names are escaped
When cloning git repos we put them in the `.spago` folder, with the ref being part of the path. This is all fine, except that branch names can contain slashes, with would mess up the path.

[spago-legacy](https://github.com/purescript/spago-legacy/blob/11e68ece50ede286660452ef370baa3ab5c33c02/test/UnitSpec.hs#L96) had tests for this, and we [have a commented out test](https://github.com/purescript/spago/blob/32d89b515e656a57ee78552b87f31cf93d8decdc/test/Spago/Install.purs#L104) (even though we should have the QuickCheck ones too) that should check for this.

The fix should be something along the lines [of the old implementation that we had](https://github.com/purescript/spago-legacy/blob/11e68ece50ede286660452ef370baa3ab5c33c02/src/Spago/FetchPackage.hs#L219)

JordanMartinez:
### Description of the change

Fixes #1004

### Checklist:

### DIFF
--- a/src/Spago/Config.purs
+++ b/src/Spago/Config.purs
@@ -25,17 +25,15 @@ import Affjax.StatusCode (StatusCode(..))
 import Data.Array as Array
 import Data.CodePoint.Unicode as Unicode
 import Data.Codec.Argonaut.Record as CAR
-import Data.Enum (fromEnum)
+import Data.Enum as Enum
 import Data.Foldable as Foldable
 import Data.HTTP.Method as Method
-import Data.Int (hexadecimal)
 import Data.Int as Int
 import Data.Map as Map
 import Data.Profunctor as Profunctor
 import Data.Set as Set
 import Data.Set.NonEmpty (NonEmptySet)
-import Data.String (CodePoint, Pattern(..), codePointFromChar)
-import Data.String as SCP
+import Data.String (CodePoint, Pattern(..))
 import Data.String as String
 import Dodo as Log
 import Effect.Uncurried (EffectFn2, runEffectFn2)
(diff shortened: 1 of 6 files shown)


test/Spago/Install.purs
```diff
@@ -104,31 +104,32 @@ spec = Spec.around withTempDir do
...
+                            , ref: "spago-test/branch-with-slash"
+                            , subdir: Nothing
+                            , dependencies: Just $ Dependencies $ Map.singleton (mkPackageName "prelude") Nothing
+                            }
+                        ]
+                    }
+                )
+            }
+        )
+      spago [ "install", "nonexistent-package" ] >>= shouldBeSuccess
```
JordanMartinez, line 132: I'm not sure what `installs-with-slash.txt` is supposed to say, but if this succeeds, is that good enough?

src/Spago/Config.purs
```diff
@@ -23,14 +23,19 @@ import Affjax.Node as Http
...
 import Data.HTTP.Method as Method
+import Data.Int (hexadecimal)
+import Data.Int as Int
 import Data.Map as Map
 import Data.Profunctor as Profunctor
 import Data.Set as Set
 import Data.Set.NonEmpty (NonEmptySet)
-import Data.String (Pattern(..))
+import Data.String (CodePoint, Pattern(..), codePointFromChar)
+import Data.String as SCP
```
f-f, line 38: We already import this as `String`

src/Spago/Config.purs
```diff
@@ -23,14 +23,19 @@ import Affjax.Node as Http
 import Affjax.ResponseFormat as Response
 import Affjax.StatusCode (StatusCode(..))
 import Data.Array as Array
+import Data.CodePoint.Unicode as Unicode
 import Data.Codec.Argonaut.Record as CAR
+import Data.Enum (fromEnum)
 import Data.Foldable as Foldable
 import Data.HTTP.Method as Method
+import Data.Int (hexadecimal)
```
f-f, line 31: Could we qualify all of these function imports? We generally import directly only types, everything else is qualified.

test/Spago/Install.purs
f-f: I think that's fine, but we should also look that the folder in the `.spago` folder has the name that we expect it to have

f-f:
The issue also mentions Quick check tests that check for the injectivity of the naming function, though I don't have a link at hand

JordanMartinez:
> The issue also mentions Quick check tests that check for the injectivity of the naming function, though I don't have a link at hand

Looks like this one? https://github.com/purescript/spago-legacy/blob/master/test/UnitSpec.hs

f-f:
Ah yes! It's this one. Not sure if it's worth porting since the implementation has been stable for years, so we don't have to necessarily have it in the PR

JordanMartinez:
> Ah yes! It's this one. Not sure if it's worth porting since the implementation has been stable for years, so we don't have to necessarily have it in the PR

I think it's worth porting.

f-f:
@JordanMartinez I think this one is good to go once the conflicts are sorted out

JordanMartinez:
Can this get an approval?

f-f (approved):
Thanks!

PR #1017, opened 2023-09-23
Do not ask for JS compilation if compiling with a backend

f-f:
This might need a test, but still not sure how it would look like


JordanMartinez:
Couldn't you test that a given module only has its CoreFn output and not the JS output if you specify a non-existant backend? Or does that backend arg get used at some point?

JordanMartinez (approved):

Issue #988, opened 2023-09-04
Make psa options (e.g. `pedantic`, `strict`) options for each package, not for the entire workspace


JordanMartinez:
In [`language-purescript`](https://github.com/JordanMartinez/purescript-language-purescript), I've defined two scripts that are helpful for that project. One is a single file whereas another has multiple files that encompass the script.

For the main library, I want to enable the `pedantic` and `strict` options. But for the scripts themselves, I don't want these enabled because they interfere with my development. However, these options can only be enabled/disabled globally on the entire workspace rather than each package individually. As a result, these scripts, which may have more imports than needed or produce warnings, cause CI to fail despite such failures being irrelevant.

Rather than defining those two options on the workspace, can they be defined on each package individually?

**EDIT**: Current status of this issue is here: https://github.com/purescript/spago/issues/988#issuecomment-1734388070

f-f:
It sounds like we'd be better off moving these options (and presumably all the psa options as well) to be a package property, rather than a workspace property - that's all good by me, as long as we don't end up with both (which is why we only have the workspace ones at the moment)

It might be a good occasion to write some tests for this psa integration, given that it doesn't seem to trigger sometimes, see #982

JordanMartinez:
If we move those into the package config, then there's some interplay there. For single-package repo, this change makes no difference. For multi-package repo, some interplay arises. These cases are described below.

```mermaid
flowchart TD
  subgraph "Case 1"
    A ---> DepsA
    B ---> DepsB
  end
  
  subgraph "Case 2"
    A2 ---> Deps1
    B2 ---> Deps1
  end

  subgraph "Case 3"
    A3 ---> Deps3
    A3 ---> B3
    B3 ---> Deps3
  end
```

In case 1, A and B both depend on different dependencies (e.g. none for A and `prelude` for B). In this case, censoring lib warnings works no differently than it does now.

In case 2, censoring lib warnings may need to work a bit differently. If A2's configuration is to censor `Deps1` warnings whereas B2 does not, what happens when one runs `spago build`, which builds both packages? In this conflict, do we censor or do we not? I assume we do not.

In case 3, if B3 has `censor-warnings` enabled but `A3` does not censor its library dependencies' warnings, and the user runs `spago build`, do we censor B3's warnings or not?

f-f:
[...]

PR #1018, opened 2023-09-26
Build workspace packages in top-sorted order

JordanMartinez:
### Description of the change

First part of #988.

- When building all workspace packages, builds them in a reverse-topological-sort order (i.e. leaves first, then roots)
- Adds a duplicate module check when building all workspace packages

Tests added cover these cases:

```mermaid
flowchart TD
  subgraph "Case 1"
    A ---> Dep0["effect, console, prelude"]
    B ---> Dep0
  end
  
  subgraph "Case 2"
    A2 ---> effect2
    A2 ---> console2
    A2 ---> prelude2
    A2 ---> Shared2
    B2 ---> effect2
    B2 ---> console2
    B2 ---> prelude2
    B2 ---> Shared2
    Shared2 ---> prelude2
  end

  subgraph "Case 3"
    A3 ---> effect3
    A3 ---> console3
    A3 ---> prelude3
    A3 ---> B3
    A3 ---> C3
    B3 ---> effect3
    B3 ---> console3
    B3 ---> prelude3
    B3 ---> C3
    C3 ---> prelude3
  end

  subgraph "Case 4 (duplicate module)"
    A4 ---> effect4
    A4 ---> console4
    A4 ---> prelude4
    B4 ---> effect4
    B4 ---> console4
    B4 ---> prelude4
    C4 ---> prelude4
  end
```

### Checklist:


bin/src/Main.purs
```diff
@@ -819,10 +819,9 @@ mkReplEnv replArgs dependencies = do
 
   purs <- Purs.getPurs
 
-  let
-    selected = case workspace.selected of
-      Just s -> [ s ]
-      Nothing -> Config.getWorkspacePackages workspace.packageSet
+  selected <- case workspace.selected of
+    Just s -> pure s
+    Nothing -> die "The REPL requires you to select a package"
```
JordanMartinez, line 824: Here's one potential downside of this change... But I guess this is just making what was an implicit problem explicit...?

f-f: Do we have to change this?

JordanMartinez: What happens when there are two modules with the same package name? I guess the only option would be to force the user to specify a package that exists only for the repl?

f-f: You mean two packages with the same module name? We shouldn't allow that - we'll probably need to run a graph check when compiling more than one package so we can detect this.

JordanMartinez:
The three cases I test for are similar to but not quite those described in the #988. Unless we publish a package that intentionally has a warning in it, it will not be possible to test for those cases specifically. So, the best we can do is test against local workspace packages that satisfy that criteria.

src/Spago/Command/Build.purs
```diff
@@ -171,11 +179,48 @@ getBuildGlobs { selected, dependencies, withTests, depsOnly } =
...
+    $ Array.fromFoldable
+    $ Map.keys
+    $ unwrap
+    $ selected.package.dependencies
+
+  -- depsOnly means "no packages from the monorepo", so we filter out the workspace packages
+  dependencyGlobs = (Tuple.uncurry $ Config.sourceGlob NoTestGlobs) =<< dependencyPkgs
+  monorepoPkgGlobs
+    | depsOnly = []
+    | otherwise = (Tuple.uncurry $ Config.sourceGlob NoTestGlobs) =<< monorepoPkgs
```
JordanMartinez, line 206: CI is failing because of the changes I made here. The underlying issue is that the globs don't include the dependencies of the package's direct dependencies. I'm not as familiar with this part of the code, so @f-f could you give some insight as to how I should best handle this?

I basically want to write
```
Set.fromFoldable $ projectGlobs <> dependencyGlobs <> [ BuildInfo.buildInfoPath ]
```
but it seems that `dependencies` includes the dependencies for all workspace packages, not just the one for which I'm currently getting globs.

bin/src/Main.purs
JordanMartinez: I've updated the code to reallow this, but haven't yet implemented the duplicate module check yet.

src/Spago/Command/Build.purs
f-f: Right, we collect all the dependencies needed in the `Spago.Command.Fetch` module - we could collect a different set of dependencies for each package there...

...however this is problematic as well: if you'd like package A to hide warnings from its dependencies, and package B to definitely show you warnings for the dependencies, then building package A first will mean that Spago will build its dependencies, hide the warnings, and keep a cache of this, so when building package B no warnings will be shown either.

A possible way out of this could be to treat the external dependencies of the project as a whole (i.e. as things are right now) - this would mean that we'd end up configuring each package PSA options separately (i.e. strict, hiding warnings, etc, but only for the code in the project), but configuring the behaviour of how to treat the warnings from dependencies at the workspace level instead.

JordanMartinez: That sounds fine to me. Most people probably don't care about the library code anyways since it's beyond their control.

JordanMartinez:
Tasks remaining:
- [ ] Check for duplicate modules in the workspace packages and error if found
- [ ] Add a test for the above check

src/Spago/Command/Fetch.purs
```diff
@@ -105,12 +137,23 @@ run { packages, ensureRanges, isTest } = do
...
   when ensureRanges do
-    { configPath, yamlDoc } <- getPackageConfigPath
+    { configPath, yamlDoc, transitiveDeps } <- case workspaceDepsOrSelectedPackage of
+      Right r ->
+        pure r
+      Left _ ->
+        die
+          [ "No package found in the root configuration."
+          , "Please use the `-p` flag to select a package to which to add ranges."
+          ]
```
JordanMartinez, line 147: Is there a util for something like
```
foo `orDieWith`
  [ "error message" ]
```
where `orDieWith` is defined as
```purs
orDieWith :: forall a msg y. Maybe a -> msg -> Spago y a
orDieWith mb msg = case mb of
  Just a -> pure a
  Nothing -> die msg
```

src/Spago/Command/Publish.purs
```diff
@@ -105,8 +106,9 @@ publish _args = do
     , logOptions: env.logOptions
     , git: env.git
     , purs: env.purs
-    , selected: env.selected
+    , selected: env.selected -- Note: should this be using the updated `selected` value?
```
JordanMartinez, line 109: Just highlighting this question...

JordanMartinez:
@f-f For this code https://github.com/purescript/spago/blob/32614da384cc10e147ec658146dd59b16769d246/src/Spago/Purs/Graph.purs#L54-L56
could you clarify what the mappings are?

f-f:
Sure - let's add a comment to clarify it for future reference: for every Package that we depend on, we note which Module we are depending on, and for each of them we note from which Module we are importing it.

JordanMartinez:
K this should be good to go. Opening thread has been updated to include a visualization of the 4 cases being tested.

f-f:
Thanks! I'll have a look through it now

src/Spago/Purs/Graph.purs
```diff
@@ -149,32 +165,39 @@ checkImports = do
...
-          (if Set.isEmpty unused then [] else [ unusedError false selected unused ])
-            <> (if Map.isEmpty transitive then [] else [ transitiveError false selected transitive ])
-            <> (if Set.isEmpty unusedTest then [] else [ unusedError true selected unusedTest ])
-            <> (if Map.isEmpty transitiveTest then [] else [ transitiveError true selected transitiveTest ])
+      pure Nothing
+    Right graph ->
+      pure $ Just graph
 
-      pure result
+runGraphCheck :: forall a. WorkspacePackage -> Set FilePath -> Array String -> Spago (PreGraphEnv a) (Array Docc)
```
f-f, line 192: We don't need this function if we are exposing `runGraph` and `toImportErrors`

src/Spago/Command/Build.purs
```diff
@@ -140,42 +97,190 @@ run opts = do
...
+        Nothing ->
+          logWarn $
+            "Due to JSON decoding failure on 'purs graph' output, \
+            \packages that define modules with the same name might not be detected."
+        Just graph -> do
+          let
+            toModulesDefinedByThisPackage :: Purs.ModuleGraphNode -> Maybe (Array (Tuple PackageName FilePath))
+            toModulesDefinedByThisPackage v =
+              [ Tuple selected.package.name v.path ] <$ (String.stripPrefix (String.Pattern selected.path) v.path)
+            modulesDefinedByThisPackage = Map.mapMaybe toModulesDefinedByThisPackage $ unwrap graph
```
f-f, line 189: I was entirely convinced that `purs graph` would fail with `DuplicateModule` in case of modules with the same name, but no, it just takes the last! Or at least I can't get it to fail, maybe I'm not trying hard enough.

f-f: If we indeed have to do all this bookkeeping, then it should be moved to the Graph module, so we can have a nicely contained `detectDuplicateModules`, or something like that

f-f: An alternative to doing the bookkeeping is of course to just build everything all together and let it fail with `DuplicateModule` from `purs`

f-f (commented):
Had a quick pass, I'll do another later/tomorrow

src/Spago/Command/Fetch.purs
```diff
@@ -2,10 +2,12 @@ module Spago.Command.Fetch
   ( FetchEnv
   , FetchEnvRow
   , FetchOpts
+  , PackageDependencyInfo
```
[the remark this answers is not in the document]
JordanMartinez: AFAIK, the behavior is unchanged. There are only 2 possible states that can occur when producing the `transitiveDeps` value here:
1. a package was selected (whether it was the root package or some other package is irrelevant). In which case, we get dependencies only for that package.
2. no package was selected. In which case, we get dependencies for all packages.

But, I updated this code so that I am also recording which package corresponds with which `PackageMap` value when calling `getTransitiveDeps`. Otherwise, the `Build.getBuildGlobs`' `dependencies` argument includes a given package's dependencies plus the other workspace package dependencies. Using `selected.package.dependencies` in `Build.getBuildGlobs` returns only the direct dependencies of that package, not the transitive ones.

src/Spago/Purs/Graph.purs
```diff
@@ -51,8 +55,20 @@ type PackageGraphNode =
...
+-- | ```
+-- | module MyModule
+-- | 
+-- | import Prelude -- from the 'prelude' package
+-- | ```
+-- | This value would be
+-- | `Map.singleton "prelude" $ Map.singleton "Prelude" $ Set.singleton "MyModule"``
+-- |
+-- | Every package can be imported by several project modules.
+-- | In each project module, several modules from the same package can be imported.
```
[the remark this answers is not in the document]
JordanMartinez: I think I'll just drop the tail.

src/Spago/Purs/Graph.purs
JordanMartinez: Oh ok. I wasn't sure if the API here was used somewhere downstream as `checkImports` was being exported but isn't used anywhere else AFAIK.

src/Spago/Command/Build.purs
JordanMartinez: > I was entirely convinced that `purs graph` would fail with `DuplicateModule` in case of modules with the same name, but no, it just takes the last! Or at least I can't get it to fail, maybe I'm not trying hard enough.

I think you're running into https://github.com/purescript/purescript/issues/4342

f-f: Oh no, what a nightmare. Do you think we should fix that now, or paper over it here?

src/Spago/Purs/Graph.purs
JordanMartinez: Done.

src/Spago/Command/Build.purs
JordanMartinez: It's always better to fix it up-stream, but the problem is that `purs graph` doesn't work until the latest release. So, we'll still need to paper over it here.

src/Spago/Command/Fetch.purs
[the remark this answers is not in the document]
f-f: I see - then I think it's hard for me to see the need for both fields in this record:
- if we have a map of dependencies for each package, then we can just take a union of all the values when we need to have all the dependencies, that sounds like we don't need to keep the `dependencies` field around?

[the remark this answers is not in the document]
JordanMartinez: I originally kept it because it led to less code churn due to the renaming of the field and its corresponding type while I worked on other things. Moreover, calculating `dependencies` here once meant I didn't need to recalculate it elsewhere in every usage thereafter. But, on the other hand, that means two pieces of state that could diverge at some point. So, I've dropped the `dependencies` field.

JordanMartinez:
Another idea came to mind this morning, but if we do implement it, I'd like to do so in a separate PR. Let's say I have three packages A, B, C, and D with the following dependency structure: C and D both depend on A and B whereas A and B are independent of one another:

```mermaid
flowchart TD
  C ---> A
  C ---> B
  D ---> A
  D ---> B
```
Let's assume that `C` will fail to build whereas `D` will succeed. If I run `spago build`, I would assume that the command will build as many packages as can be built. In other words, let's say we build `C` before `D`. Even though `C` fails to build, `D` should still be built. Right now, if `C` fails to build, I believe `D` won't even be attempted.

f-f:
> Even though `C` fails to build, `D` should still be built. Right now, if `C` fails to build, I believe `D` won't even be attempted.

Does it matter? The build is broken and needs fixing 🙂 

In all seriousness - we could attempt all the builds and give all the errors together, but not sure it would be very useful? It's often the case that one needs to go through the fix-compile-repeat loop many times as the compiler errors out as soon as it finds errors in the modules it's building and can't build their dependants.

JordanMartinez:
Good points. The user could also call `spago build -p` manually for the packages that did build, so this doesn't need to be implemented.

src/Spago/Config.purs
```diff
@@ -453,6 +455,19 @@ getWorkspacePackages = Array.mapMaybe extractWorkspacePackage <<< Map.toUnfoldab
     Tuple _ (WorkspacePackage p) -> Just p
     _ -> Nothing
 
+getToplogicallySortedWorkspacePackages :: PackageSet -> Array WorkspacePackage
+getToplogicallySortedWorkspacePackages packageSet = do
```
f-f, line 459:
```suggestion
getTopologicallySortedWorkspacePackages :: PackageSet -> Array WorkspacePackage
getTopologicallySortedWorkspacePackages packageSet = do
```

src/Spago/Command/Run.purs
```diff
@@ -97,11 +97,11 @@ run = do
       -- We check that the module we're about to run is included in the build and spit out a nice error if it isn't (see #383)
       let
         globs = Build.getBuildGlobs
-          { dependencies
+          { dependencies: foldl (Map.unionWith (\l _ -> l)) Map.empty packageDependencies
```
f-f, line 100: `Fetch.getAllDependencies` here too

src/Spago/Command/Publish.purs
f-f: Actually yeah, let's use the new value

src/Spago/Command/Build.purs
```diff
@@ -140,42 +96,193 @@ run opts = do
...
+    | otherwise = (Tuple.uncurry $ Config.sourceGlob NoTestGlobs) =<< monorepoPkgs
+
+type BuildWorkspaceGlobsOptions =
+  { withTests :: Boolean
+  , depsOnly :: Boolean
+  , workspacePackages :: Array WorkspacePackage
+  , dependencies :: Map PackageName Package
+  }
+
+getEntireWorkspaceGlobs :: BuildWorkspaceGlobsOptions -> Set FilePath
```
f-f, line 277: Are these implementations different enough to warrant a split? It feels like we could cover everything with a few datatypes (I mean, the previous implementation was covering everything, so why not) - adding a different one means they can eventually diverge, which I'd rather not do

f-f (commented):
Had another round of reviewing - I don't seem to be able to wrap my head around the patch in one go, but we'll get there 😄

f-f:
Ah, I also tested it and it's pleasantly fast - I was a bit concerned that it would be slower to call purs so many times, but it's actually alright.

Something else that I noticed was that we show the PSA results for every package (and the `Build succeeded` message too) - I wonder if we should gather all the things in a single PSA display?

image

src/Spago/Command/Init.purs
```diff
@@ -84,42 +88,51 @@ run opts = do
...
-  , workspace: case withWorkspace of
-      false -> Nothing
-      true -> Just
-        { extra_packages: Just Map.empty
-        , package_set: setVersion # map \set -> SetFromRegistry { registry: set }
-        , build_opts: Nothing
-        , backend: Nothing
-        , lock: Nothing
-        }
+  , workspace: (These.theseRight opts) <#> \{ setVersion } ->
```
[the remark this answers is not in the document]
JordanMartinez: Well, a config file is one of three values: 
- A package only 
- A workspace only 
- Both of them 

These summarizes this idea without allowing for am impossible state like two Nothings if it's represented using two Maybes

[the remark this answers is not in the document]
JordanMartinez: Since each field here (i.e. package and workspace) are maybe values, I'm pulling out the corresponding value from the These value (if it exists).

src/Spago/Command/Fetch.purs
```diff
@@ -431,3 +459,21 @@ getVersionFromPackage :: Package -> Version
 getVersionFromPackage = case _ of
   RegistryVersion v -> v
   _ -> unsafeFromRight $ Version.parse "0.0.0"
+
+-- TODO I guess this should be configurable
+replSupportPackageName :: PackageName
```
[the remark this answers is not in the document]
JordanMartinez: Should i put it in a module like Repl.SupportPackage?

src/Spago/Command/Init.purs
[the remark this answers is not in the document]
f-f: Oh I see now! I think we might be better off with a custom type that has three possible values - it's equivalent to `These`, but more self-explanatory and less surprising. We might want to do the same with the new `Either`s that are being added too

src/Spago/Command/Fetch.purs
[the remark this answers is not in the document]
f-f: I guess we could have a `Spago.Repl` module with it

JordanMartinez:
Latest commits address some of the feedback, but I'll have to get to the rest of this on Monday.

JordanMartinez:
> Something else that I noticed was that we show the PSA results for every package (and the `Build succeeded` message too) - I wonder if we should gather all the things in a single PSA display?
> image

I think this should be further discussed before we work on it. At the very least, I don't plan on doing anything for this in this PR.

JordanMartinez:
I believe I've addressed all feedback. I'm still running tests locally to verify that everything still passes.

bin/src/Main.purs
```diff
@@ -515,7 +516,7 @@ main =
           Install args@{ packages, selectedPackage, ensureRanges, testDeps } -> do
             { env, fetchOpts } <- mkFetchEnv { packages, selectedPackage, ensureRanges, testDeps }
             -- TODO: --no-fetch flag
-            dependencies <- runSpago env (Fetch.run fetchOpts)
+            packageDependencies <- runSpago env (Fetch.run fetchOpts)
```
f-f, line 519:
```suggestion
            dependencies <- runSpago env (Fetch.run fetchOpts)
```

This is a bit of a nit, but after these few iterations I think it would make sense to go back to calling this `dependencies`, mostly for terseness, but also so that we can use `packageDependencies` for when we extract the set of dependencies from a package

bin/src/Main.purs
```diff
@@ -844,23 +845,23 @@ mkPublishEnv dependencies = do
...
 
   purs <- Purs.getPurs
 
-  let
-    selected = case workspace.selected of
-      Just s -> [ s ]
-      Nothing -> Config.getWorkspacePackages workspace.packageSet
+  selected <- case workspace.selected of
+    Just s -> pure $ Build.SinglePackageGlobs s
+    Nothing -> pure $ Build.AllWorkspaceGlobs $ Config.getWorkspacePackages workspace.packageSet
```
f-f, line 859: I think `AllWorkspaceGlobs` always implies that we `getWorkspacePackages`, so we should call this then we go use the value rather than when we construct it (or use a smart constructor)
```suggestion
    Nothing -> pure $ Build.AllWorkspaceGlobs workspace.packageSet
```

src/Spago/Command/Publish.purs
```diff
@@ -125,9 +126,12 @@ publish _args = do
     )
 
   -- We then need to check that the dependency graph is accurate. If not, queue the errors
-  let globs = Build.getBuildGlobs { selected: [ selected ], withTests: false, dependencies, depsOnly: false }
-  graphCheckErrors <- Graph.runGraphCheck selected globs []
-  for_ graphCheckErrors addError
+  let dependencies = Fetch.getAllDependencies packageDependencies
```
f-f, line 129: Looking at this diff reminds me that we need a dry-run kind of test for the publish command, there is nothing that ensures that it keeps working at the moment 😅

src/Spago/Command/Fetch.purs
```diff
@@ -250,7 +297,10 @@ run { packages, ensureRanges, isTest } = do
       LocalPackage _ -> pure unit
       WorkspacePackage _ -> pure unit
 
-  pure transitivePackages
+  pure packageDependencies
+
+getAllDependencies :: Map PackageName PackageMap -> PackageMap
```
f-f, line 302: Uhhh I think I suggested this name, but `Fetch.getAllDependencies` has a "download things" sort of vibe. Not sure what a better name would be though, maybe just `allDependencies`? 🤔 

Then we could have a type alias `type AllDependencies = Map PackageName PackageMap` to make it extra clear

src/Spago/Command/Fetch.purs
```diff
@@ -33,18 +34,28 @@ import Registry.Sha256 as Sha256
...
 
+data PackageSelection
+  = SinglePackage
+      { workspacePackage :: WorkspacePackage
+      , configPath :: FilePath
+      , yamlDoc :: YamlDoc Core.Config
+      , transitiveDeps :: PackageMap
+      }
+  | AllWorkspacePackages (Map PackageName PackageMap)
+
```
f-f, line 58: There's a duality between this type above and this type defined in the `Build` module:

```
data SelectedPackageGlob
  = SinglePackageGlobs WorkspacePackage
  | AllWorkspaceGlobs (Array WorkspacePackage)
```

I wonder if it's worth collapsing everything into this one so we avoid the wrapping/unwrapping, something like
```suggestion
data PackageSelection
  = SinglePackage
      { workspacePackage :: WorkspacePackage
      , configPath :: FilePath
      , yamlDoc :: YamlDoc Core.Config
      , transitiveDeps :: PackageMap
      }
  | AllWorkspacePackages (Map WorkspacePackage PackageMap)

```

src/Spago/Command/Fetch.purs
```diff
@@ -109,12 +140,28 @@ run { packages, ensureRanges, isTest } = do
 
   -- if the flag is selected, we kick off the process of adding ranges to the config
   when ensureRanges do
-    { configPath, yamlDoc } <- getPackageConfigPath
+    { configPath, yamlDoc, transitiveDeps } <- case selectedPackage of
+      SinglePackage p -> pure p
+      AllWorkspacePackages _ -> die
+        [ "No package found in the root configuration."
+        , "Please use the `-p` flag to select a package to which to add ranges."
+        ]
```
f-f, line 148: This error message was packaged in a single function and is now repeated three times, could we reintroduce that?

src/Spago/Command/Build.purs
```diff
@@ -140,56 +96,188 @@ run opts = do
...
+                              $ map
+                                  ( \(Tuple pkg moduleFilePath) ->
+                                      Dodo.text $ "- " <> PackageName.print pkg <> "   at path: " <> moduleFilePath
+                                  )
+                              $ Array.sort ps
+                          ]
+                    )
+                # Dodo.lines
+            ] :: Docc
+        )
```
f-f, line 233: Let's extract this formatting stuff to a separate function

src/Spago/Command/Build.purs
```diff
@@ -140,56 +96,188 @@ run opts = do
...
-      map workspacePackageGlob selected
-
   testGlobs = case withTests of
     true -> WithTestGlobs
     false -> NoTestGlobs
 
   workspacePackageGlob :: WorkspacePackage -> Array String
   workspacePackageGlob p = Config.sourceGlob testGlobs p.package.name (WorkspacePackage p)
 
+  { projectGlobs, monorepoTestGlobs } = case depsOnly of
```
f-f, line 259: There was a comment detailing the reason why the depsOnly works that way, let's reintroduce it here

test/Spago/Build/Polyrepo.purs
```diff
@@ -0,0 +1,349 @@
...
+
+  let
+    spagoInitCleanupNonPackageFiles spago = do
+      spago [ "init" ] >>= shouldBeSuccess
+      FSA.unlink $ Path.concat [ "src", "Main.purs" ]
+      FSA.unlink $ Path.concat [ "test", "Test", "Main.purs" ]
+      FSA.unlink "spago.yaml"
+      FS.writeYamlFile Config.configCodec "spago.yaml"
+        $ Init.defaultConfig'
+        $ WorkspaceOnly { setVersion: Just $ unsafeFromRight $ Version.parse "0.0.1" }
```
f-f, line 30: I think this could just be this?

```suggestion
    spagoInitConfig spago = do
      FS.writeYamlFile Config.configCodec "spago.yaml"
        $ Init.defaultConfig'
        $ WorkspaceOnly { setVersion: Just $ unsafeFromRight $ Version.parse "0.0.1" }
```

test/Spago/Build/Polyrepo.purs
```diff
@@ -0,0 +1,349 @@
...
+import Node.Path as Path
+import Registry.Version as Version
+import Spago.Command.Init (DefaultConfigOptions(..))
+import Spago.Command.Init as Init
+import Spago.Core.Config as Config
+import Spago.FS as FS
+import Test.Spec (SpecT)
+import Test.Spec as Spec
+import Test.Spec.Assertions as Assert
+
```
f-f, line 18: Might be worth linking this PR here in this module, so that visitors from the future will be able to find the diagrams in the PR description, as understanding the various cases with those is easier

f-f (commented):
Finally managed to go through the whole thing. Good work @JordanMartinez! I left a few small notes but it's otherwise good to go

f-f:
> I think this should be further discussed before we work on it. At the very least, I don't plan on doing anything for this in this PR.

Ah yes about this: agreed that we can shrink the PSA display in a future PR, but I'd like to address the multiple `Build Succeeded` messages in here, as they don't really make sense right now

src/Spago/Command/Fetch.purs
JordanMartinez: Originally... In `Fetch`, the additional fields (i.e. `configPath` and `yamlDoc`) were added so we wouldn't need to construct them twice / do `case _ of` twice part when running the code under `when ensureRanges` and `when (not $ Array.null packages)`. In `Build`, this extra info isn't needed and may be hard to construct.

So, you might say, let's define a type:
```purs
data PackageSelection a b
  = SinglePackage a
  | AllWorkspacePackges b
```
the works across these two. I guess that works; this just feels super nitpicky to me.

src/Spago/Command/Fetch.purs
JordanMartinez: Personally, I hate type aliases because they disappear in errors and because they more often hide the underlying structure that is informative. Would you be against defining this as a newtype?

bin/src/Main.purs
JordanMartinez: If `Map PackageName PackageMap` was defined as a type (preferably a newtype per my comment about hating type aliases below), then this value could be named after that type. Not sure what that name would be. Perhaps `packageToDependencies`? Verbose, yes, but still leaves room for `packageDependencies`, `dependencies`. Moreover, it implies a mapping of some sort.

src/Spago/Command/Publish.purs
JordanMartinez: So, is this you saying that we need to add a test for that inside or outside this PR? I'm hoping outside?

src/Spago/Command/Fetch.purs
JordanMartinez: The reason why I split this up was because the error message was off for the second case:
1. Error message makes sense here because we're installing something: https://github.com/purescript/spago/pull/1018/files#diff-4529e47b55cd312f9954aa4027ac032176d4c22848d7e5ccf2c7adb942587e38L99
2. but not here because we could NOT pass the `install` flag and instead just be doing the ensure ranges. In that situation, why would the error mention anything about installation? https://github.com/purescript/spago/pull/1018/files#diff-4529e47b55cd312f9954aa4027ac032176d4c22848d7e5ccf2c7adb942587e38L112

Oh, and where is the third time it's being duplicated?

test/Spago/Build/Polyrepo.purs
JordanMartinez: What happens if we add more tests here? Do we link to those PRs as well?

I think I'd rather copy-paste the Mermaid diagram into this module's docs. If we add a new test to this situation, we can update the diagrams as well.

JordanMartinez:
> > I think this should be further discussed before we work on it. At the very least, I don't plan on doing anything for this in this PR.
> 
> Ah yes about this: agreed that we can shrink the PSA display in a future PR, but I'd like to address the multiple `Build Succeeded` messages in here, as they don't really make sense right now

What about them doesn't make sense? I don't have enough of an idea for how to move forward on this.

bin/src/Main.purs
f-f: I think what I'm getting at is that the content carried by this name has not changed so much to warrant a project-wide rename

src/Spago/Command/Fetch.purs
f-f: > In Build, this extra info isn't needed and may be hard to construct.

Ah you're right indeed - Build needs less stuff. Nevermind then!

src/Spago/Command/Fetch.purs
f-f: Lines 104 and 128 are the same messages with different structures, 146 is a slightly different but with the same structure as 128.

I guess this is alright, but we should at least collect the shared error message in one place

src/Spago/Command/Fetch.purs
f-f: Newtypes are good for things that can be mixed up with other things, but I doubt we can misuse a `Map PackageName PackageMap` in the same way you'd misuse a `String`.

This is the only place where type aliases shine, because it doesn't really matter if the compiler will make it disappear, it's just there to hide the inner structure (which is a feature in this case. You'd only consume this type with the functions from this file)

test/Spago/Build/Polyrepo.purs
f-f: Ah yes good point. Let's go with adding the mermaid diagrams here.

test/Spago/Build/Polyrepo.purs
f-f: I don't understand - `spago init` boils down to "make a Main, a Test.Main, and a basic spago.yaml". (there are a few more details, but they are negligible)

In the snippet above you call `spago init`, delete the three main things it creates, and then make a new configuration.
Hence why I'm saying: we might as well just make the configuration without calling the `init` first?

src/Spago/Command/Publish.purs
f-f: Of course, definitely not here! I was just issuing a general reminder for both of us 😄

f-f:
> What about them doesn't make sense? I don't have enough of an idea for how to move forward on this.

The `build` is - even if split into multiple parts as we are doing here - a single process with a boolean outcome: either it's good and you can move on, or it's broken and needs fixing.

The problem with multiple "Build succeeded" messages is that it breaks this perspective, a build can now succeed multiple times, but fail only once.

To tackle this we should move this snippet out of the `Psa` module and into the `Build` module:
https://github.com/purescript/spago/blob/899b4939b782aa04932a4c038e72479b9cc8705c/src/Spago/Psa.purs#L109-L113

We'd return this list of errors from the Psa command instead of killing the process there, and draw the final verdict on the build only once they all succeed or one of them fails.

test/Spago/Build/Polyrepo.purs
JordanMartinez: Ah, I see your point now.

src/Spago/Command/Fetch.purs
JordanMartinez: Hm.... I looked at sharing the error message, but then came up with a different idea. What if we defined the current package install block of code and 'ensureRanges' block of code into let-bound functions and then just call those functions in their respective places?

f-f: I'm not sure how that would look like, but we can try it out 😄

JordanMartinez: I think there's a bug in the current PR. If I have a package in the workspace `spago.yaml`, and don't select a package BUT I do have the `ensureRanges` flag, then the the root package doesn't get ranges added to it. I believe this occurs because I only mark the root package as a `SinglePackage` if the `packages` to install is not null. However, I don't also check whether the `ensureRanges` value is true.

f-f: Ah I see - can we write a test to verify this?

JordanMartinez: I still need to add a test, but see latest commits for what I did here.

src/Spago/Command/Build.purs
JordanMartinez: Done.

test/Spago/Build/Polyrepo.purs
JordanMartinez: Done.

src/Spago/Command/Fetch.purs
JordanMartinez: Hm... This still seems to differ from the current behavior. Sigh...

JordanMartinez: I've updated the code back to the original approach here. Seems simpler at the end of the day.

src/Spago/Command/Fetch.purs
JordanMartinez: I went with `PackageTransitiveDeps` for a type alias and `toAllDependencies` due to working on the code before seeing this. Would you like me to use your naming?

JordanMartinez:
> We'd return this list of errors from the Psa command instead of killing the process there, and draw the final verdict on the build only once they all succeed or one of them fails.

How does this work when a non-js backend is used? There's already a "Backend build succeeded" message.

JordanMartinez:
Remaining tasks:
- the `ensureRanges` test I have yet to add
- address feedback on latest changes
- address the psa `Build Succeeded` issue

f-f:
> How does this work when a non-js backend is used? There's already a "Backend build succeeded" message.

Ah yes, indeed - when building multiple packages with a backend we'd first build with purs, then build with the backend, then repeat. Alright, let's leave this out of this PR and just open an issue once we merge

src/Spago/Command/Fetch.purs
f-f: No, this is great!

f-f (approved):
This is great now - thanks @JordanMartinez!

Let's add the last test for `ensureRanges`, then we can merge in

JordanMartinez:
In latest commits, I...
- moved the build-in-topological-order tests into their own `describe` block
- kept case 4 (duplicate module check) in the same location, but renamed it for clarity
- moved Mermaid diagrams above each test for readability, and linked to the guide for how to write one for future users
- added `--ensure-ranges` tests:
   - poly repo where a root package does and does not exist
   - monorepo where a root package exists. I didn't add one for when a root package does not exist as I don't think this will ever happen.
   
I won't merge this despite the approval in case you have more feedback. But if it looks good, then feel free to merge it.

JordanMartinez:
> > How does this work when a non-js backend is used? There's already a "Backend build succeeded" message.
> 
> Ah yes, indeed - when building multiple packages with a backend we'd first build with purs, then build with the backend, then repeat. Alright, let's leave this out of this PR and just open an issue once we merge

I've tracked our conversation about this here and added it to the original issue: https://github.com/purescript/spago/issues/988#issuecomment-1734388070

Issue #1020, opened 2023-09-27
Reinstate the `spago docs` command


f-f:
We currently do not have a `spago docs` command in the PureScript codebase.

We would need to port it from the [old codebase](https://github.com/purescript/spago-legacy/blob/11e68ece50ede286660452ef370baa3ab5c33c02/src/Spago/Build.hs#L500), with the difference that since both projects are in PureScript we could just use it as a library instead of calling a separate node process.

So a tentative checklist for things to do could be:
- update [`docs-search`](https://github.com/purescript/purescript-docs-search) to use the new Spago
- add it as a dependency
- port [the implementation for the `docs` command from Haskell](https://github.com/purescript/spago-legacy/blob/11e68ece50ede286660452ef370baa3ab5c33c02/src/Spago/Build.hs#L500)
- since the `docs` command injects a Halogen app in the generated html, we'd need to bundle that app when bunling/publishing Spago and put it where Spago can find it at runtime, so it can copy it over

CharlesTaylor7:
I have a draft PR for the first item:
https://github.com/purescript/purescript-docs-search/pull/69

f-f:
@CharlesTaylor7 lovely, thanks! There's an unfortunate amount of incidental complexity in trying to get that repo up to date, so let's try something different.
I have ported the codebase [here inside this repo](https://github.com/purescript/spago/tree/master/docs-search) so we can do the change without the complications of CI and keeping backwards compatibility with the existing stuff.

Can you apply the same changes here? We won't have to get it working, it would just need to compile to merge this in.

CharlesTaylor7:
@f-f Ok, understood. I made a new draft PR against this repo: https://github.com/purescript/spago/pull/1023

CharlesTaylor7:
Rough todo list of remaining work:
- [ ] In `docs-search`, remove `argonaut-generic` dep and related generic instances in favor of `codec-argonaut` codecs.
- [ ] In `docs-search`, Swap out doc types and ast types for the ones defined in `language-purescript`
- [ ] Enable search indexing in the `spago docs` command
- [ ] Bundle `docs-search`s halogen app during spago publish. Inject the app at runtime after indexing.

PR #1025, opened 2023-09-28
Docs command

CharlesTaylor7:
### Description of the change

Reimplements the basic `spago docs` command.

Supports the `--format` flag, and the new `--package` flag, that the new spago commands use.
Otherwise the remaining Spago docs commands have not been ported yet.

Part of #1020

### Checklist:

### DIFF
--- a/test/Spago/Docs.purs
+++ b/test/Spago/Docs.purs
@@ -1,0 +1,29 @@
+module Test.Spago.Docs where
+
+import Test.Prelude
+
+import Node.FS.Aff as FSA
+import Node.Path as Path
+import Registry.Version as Version
+import Spago.Command.Init as Init
+import Spago.Core.Config as Config
+import Spago.FS as FS
+import Test.Spec (Spec)
+import Test.Spec as Spec
+import Test.Spec.Assertions as Assert
+
+spec :: Spec Unit
+spec = Spec.around withTempDir do
+  Spec.describe "docs" do
+
+    Spec.it "documents successfully with no flags" \{ spago } -> do
+      spago [ "init" ] >>= shouldBeSuccess
+      spago [ "docs" ] >>= shouldBeSuccess
+
+    Spec.it "builds successfully a solver-only package" \{ spago } -> do
+      spago [ "init", "--name", "aaa", "--use-solver" ] >>= shouldBeSuccess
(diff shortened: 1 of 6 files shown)


f-f:
Think we can skip the `--package` flag - I struggle to think about a case where one would want docs for only a chunk of the project and not the whole thing

CharlesTaylor7:
> Think we can skip the `--package` flag - I struggle to think about a case where one would want docs for only a chunk of the project and not the whole thing

I can see it being useful for the same reasons you may use it for `spago build` or `spago test` it's a slow operation and you may want to incrementally work on a single package. I feel like these commands establish a bit of a precedent too. In the interest of "Least Surprise", I think it makes sense to include.

For example consider a full stack purescript app with frontend and backend packages. Maybe you're working offline and want to look at dependencies for the frontend package?

CharlesTaylor7:
> Think we can skip the `--package` flag - I struggle to think about a case where one would want docs for only a chunk of the project and not the whole thing

Ok, I went ahead and dropped it, my mentioned use case was only hypothetical.

test/Spago/Docs.purs
```diff
@@ -0,0 +1,29 @@
...
+import Test.Spec as Spec
+import Test.Spec.Assertions as Assert
+
+spec :: Spec Unit
+spec = Spec.around withTempDir do
+  Spec.describe "docs" do
+
+    Spec.it "documents successfully with no flags" \{ spago } -> do
+      spago [ "init" ] >>= shouldBeSuccess
+      spago [ "docs" ] >>= shouldBeSuccess
```
f-f, line 21: Could we check that there exist some html files in the place where we are expecting them?

f-f (approved):
One little thing then it's good to go 👏

f-f:
> Ok, I went ahead and dropped it, my mentioned use case was only hypothetical.

Thanks - I had no strong preference on this and I think the one you detailed is a valid use case; I mostly wanted to make things easier to get going, especially given that the package selection machinery is being reshuffled in #1018

test/Spago/Docs.purs
f-f: @CharlesTaylor7 actually I'll merge this, let's get to this in the next PR

PR #1026, opened 2023-09-28
Add 'spago registry package-sets' command

### LINKED ISSUE #997: Add `spago registry package-sets` command to list available package sets
We should have a table (like in `spago ls packages`) that lists the package set version, the compiler version, and the publish date.

Since there are so many of them we should probably list only the latest for each compiler, and have a `--all` flag to list them all.

f-f:
We add a new command to list package sets:

```console
$ spago registry package-sets --latest
+---------+------------+----------+
| VERSION | DATE       | COMPILER |
+---------+------------+----------+
| 10.0.0  | 2023-01-05 | 0.15.4   |
| 20.0.3  | 2023-04-08 | 0.15.7   |
| 27.2.0  | 2023-06-17 | 0.15.8   |
| 29.1.0  | 2023-07-18 | 0.15.9   |
| 42.1.0  | 2023-09-26 | 0.15.10  |
+---------+------------+----------+
```

There are quite a few things happening in the background:
- we now have a sqlite database, so we can cache all of the package sets info, and the index in the future
- there's a fix for the bundler - we now allow arbitrary flags to be passed to the underlying bundler. This is useful in the immediate term - to include the `better-sqlite3` library with the bundle - but should also help for when we support arbitrary bundlers via the backends

Fix #997


JordanMartinez:
Shouldn't this sort the releases by compiler version with the latest one on top?

f-f:
I have no strong preference on this - my only reasoning was that if there are enough sets (e.g. if you don't pass the `--latest` flag) then the output will scroll and the latest, being at the bottom, is the first one you'll see, which is useful.

This is something that I struggle with in `purs` - the first error that you'd like to see is actually all the way up and that requires endless scrolling all the time 🙂

JordanMartinez:
Mm... good point!

f-f:
And don't get me started on what happens when using `--verbose-errors` 😄

PR #1030, opened 2023-09-30
Reintroduce source-map support

### LINKED ISSUE #877: Sourcemaps don't work
Hi, this is more  a question than an issue:

The docs say to build with `--purs-args "-g sourcemaps"` (which didn't work for me):
https://github.com/purescript/spago#get-source-maps-for-my-project

But there is also [a flag `--source-maps`](https://github.com/purescript/spago/pull/562) provided by spago. 

Is there a difference? Are the docs outdated? When to use what?

Maybe it's possible to clarify this in the docs :)

CharlesTaylor7:
Resolves #877 
Reintroduces source map support to `spago bundle`. 

Spago already enables source maps for purs builds. All that was needed was to pass the `--sourcemap` flag to esbuild for the bundling step to have end to end source map support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -871,15 +871,15 @@ $ node -e "import('./index.js').then(m => console.log(m.main))"
 When bundling, you can include `--source-maps` to generate a final source map for your bundle.
 
 Example:
-```
+```console
 spago bundle -p my-project --source-maps --minify --outfile=bundle.js
 ```
 will generate a minified bundle: `bundle.js`, and a source map: `bundle.js.map`.
 
 #### Node
 If your target platform is node, then you need to ensure your node version is >= 12.2.0 and [enable source maps](https://nodejs.org/dist/latest-v20.x/docs/api/cli.html#--enable-source-maps
-) when executing your bundle:
-```
+) when executing your script:
+```console
 spago bundle -p my-project --platform node --source-maps --minify --outfile=bundle.js
 node --enable-source-maps bundle.js
 ```
@@ -888,7 +888,7 @@ node --enable-source-maps bundle.js
 If you are targeting browsers, then you will need to ensure your server is configured to serve the source map from the same directory as your bundle.
 
 So for example if your server is configured to serve files from `public/`, you might run:
-```
+```console
 spago bundle -p my-project --platform browser --source-maps --minify --outfile=public/bundle.js
 ```
 
--- a/README.md
+++ b/README.md
@@ -876,6 +876,14 @@ spago bundle -p my-project --source-maps --minify --outfile=bundle.js
 ```
 will generate a minified bundle: `bundle.js`, and a source map: `bundle.js.map`.
 
+If your target platform is 'node', then you need to [enable source maps](https://nodejs.org/dist/latest-v20.x/docs/api/cli.html#--enable-source-maps
+) when executing your bundle:
+```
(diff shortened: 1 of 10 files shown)


bin/src/Flags.purs
```diff
@@ -306,3 +306,10 @@ ensureRanges =
     ( O.long "ensure-ranges"
         <> O.help "Add version bounds for all the dependencies of the selected project"
     )
+
+sourceMap :: Parser Boolean
+sourceMap =
+  O.switch
+    ( O.long "source-map"
```
CharlesTaylor7, line 313: Technically the command creates a single source map, but the old spago command was plural. `--source-maps`. The esbuild flag is singular though.

f-f: I'd prefer the plural version,`purs` uses that

f-f (commented):
Great to hear that we don't need much to support this! Could we add:
- a test that checks that esbuild generates the sourcemaps
- documentation on how to make and use them

CharlesTaylor7:
Is spago normally flaky for the windows test suite? It is timing out for 1 test.

f-f:
@CharlesTaylor7 yes it is flaky - we suspect that this will be fixed once we upgrade to the latest package set, which has improvements for Node libraries and for Execa

README.md
```diff
@@ -865,6 +866,17 @@ $ node -e "import('./index.js').then(m => console.log(m.main))"
...
+
+The `spago build` step generates source maps automatically.
+When bundling, you can include `--source-maps` to generate a final source map for your bundle.
+
+Example:
+```
+spago bundle -p my-project --source-maps --minify --outfile=bundle.js
+```
+will generate a minified bundle: `bundle.js`, and a source map: `bundle.js.map`.
+
```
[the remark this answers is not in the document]
CharlesTaylor7: Ok, I'm not sure about node. 
But for chrome, it's automatic as long as your server serves the source map from the same directory as the js bundle. When you open the dev tools, chrome will make an additional network requests for source maps of your loaded assets. 
How do you want to summarize that?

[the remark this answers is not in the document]
CharlesTaylor7: Ok it looks like node has support for source maps since version 12.12.0. It does require an additional flag, so I'll document it as well as the dev tools experience.

https://nodejs.org/en/blog/release/v12.12.0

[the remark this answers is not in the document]
f-f: I think it would be fine to just link [this](https://developer.chrome.com/docs/devtools/developer-resources/) for Chrome, and just say that Chrome should pick it up if it can be found alongside your code

test/Spago.purs
```diff
@@ -37,18 +37,21 @@ main = Aff.launchAff_ $ void $ un Identity $ Spec.Runner.runSpecT testConfig [ S
     -- TODO:
     -- Upgrade set?
     -- Script?
-
+    {-
```
f-f, line 40: Let's run all the tests 😄 

Note that you can use `itOnly` and `describeOnly` to have the `spec` run only that test/suite

CharlesTaylor7: 😅 My bad, thanks for the tip.

f-f: No worries!

f-f (approved):
This is great, thank you!

PR #1034, opened 2023-10-01
docs-search: Upgrade to new Web.Bower.Json module

CharlesTaylor7:
Updates `docs-search` to depend on `purescript-language-purescript` instead of `bower-json` for the module `Web.Bower.PackageMeta`.

Note: most of the `purescript-language-purescript` library functionality is not used, I'm aiming to replace the types and codecs in `docs-search` piecemeal.

### DIFF
--- a/docs-search/src/Docs/Search/PackageIndex.purs
+++ b/docs-search/src/Docs/Search/PackageIndex.purs
@@ -2,8 +2,8 @@ module Docs.Search.PackageIndex where
 
 import Docs.Search.Config as Config
 import Docs.Search.Extra (stringToList)
-import Docs.Search.Score (Scores, getPackageScoreForPackageName, normalizePackageName)
-import Docs.Search.Types (PackageName, RawPackageName(..), PackageScore)
+import Docs.Search.Score (Scores, getPackageScoreForPackageName)
+import Docs.Search.Types (PackageName(..), PackageScore)
 import Docs.Search.Loader as Loader
 
 import Prelude
@@ -15,8 +15,11 @@ import Data.Maybe (Maybe)
 import Data.Newtype (unwrap)
 import Data.Search.Trie (Trie)
 import Data.Search.Trie as Trie
+import Data.Tuple as Tuple
 import Effect.Aff (Aff)
+import Safe.Coerce (coerce)
 import Web.Bower.PackageMeta (PackageMeta(..))
+import Web.Bower.PackageMeta as Bower
 
 type PackageResult =
   { name :: PackageName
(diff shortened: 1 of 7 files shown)


docs-search/src/Docs/Search/PackageIndex.purs
```diff
@@ -56,14 +59,11 @@ mkPackageInfo packageScores pms =
...
-          unwrap dependencies <#>
-            _.packageName >>> RawPackageName >>> normalizePackageName
-      , repository: repository <#> (_.url)
+      , dependencies: coerce $ dependencies <#> Tuple.fst
+      , repository: repository <#> unwrap >>> (_.url)
       }
-
     where
-    packageName = normalizePackageName $ RawPackageName name
+    packageName = coerce name
```
CharlesTaylor7, line 66: The left hand side is a `PackageName` from `Docs.Search.Types`. The right hand side `name` is a `PackageName` from `Bower.Web.PackageMeta`.
Both are functionally the same, newtypes around string.

You might ask, why am I not using the latter? 

Well it is because the latter type does not implement `EncodeJson` or `DecodeJson`. 

Much of this library relies on generic deriving of these classes. For example, `PackageInfo` contains `PackageName` if I change out the definitions, then the generic instances for `PackageInfo` break.

f-f: Ah yes, great points. We should:
- not rely on generic Json derivation, because it's opaque - if the type itself changes then it's hard to see how the Json conversion changes as well, which means having explicit implementations for Json conversions
- use [codecs](https://pursuit.purescript.org/packages/purescript-codec-argonaut/10.0.0) instead of typeclasses for encoding json - this is how it's done in the rest of the Spago codebase, you can see examples if you search for `JsonCodec`

This is a small patch so I wouldn't mind doing the porting here, but we can also take it in a new PR if you'd like

spago.yaml
```diff
@@ -120,4 +150,148 @@ workspace:
...
+        - nullable
+        - ordered-collections
+        - partial
+        - prelude
+        - record
+        - safe-coerce
+        - strings
+        - these
+        - tuples
+        - unsafe-coerce
```
f-f, line 276: This one is in the registry. It can just be:

```suggestion
    json-codecs: "4.0.0"
```

You can use `spago registry search $PACKAGENAME` to see if packages are in the registry

docs-search/src/Docs/Search/PackageIndex.purs
CharlesTaylor7: There are 13 types in `docs-search` that derive generic Encode/Decode Json instances. 
Instead of changing just 1 or 2 in this PR, can I do separate PR to convert all those instances?

f-f: Of course!

PR #1035, opened 2023-10-03
`docs-search` cleanup: Replace generic codecs with `codec-argonaut` ones

CharlesTaylor7:
Part of #1020 

There are no user facing changes, this is refactoring.

Replaces some but not all of the generic `EncodeJson`/`DecodeJson` instances with codecs derived from `codec-argonaut`.

These modules were left mostly as is:
- `Docs.Search.DocsJson`
- `Docs.Search.TypeDecoder`

And that's because they are going to be replaced with equivalent functionality from the `language-purescript` package.


docs-search/src/Docs/Search/JsonCodec.purs
```diff
@@ -0,0 +1,51 @@
...
+inject = inj (Proxy :: _ sym)
+
+fromJsonUnidirectional
+  :: forall a
+   . (Json -> Either DecodeError a)
+  -> (Json -> Either JsonDecodeError a)
+fromJsonUnidirectional = map $ left convertError
+  where
+  convertError :: DecodeError -> JsonDecodeError
+  convertError = const $ UnexpectedValue $ fromString ""
```
CharlesTaylor7, line 38: This block of code is why the PR is still in draft mode. I need to figure how to best map error types between Jordan's `json-codecs` and `codec-argonaut`.

docs-search/src/Docs/Search/JsonCodec.purs
```diff
@@ -0,0 +1,51 @@
...
+inject = inj (Proxy :: _ sym)
+
+fromJsonUnidirectional
+  :: forall a
+   . (Json -> Either DecodeError a)
+  -> (Json -> Either JsonDecodeError a)
+fromJsonUnidirectional = map $ left convertError
+  where
+  convertError :: DecodeError -> JsonDecodeError
+  convertError = printDecodeError >>> TypeMismatch
```
CharlesTaylor7, line 38: I'm cheating here and using `TypeMismatch` as a custom error type, to easily convert from a `json-codecs` error into a `codec-argonaut` error.

@JordanMartinez @f-f Thoughts?

f-f:
> These modules were left mostly as is

Shall we port them already in this PR? So we can drop the typeclass-based approach all in one go instead of being in a transitional state

CharlesTaylor7:
@f-f 
> Shall we port them already in this PR? So we can drop the typeclass-based approach all in one go instead of being in a transitional state

I structured it this way with ease of review in mind. And to make sure I was on the right track.
The next set of changes will be probably much larger than this PR. 
I'm fine to do them separately or together, but if have any thoughts on the PR as it is currently I'm all ears.

f-f:
Ah yes sounds good, I'll have a look at this now then 👍

docs-search/src/Docs/Search/BrowserEngine.purs
```diff
@@ -2,21 +2,22 @@
 module Docs.Search.BrowserEngine where
 
 import Docs.Search.Config as Config
-import Docs.Search.PackageIndex (queryPackageIndex)
 import Docs.Search.Engine (Engine, EngineState, Index)
-import Docs.Search.SearchResult (SearchResult)
+import Docs.Search.ModuleIndex as ModuleIndex
+import Docs.Search.PackageIndex (queryPackageIndex)
+import Docs.Search.SearchResult (SearchResult, searchResultCodec)
```
f-f, line 8: The import style throughout the whole Spago codebase is to:
- import everything qualified
- ..except that types can be imported unqualified

So this would become:
```suggestion
import Docs.Search.ModuleIndex as ModuleIndex
import Docs.Search.PackageIndex as PackageIndex 
import Docs.Search.SearchResult as SearchResult
import Docs.Search.SearchResult (SearchResult)
```

CharlesTaylor7: Sure, I'll make the changes to use qualified names for the imports that I touch in this PR (and future ones).

docs-search/src/Docs/Search/JsonCodec.purs
CharlesTaylor7: The more I think about this, I don't think this deviates too far from the intended usage of `TypeMismatch`.

JordanMartinez: Whenever I need to use a custom message in codec-argonaut/argonaut-codec, I use `UnexpectedValue $ encodeJson "Here's my custom message"`.

docs-search/src/Docs/Search/JsonCodec.purs
JordanMartinez: Print my lib's error and then encode that as JSON?

f-f (approved):

PR #1037, opened 2023-10-03
Drop  docs-search cli and interactive mode

CharlesTaylor7:
Spago is going to interface with `docs-search` as a library instead of a CLI. 
So with that in mind, we're dropping support for the CLI and interactive mode offered by the package. 

This reduces the surface area of maintenance while working on issue #1020.

The modules dropped are not required to reach parity with the old spago docs command.
But they can be brought back and upgraded if there's a desire to reintroduce the interactive search feature under a new command such as `spago docs --interactive`.


f-f (approved):

docs-search/src/Docs/Search/Main.purs
```diff
@@ -1,162 +0,0 @@
...
-    )
-
-defaultDocsFiles :: Array String
-defaultDocsFiles = [ "output/**/docs.json" ]
-
-defaultBowerFiles :: Array String
-defaultBowerFiles = [ ".spago/*/*/bower.json", "bower_components/purescript-*/bower.json" ]
-
-defaultSourceFiles :: Array String
-defaultSourceFiles = [ "src/**/*.purs" ]
```
CharlesTaylor7, line 156: Here.

f-f: ?

PR #1040, opened 2023-10-03
Allow reading a custom package set from a local path

### LINKED ISSUE #1031: Allow for filesystem-local package sets
We should really allow users to use local package sets. So this bit
https://github.com/purescript/spago/blob/86babae467a4723d1b0f4732697d758b6008be18/core/src/Config.purs#L383-L385

would become:
```purescript
 data SetAddress 
   = SetFromRegistry { registry :: Version } 
   | SetFromUrl { url :: String, hash :: Maybe Sha256 }
   | SetFromPath { path :: FilePath }
```

f-f:
Fix #1031: it's now possible to point Spago at a local package set, like so:

```yaml
workspace:
  extra_packages: {}
  package_set:
    path: local-package-set.json
```

The format for package sets is the usual one, as detailed in the [docs](https://github.com/purescript/spago#custom-package-sets)


test-fixtures/local-package-set-config.yaml
```diff
@@ -0,0 +1,13 @@
...
+    - effect
+    - prelude
+  name: aaaa
+  test:
+    dependencies: []
+    main: Test.Main
+workspace:
+  extra_packages: {}
+  package_set:
+    path: local-package-set.json
```
anttih, line 13: Do we know if relative paths like `../../file.json` work here?

f-f: It does, but I'll add a test

f-f: Done

PR #1041, opened 2023-10-04
Split docs-search into packages

CharlesTaylor7:
Splits docs-search into 3 packages.

- docs-search-index
- docs-search-common
- docs-search-client-halogen


CharlesTaylor7:
(PS: You might think me crazy for introducing this while #1035 is in the works. But I tested rebasing #1035 off these changes and git handles rebasing off renamed files just fine. )

docs-search/client-halogen/src/Docs/Search/BrowserEngine.js
```diff
@@ -0,0 +1,29 @@
+/* global exports */
```
CharlesTaylor7, line 1: Git doesn't detect this file as a rename, for some reason. But it is.

docs-search/client-halogen/src/Docs/Search/BrowserEngine.purs
```diff
@@ -108,14 +107,6 @@ browserSearchEngine =
   , queryModuleIndex: ModuleIndex.queryModuleIndex
   }
 
--- | Find in which part of the index this path can be found.
-getPartId :: List Char -> PartId
-getPartId (a : b : _) =
-  PartId $ (Char.toCharCode a + Char.toCharCode b) `mod` Config.numberOfIndexParts
-getPartId (a : _) =
-  PartId $ Char.toCharCode a `mod` Config.numberOfIndexParts
-getPartId _ = PartId 0
```
CharlesTaylor7, line 117: `getPartId` has been moved to `Docs.Search.Config`, this was the only change necessary so that `Docs.Search.IndexBuilder` did not depend on `Docs.Search.BrowserEngine`.

docs-search/client-halogen/spago.yaml
```diff
@@ -0,0 +1,9 @@
+package:
+  name: docs-search-client-halogen
+  dependencies:
+    - docs-search-common
+    - halogen
+    - halogen-css
+    - js-uri
+    - markdown-it-halogen
+    - profunctor-lenses
```
CharlesTaylor7, line 9: I believe old spago was more picky about listing out dependencies, than new spago is.
Basically, what I did to construct this list:

1. Copy over all the deps from `docs-search/spago.yaml`
2. Clear built artifacts: `rm -rf .spago output`
3. Comment out all the deps here
4. Attempt to build: `spago build -p docs-search-client-halogen`
5. Renable a package to fix a build error, then go to step 4.

And surprisingly the package builds even though required packages like `prelude` are not listed here.

Old spago would warn if an imported module was not listed as a package dependency. In new spago, I'm able to omit deps which are pulled in transitively. And that doesn't seem intended.

docs-search/client-halogen/src/Docs/Search/BrowserEngine.js
f-f: Did you use `git mv`?

docs-search/client-halogen/spago.yaml
f-f: The behaviour is still there, it's just better in the new Spago:
- it's optional rather than mandatory, and you'll need the `pedantic-packages` flag/config bit to make it check that your packages are exactly the ones you need
- and when it complains, it gives you pointers to modules
- when publishing a package this is however turned on by default, so we only publish "correct" dependencies

f-f: I would be ok with enabling `pedantic-packages` in the global config, so we always detect incorrect dependencies

CharlesTaylor7: Ok, thanks for the heads up. The UX is indeed improved!
I liked being able to copy the commands directly from the terminal. 

I noticed that its pedantic about test deps too.
I'm straying from the original topic, but can I suggest a change in behavior regarding test deps?

I want spago to treat the test modules deps as if they were declared as the union of the project deps and the test deps.
 
That is, it would still pedantically warn if I screw up the new test dependency list, but it wouldn't require me to reiterate project deps.

CharlesTaylor7: I noticed that its pedantic about test deps too.
I'm straying from the original topic, but can I suggest a change in behavior regarding test deps?

I want spago to treat the test modules deps as if they were declared as the union of the project deps and the test deps.
 
That is, it would still pedantically warn if I screw up the new test dependency list, but it wouldn't require me to reiterate project deps.

f-f:
@CharlesTaylor7 we generally try to not force push in PR branches here, as it makes it hard to figure out how the PR is evolving - merging from trunk is much preferred.
There's no need to force-push in any case, as we squash the PR in one commit when merging.

f-f (approved):
This is good to go I think 👏

docs-search/client-halogen/spago.yaml
CharlesTaylor7: It's possible, I may have jumped to conclusions too rapidly. 
When running `spago build --pedantic-package -p spago`.

I got output this output: 
```console
  Run the following command to install them all:
  spago install -p spago aff-promise ansi console control enums exceptions formatters node-fs-aff routing-duplex unfoldable unicode
  
  ...
  
  Run the following command to install them all:
  spago install --test-deps -p spago console enums identity newtype node-fs-aff quickcheck

```

So all that is really happening is that the missing deps are reported for both the project and test deps.

CharlesTaylor7: I may still open an issue for this. I have two semi related problems:
(1) Re-reporting deps that belong in the project dep list and not the test deps will lead to users doing the wrong thing.
(2) I would like to see all the recommended install commands output at the bottom of the terminal output. Scrolling up and down the list to copy the commands is a bit painful.

CharlesTaylor7:
@f-f I fixed up the dependency lists for the `docs-search` packages. But I deferred adding a pedantic package flag, and any changes to the other spago.yaml files, just to keep this PR well scoped.

docs-search/client-halogen/spago.yaml
f-f: @CharlesTaylor7 ah I see now - agreed that these should both be fixed 👍

f-f:
Yep, sounds great! I'll merge.

PR #1045, opened 2023-10-06
Migrate PSA options to package config

JordanMartinez:
### Description of the change

Second part of #988. 

Drops the `show-source` flag/config and defaults that to `true`
Drops the `persistent-warnings` flag/config and stops using cache files.

#### Explanation behind `PathDecision` type

To understand the changes made in this PR, we need to first look at the `Spago/Psa/Output.purs` file. There are three changes there that basically propagate to the rest of the codebase.

The first change is `pathOf`. In the original code, `pathOf` determines whether an error/warning is associated with library source code or the project source code by resolving the error/warning's file name to an absolute file path and then seeing if that path is a prefix of a known library path. Once this is known, various checks later in the `Output.purs` file (i.e. the next two changes) will only run if the path is known to be `Lib` or `Src`. 

The second change is how `update`'s `tag'` value is calculated. Here is where the `--strict` flag comes into play. If we're printing the error and we know (based on the `pathOf` computation above) that the path is a `Src` path, then we promote it to Error if `--strict` was passed.

The third change is `shouldShowError`. It always shows errors, but it sometimes hides warnings. This is where `--censorBuildWarnings`, `--censor-codes`, and `--filter-codes` comes into play. The important bit here is that various checks are made depending on whether the `pathOf` computation above was `Lib` or `Src`.

When reviewing the code, I noticed that it checked for `Lib` and `Src` multiple times. I thought it would be more convenient if `pathOf` calculated all of that information, and then we could use that information later. It seemed to be simpler and opens up more possibilities for what kinds of warnings we can hide (e.g. hide warnings that start with "Data.Map's `Semigroup` instance is now unbiased").

So, what information do we need to calculate for each warning/error filename? Just the following:
1. whether this file is within a `Lib` or `Src` directory, which is used in stats calculation
2. whether warnings for files in this directory should be promoted to errors
3. whether warnings for files in this directory based on some code should be printed

This leads to a type called `PathDecision` (see `Spago/Psa/Types.purs`). (I would use `PsaPath` rather than `PsaPathType` there, but I want to ensure the `String` arguments taken by the `PsaPath` type's `Lib` and `Src` data constructors are the correct paths (in this case, the `short` arg in `Spago/Psa/Output.purs`, the paths relative to the PWD).)

We need to easily specify when to hide warnings for some files (e.g. `.spago` files), show warnings from other files (e.g. `package-a`), and hide warnings from others (e.g. `package-b`). So, my solution was to pass in an `Array` of `String -> Maybe PathDecision`. We iterate through the array, passing the error/warning's absolute file path as the `String` argument. When that file path is a child of some parent directory via a prefix check, we provide the `PathDecision` to use for that path. If the prefix check fails, we try the next one until one matches. If none work, we fallback to a default value.

#### From CLI/Configuration values to `PathDecision`

However, the current CLI interface and configuration format assumes the following concepts:
- `censorBuildWarnings`: an all-or-nothing censoring of warnings
- `censorCodes`: hide these codes
- `filterCodes`: show these codes if not hidden by `censorCodes`

So, we need a way to translate from that interface to this simplified format. So, I relocated the previous `Spago/Psa/Output.purs` `shouldShowError` code into `Spago/Psa.purs`' `toPathDecision`.

We also allow CLI flags to override the configuration. So, I had to determine where in our codebase this overriding (e.g. `fromMaybe defaultValue $ cliFlag.value <|> config.value`) should now occur. Previously, this was done in `bin/src/Main.purs`' `mkBuildEnv` because only the workspace was affected. In the new design, we're using information from both the workspace and each workspace package included in a given build. On a multi-package build, this may include multiple packages. So, I moved the CLI-flag-overriding code into `Spago/Command/Build.purs` because I don't know how to override a package's `build` options (e.g. `--strict`) earlier in the process.

After that, I split the following CLI flags / config values into their corresponding parts:
- `censorBuildWarnings`: an all-or-nothing censoring of warnings
    - `censorLibWarnings`: for `.spago` files
    - `censorProjectWarnings`: for workspace package files
- `censorCodes`: hide these codes
    - `censorLibCodes`: for `.spago` files
    - `censorProjectCodes`: for workspace package files
- `filterCodes`: show these codes if not hidden by `censorCodes`
    - `filterLibCodes`: for `.spago` files
    - `filterProjectCodes`: for workspace package files

#### Further considerations:

- ~We're adding 3 more CLI flags/config values. Unlike `--strict`, I don't think this will be used all that often. So, I propose we either drop support for these flags on the CLI and just use config values OR simplify this to just two CLI arguments, `censor-library-warnings` and `censor-project-warnings`. Each would take one of four values on the CLI: `none`, `all`, `censor=CodeName`, or `filter=CodeName`.~
    - Fabrizio voted for dropping these CLI args.
- I don't think the PSA stats and other output-related things should be handled in this PR yet. I think that should be done in the third task once this is finished.

### Checklist:


f-f:
Cool, I'll merge #1044 first then, so we can base these changes on that one

JordanMartinez:
Opening comment has been updated. This is now ready for review.

JordanMartinez:
I still need to add a test for a multi-package build, but I can do that after this gets an initial review.

f-f:
> I propose we either drop support for these flags on the CLI and just use config values

@JordanMartinez I vote for this one

core/src/Config.purs
```diff
@@ -142,6 +143,21 @@ backendConfigCodec = CAR.object "BackendConfig"
   , args: CAR.optional (CA.array CA.string)
   }
 
+type PackageBuildOptionsInput =
+  { censor_project_warnings :: Maybe CensorBuildWarnings
+  , censor_project_codes :: Maybe (NonEmptySet.NonEmptySet String)
+  , filter_project_codes :: Maybe (NonEmptySet.NonEmptySet String)
+  , strict :: Maybe Boolean
```
f-f, line 150: Doesn't `pedantic-packages` belong here as well?

test/Spago/Build.purs
```diff
@@ -101,7 +101,7 @@ spec = Spec.around withTempDir do
         }
       spago [ "build", "--strict" ] >>= shouldBeFailure
 
-    Spec.it "built_opts 'strict: true' causes build to fail if there are warnings" \{ spago, fixture } -> do
+    Spec.it "A package config with build 'strict: true' causes build to fail if there are warnings" \{ spago, fixture } -> do
```
f-f, line 104: all the other test cases are lowercase

```suggestion
    Spec.it "having 'strict: true' in a package config fails the build if there are warnings" \{ spago, fixture } -> do
```

f-f (commented):
This looks good - I'll have another go after we remove the flags, but otherwise I like this approach 👍

core/src/Config.purs
JordanMartinez: `pedantic-packages` is not a PSA arg. I do want to move it into the package config, but in a separate PR as I imagine this will affect the order of outputs (e.g. "Build succeeded" messages).

f-f: Not a PSA arg, but the block I'm commenting on is the `build:` section of the package, which should contain that one as well. But indeed, let's have that change in a separate PR

JordanMartinez:
Flags have been removed. I've also added a test for a multi-package build.

f-f (approved):
Fantastic 👏 

The only thing missing is [updating the docs for the configuration file](https://github.com/purescript/spago#the-configuration-file), then we can merge in

JordanMartinez:
:thinking: Actually... there's one other thing. When we enable `strict` for a package, we likely don't want its tests to also inherit that `strict`ness. Should `test` have its own build option as well?

README.md
```diff
@@ -1242,6 +1235,27 @@ package:
   # Optional description for the package
   description: "a useful package"
 
+  # Optional section to further customise the build for this package.
+  build_opts:
```
f-f, line 1239: I think this is just `build`?

```suggestion
  build:
```

f-f:
> Should `test` have its own build option as well?

Thought about it, but I think we can include tests in the overall behaviour, and if someone wants non-strict tests they can move them to another package. Unless this is something we want for e.g. all the core packages, in which case we should treat tests separately. I am a bit shy about this route because tests are fairly intertwined with the rest of the package (they inherit the same dependencies, we always compile them together, etc) so splitting all of that off might not be pleasant

JordanMartinez:
> > Should `test` have its own build option as well?
> 
> Thought about it, but I think we can include tests in the overall behaviour, and if someone wants non-strict tests they can move them to another package. Unless this is something we want for e.g. all the core packages, in which case we should treat tests separately. I am a bit shy about this route because tests are fairly intertwined with the rest of the package (they inherit the same dependencies, we always compile them together, etc) so splitting all of that off might not be pleasant

Isn't that kind of ironic though? `--strict` is one of the main usages of PSA things and moving tests to their own package makes `spago test` pointless, doesn't it?

README.md
JordanMartinez: Thanks.

f-f:
> makes `spago test` pointless, doesn't it?

It does, which is why I am in fact thinking of a possible escape route 😄 

I wonder if we could use the new `PathDecision` stuff to figure out if something is a test path, and exempt it from strictness in this way?
That would avoid adding additional config values specifically for tests (we assume that one never wants to be strict with tests), and lets us keep compiling things all together.

JordanMartinez:
Hacky... but that works for me. It does mean we can't censor warnings in tests, but that's probably fine.

core/src/Config.purs
JordanMartinez: Tracking this here: https://github.com/purescript/spago/issues/1049

PR #1047, opened 2023-10-06
`docs-search`: Update AST & doc types

CharlesTaylor7:
## Description
Replaces the AST types with the relevant ones in `language-purescript`.

The ast definitions support an annotation for every node. This is useful upstream, but not in use here. Instead all annotations have been provided with the value of `unit`. 
Likewise the old AST type  `Type` and the new AST type is a type alias `Type' = Type Unit`.
Likewise `Constraint` is now a higher kinded constructor and the type alias `Constraint'` is for constraints with no annotation.

In addition to annotations multiple AST nodes were added, and the arguments to existing nodes were changed. I apologize for the length of the diff, but I didn't see any feasible way to make it significantly shorter.


docs-search/client-halogen/src/Docs/Search/App/SearchResults.purs
```diff
@@ -594,24 +594,28 @@ renderType = case _ of
...
+  ParensInType _ ty ->
     HH.span_
       [ HH.text "("
       , renderType ty
       , HH.text ")"
       ]
 
+  -- FIXME(ast)
+  Skolem _ _ _ _ _ -> HH.text "Skolem"
+  TUnknown _ _ -> HH.text "TUnknown"
```
CharlesTaylor7, line 606: I may be wrong; but I have the impression that `Skolem` and `TUnknown` only show up during type checkin. I don't believe they exist in well formed `docs.json` files. 
But I would appreciate guidance on how to handle these.

docs-search/common/src/Docs/Search/Declarations.purs
```diff
@@ -110,80 +121,82 @@ resultsForDeclaration scores moduleName indexEntry@(Declaration entry) =
...
--- | links ("v", "t" or "k").
-data DeclLevel = ValueLevel | TypeLevel | KindLevel
+-- | links ("v" or "t" ).
+data DeclLevel = ValueLevel | TypeLevel
 
 declLevelToHashAnchor :: DeclLevel -> String
 declLevelToHashAnchor = case _ of
   ValueLevel -> "v"
   TypeLevel -> "t"
-  KindLevel -> "k"
```
CharlesTaylor7, line 156: Distinguishing between types and kinds is not necessary since they share a namespace. 
We only need hash anchors to distinguish types and values. 

Also because the syntax for declaring kinds has changed from `foreign import kind Foo` 
to `foreign import data Foo :: Type` it would introduce unneeded complexity in the code to distinguish the type level from the kind level.

docs-search/index/test/Docs/Search/TypeJson.purs
CharlesTaylor7, line 1: This module tested the docs.json decoders. Since we're using `language-purescript` now, we can rely on it to do its own testing. And I checked; it has very nice test suite that parses several commonly used core modules.

docs-search/index/test/Docs/Search/TypeQuery.purs
```diff
@@ -21,8 +21,7 @@ import Data.Maybe (Maybe(..))
 
 tests :: Spec Unit
 tests = do
-  describe "TypeQuery parser" do
-
+  describe "TypeQuery parser" $ do
```
CharlesTaylor7, line 24:
```suggestion
  describe "TypeQuery parser" do
```

f-f (approved):
This looks good! I think we can merge after we sort out the Skolem constructor thing

JordanMartinez:
I'm pretty sure skolems only have a value when type checking. If we're wrong, we can always fix it later.

PR #1063, opened 2023-10-08
Spago docs, now with indexing and search

CharlesTaylor7:
Part of #1020 

How to demo:
```console
# checkout my branch
git checkout spago-docs-search

# install new punycode dep, transitive dep of markdown-it
npm i

# bootstrap spago
./bin/bundle.js build

# bundle spago
./bin/index.dev.js bundle -p spago-bin

# bundle docs-search
./bin/bundle.js bundle -p docs-search-client-halogen

# run docs command
./bin/bundle.js docs --open
```

You should see output like:
```
Generating documentation for the project. This might take a while...
[1 of 1] Compiling documentation for Docs.Search.App
Documentation written to: generated-docs/html
Building the search index...
"bower.json" decoding failed for .spago/packages/avar-5.0.0/bower.json: An error occurred while decoding a JSON value:
  Expected value of type 'ROOT."main" - Expected a value of type Array but got String'.
Indexing 1004 modules from 123 packages...
Added 8062 definitions and 9634 type definitions from 123 packages to the search index.
Link: file:///Users/chuck/repos/spago/generated-docs/html/index.html

```

For the issue decoding the bower.json `avar-5.0.0`, [I opened an upstream PR to fix](https://github.com/purescript-contrib/purescript-avar/pull/31). However it is just a log message and the indexing continues in the presence of the error.

<details><summary>Click me for a Gif Demo</summary>

Oct-08-2023 11-29-43

</details>

### Checklist:


CharlesTaylor7:
@f-f We're getting really close to resolving #1020 .

Unfortunately while demoing this I realized type directed search is not working, the index is only responding to searches for concrete types and values.

 I'm planning to get type directed search working in a follow up PR.

.github/workflows/build.yml
```diff
@@ -75,8 +75,8 @@ jobs:
       - name: Bootstrap executable
         run: ./bin/index.dev.js bundle -p spago-bin
 
-      - name: Check code format
-        run: npm run format:check
```
CharlesTaylor7, line 79: I introduced this check a few PRs ago, but completely missed that this step already exists on line 84.

docs-search/common/src/Docs/Search/Config.purs
```diff
@@ -84,13 +84,22 @@ penalties =
...
+defaultPackageName = wrap "<local package>"
+
+defaultDocsFiles :: Array String
+defaultDocsFiles = [ "output/**/docs.json" ]
+
+defaultBowerFiles :: Array String
+defaultBowerFiles = [ ".spago/*/*/bower.json", "bower_components/purescript-*/bower.json" ]
+
+defaultSourceFiles :: Array String
+defaultSourceFiles = [ "src/**/*.purs" ]
```
CharlesTaylor7, line 105: These are copied verbatim from here: https://github.com/purescript/spago/pull/1037/files#r1349736974

spago.lock
```diff
@@ -283,7 +283,7 @@ workspace:
         - unicode
         - versions
       git: https://github.com/JordanMartinez/purescript-language-purescript.git
-      ref: v0.1.1
+      ref: db4377dea03ba9c5273a93a8368a53f1d87882c1
```
CharlesTaylor7, line 286: Necessary patch until we get a new release:
https://github.com/JordanMartinez/purescript-language-purescript/commit/db4377dea03ba9c5273a93a8368a53f1d87882c1

docs-search/client-halogen/spago.yaml
```diff
@@ -33,3 +33,12 @@ package:
...
     - web-uievents
+  bundle:
+    type: "app"
+    minify: true
+    sourceMaps: true
+    module: Docs.Search.App
+    outfile: "../../output/Docs.Search.IndexBuilder/docs-search-app.js"
+    platform: browser
+    extra_args:
+      - "--alias:punycode=punycode/"
```
CharlesTaylor7, line 44: punycode is a builtin node module. Since we're bundling for browser, `esbuild` complains about it.
The node module is also considered deprecated and recommends using the upstream npm package `punycode`. So its an easy swap-in. 

The extra `/` at the end is how you tell node and esbuild to override a builtin node package with a user-space package.

https://github.com/JordanMartinez/purescript-language-purescript/commit/db4377dea03ba9c5273a93a8368a53f1d87882c1

docs-search/index/src/Docs/Search/IndexBuilder.js
```diff
@@ -1,13 +1,17 @@
...
+import { fileURLToPath } from "node:url";
 
-export function getDirname () {
-  return __dirname;
-};
+export function getDirname() {
+  const fileName = fileURLToPath(import.meta.url);
+  const dirName = path.dirname(fileName);
+  return dirName;
+}
```
CharlesTaylor7, line 11: `spago bundle` has a workaround for patching `__dirname` but it is only applied when bundling for platform=node.

I opted to just patch the source directly here though instead of hacking the `spago bundle` code to shim these for browsers.

docs-search/client-halogen/spago.yaml
```diff
@@ -33,3 +33,12 @@ package:
     - web-html
     - web-storage
     - web-uievents
+  bundle:
+    type: "app"
+    minify: true
+    sourceMaps: true
+    module: Docs.Search.App
+    outfile: "../../output/Docs.Search.IndexBuilder/docs-search-app.js"
```
CharlesTaylor7, line 41: Bizarrely enough, this is where IndexBuilder expects to find the bundled code.
Is there a better way to do this?

https://github.com/purescript/spago/blob/29cd3cad10e0b2dca093ade0a7a2d114186002ba/docs-search/index/src/Docs/Search/IndexBuilder.purs#L411-L412

docs-search/client-halogen/spago.yaml
f-f: Sounds great! Could you add the above comment as a comment in the source just above this line?

docs-search/client-halogen/spago.yaml
CharlesTaylor7: Nevermind, I'm bundling to the bin/ folder now.

docs-search/common/src/Docs/Search/Config.purs
f-f: Ah I see now

docs-search/index/src/Docs/Search/IndexBuilder.js
f-f: Yup this is good 👍

f-f (approved):
This is excellent - let's add that comment about punycode to the Spago config then this is good to merge 👏

CharlesTaylor7:
@f-f Heads up, the `docs-search-app.js` halogen bundle will only get properly embedded when you use `bin/bundle.js` or an official spago release.

But running `./bin/index.dev.js docs` will fail. Because the app bundle is assumed to be a sibling of the spago output, and that's not true for the unbundled dev builds.

CharlesTaylor7:
@f-f The spago offline tests are failing. Do you know what's happening there?

f-f:
@CharlesTaylor7 could you set [this bit](https://github.com/purescript/spago/blob/29cd3cad10e0b2dca093ade0a7a2d114186002ba/test/Prelude.purs#L109) to false so we can have a cleaner output in CI? Right now it's all crowded with logs.

f-f:
> But running `./bin/index.dev.js` docs will fail. Because the app bundle is assumed to be a sibling of the spago output, and that's not true for the unbundled dev builds.

Can we symlink it in the right place so that the bundle picks up the real one and the `index.dev.js` picks up the link?

CharlesTaylor7:
> > But running `./bin/index.dev.js` docs will fail. Because the app bundle is assumed to be a sibling of the spago output, and that's not true for the unbundled dev builds.
> 
> Can we symlink it in the right place so that the bundle picks up the real one and the `index.dev.js` picks up the link?

Ok just realized, the test failure is related to this.
I was able to symlink and get this working with both `bin/index.dev.js` and `bin/bundle.js` when used locally. 
But the tests are failing to find the bundled halogen app. So I'm troubleshooting that now.

CharlesTaylor7:
@f-f The issue was twofold. The doc tests were failing. I have fixed those. I suspected some of the tests weren't sufficiently future proofed, so I reran the workflow for the latest commit to master.

The test suite is failing for that commit too:
https://github.com/purescript/spago/actions/runs/6443301225/job/17533396875#step:14:11115

docs-search/common/src/Docs/Search/Declarations.purs
```diff
@@ -208,7 +208,7 @@ extractPackageName _ Nothing = UnknownPackage
 extractPackageName _ (Just (SourceSpan { name })) =
   fromMaybe LocalPackage do
     topLevelDir <- dirs !! 0
-    if topLevelDir == ".spago" then Package <<< PackageName <$> dirs !! 1
+    if topLevelDir == ".spago" then Package <<< PackageName <$> dirs !! 2
```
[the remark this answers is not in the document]
f-f: We already do a similar matching in the `Purs.Graph` module, and in there we basically:
- construct Globs for each package we have in the build
- [match each module to its glob](https://github.com/purescript/spago/blob/29cd3cad10e0b2dca093ade0a7a2d114186002ba/src/Spago/Purs/Graph.purs#L91)

Not sure if we have a list of packages available here - if we don't we might just want to keep this approach and change how we structure the `packages` folder instead.
The path of a dependency is defined [here](https://github.com/purescript/spago/blob/29cd3cad10e0b2dca093ade0a7a2d114186002ba/src/Spago/Config.purs#L415), so we can change the registry packages to be `packages/$package/$version` rather than `packages/$package-$version`. (if we go for this we'll have to add a `mkdirp` [in Fetch](https://github.com/purescript/spago/blob/29cd3cad10e0b2dca093ade0a7a2d114186002ba/src/Spago/Command/Fetch.purs#L272) to create `packages/$package`)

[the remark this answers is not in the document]
f-f: Something else that we should take care of is the `<local package>` bit: the new Spago allows for more than one package in the repo, so we'd need to find a way to split the local files into their proper packages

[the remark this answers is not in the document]
f-f: If that makes sense to you then we can just merge this as "good enough" and open an issue to track implementing the "proper way" of getting package info

f-f:
Ah it looks like Spago can't find the client bundle on Windows

CharlesTaylor7:
@f-f I don't think windows likes the Unix style symlink. I replaced it with a stub file with the right name, but does nothing except log an error message. 

I have a better solution in mind for a development workflow, which is better than a symlink and rebundling the app constantly. It'll be in a followup PR though.

PR #1066, opened 2023-10-10
Fixup CI

CharlesTaylor7:
Fixes the 7 tests failing on the latest commit of master.


test/Spago/Publish.purs
```diff
@@ -43,11 +43,12 @@ spec = Spec.around withTempDir do
...
-    Spec.it "can get a package ready to publish" \{ spago, fixture } -> do
+    Spec.itOnly "can get a package ready to publish" \{ spago, fixture } -> do
       FS.copyFile { src: fixture "spago-publish.yaml", dst: "spago.yaml" }
       FS.mkdirp "src"
       FS.copyFile { src: fixture "publish.purs", dst: "src/Main.purs" }
       spago [ "build" ] >>= shouldBeSuccess
+      spago [ "install" ] >>= shouldBeSuccess
       doTheGitThing
       -- It will fail because it can't hit the registry, but the fixture will check that everything else is ready
       spago [ "publish", "--offline" ] >>= shouldBeFailureErr (fixture "publish.txt")
```
CharlesTaylor7, line 54: An extra install after build doesn't do anything, because the build step already did that.

An extra install after `doTheGitThing` causes the test to fail with "dirty git index".

Because the `output` directory is getting committed in the test.
The publish command is trying to build version `6.1.0` of Effect.Console, whereas the first build command builds `v6.0.0`. I don't understand why there's a discrepancy though.

f-f: Yeah the issue is that as part of the publishing flow we test that the version bounds for the library are correct, and the Registry solver is coming up with a plan that includes the new console version (published a couple days ago, which is why it stopped working then)

f-f: I think a way to fix it would be to:
- add `console: 6.1.0` to the `extra_packages` in the spago file
- run `spago fetch` before the publish, which should download (but not compile, so no output folder) the right things

f-f: The long-term plan to fix this could be a `spago fetch --solver` flag that will come up with a solver-based plan even when the config is configured with a package set

f-f (approved):
Amazing, thanks!

PR #1070, opened 2023-10-11
Revert `spago build` back to using a single `purs compile` call when building multi-package workspace

### LINKED ISSUE #1051: Rework how we apply PSA options and get back to building all together
Now that we have #1018, building all the packages vs building the package that depends on everything else take very different amount of time even though they are doing the same work:
```
spago build  18.07s user 3.75s system 219% cpu 9.960 total
```
vs
```
spago build -p spago-bin  9.58s user 2.26s system 178% cpu 6.631 total
```

I am not sure now if the latter command is doing "the right thing" (e.g. if the dependencies have psa options then it's probably ignoring those?), but if we are fine with this then we should add a `default_package` option to the workspace section, so that running `spago build` will select the default package instead of all of them.

If we are not fine with this - i.e. we should always compile all the things with the correct PSA options - then I think the only way possible would be to go back to compiling all together, and fix the two reasons why we are compiling in topological order by instead:
- using the graph check to see if there are circular dependencies between workspace packages
- and reworking the PSA path decisions so that it can "do the right thing" even when we are compiling everything together (this will probably need to be assisted by the graph)

JordanMartinez:
### Description of the change

Fixes #1051

### Checklist:

### DIFF
--- a/src/Spago/Command/Build.purs
+++ b/src/Spago/Command/Build.purs
@@ -1,33 +1,26 @@
 module Spago.Command.Build
   ( run
   , BuildEnv
-  , SelectedPackageGlob(..)
   , getBuildGlobs
   ) where
 
 import Spago.Prelude
 
 import Data.Array as Array
-import Data.Either (note)
 import Data.Map as Map
 import Data.Set as Set
-import Data.String as String
 import Data.Traversable (sequence)
 import Data.Tuple as Tuple
-import Dodo as Dodo
-import Effect.Ref as Ref
 import Record as Record
-import Registry.PackageName as PackageName
 import Spago.BuildInfo as BuildInfo
 import Spago.Cmd as Cmd
 import Spago.Command.Fetch as Fetch
-import Spago.Config (Package(..), PackageMap, PackageSet, WithTestGlobs(..), Workspace, WorkspacePackage)
+import Spago.Config (Package(..), PackageMap, WithTestGlobs(..), Workspace, WorkspacePackage)
 import Spago.Config as Config
 import Spago.Git (Git)
 import Spago.Psa as Psa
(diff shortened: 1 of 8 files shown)


src/Spago/Command/Build.purs
```diff
@@ -99,188 +92,105 @@ run opts = do
...
 getBuildGlobs { selected, dependencies, withTests, depsOnly } =
   Set.fromFoldable $ projectGlobs <> monorepoPkgGlobs <> dependencyGlobs <> [ BuildInfo.buildInfoPath ]
   where
+  -- Here we select the right globs for a monorepo setup with a bunch of packages
+  projectGlobs = case depsOnly of
+    true -> []
+    false ->
+      -- We just select all the workspace package globs, because it's (1) intuitive and (2) backwards compatible
+      workspacePackageGlob =<< selected
+
```
JordanMartinez, line 180: The code here should look like it did before I did #1018.

src/Spago/Command/Build.purs
```diff
@@ -99,188 +92,105 @@ run opts = do
...
             die' errors
-
-  when buildingMultiplePackages do
-    moduleMap <- liftEffect $ Ref.read duplicateModuleRef
-    let duplicateModules = Map.filter (\packages -> Array.length packages > 1) moduleMap
-    unless (Map.isEmpty duplicateModules) do
-      die $ duplicateModulesError duplicateModules
+        Nothing -> do
+          -- TODO: here we could go through all the workspace packages and run the check for each
+          -- The complication is that "dependencies" includes all the dependencies for all packages
```
JordanMartinez, line 154: Just FYI, but this work is likely unblocked now.

f-f (approved):
Looks great 👏

PR #1072, opened 2023-10-11
docs-search: Introduce local dev server for Halogen client

CharlesTaylor7:
Introduces a better local development workflow for the halogen client. We are able to skip the bundling step, by using local dev server. 
There are instructions under `docs-search/client-halogen/README.md` for how to make use of this.

Note: I found out while working on this, that you can't just link to ES modules and expect the file system to serve them. There has to be a running server. This is a difference between plain JS and ES modules, due to stricter security measures:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#other_differences_between_modules_and_standard_scripts


docs-search/client-halogen/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "vite-dev",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "vite build",
+    "preview": "vite preview"
```
CharlesTaylor7, line 9:
```suggestion
    "dev": "vite"
```

f-f:
Great! I'll try this out locally 🙂

f-f:
I tried this out, and got this error in the generated docs page:
```
[plugin:vite:import-analysis] Failed to resolve import "../output/Docs.Search.App/index.js" from "../../generated-docs/html/docs-search-app.js". Does the file exist?
```

I don't know if this is what you were referring to?

I also found weird that Vite tried to open a page in Chrome, even tho that's not my default browser..

Generally I'm not sure it makes sense to involve yet another tool in the pipeline, so might be worth taking a step back and thinking what is it that we're trying to solve.

I think there are a few related-but-not-entirely-overlapping goals that we want to get at:
1. enable quick development for the halogen client
2. make sure that we include the halogen-client bundle in the spago release
3. make it possible to access the halogen-client bundle when we use the development version of spago

For (2) we are using the NPM scripts, so that's covered (we could put together a Makefile to automate releases more, but that can wait)

For (1) we are already depending on having `esbuild` in path, so we could use the `--watch` flag for building the bundle on every change, so we get some kind of hot reload (more like a warm reload since you'd still have to refresh the page, but we wouldn't need a dev server).
I think `esbuild` is quick enough for this to be usable, and we could cover (3) as well with this, if we wrap this in a npm script so we don't have to type down the esbuild command all the time.
The additional advantage of this would be to avoid having a `package.json` file in the `client-halogen` package, which would be a bit confusing.

CharlesTaylor7:
> I tried this out, and got this error in the generated docs page:
> 
> ```
> [plugin:vite:import-analysis] Failed to resolve import "../output/Docs.Search.App/index.js" from "../../generated-docs/html/docs-search-app.js". Does the file exist?
> ```
> 

Did you build the client ahead of time? This error just means that docs-search-client-halogen hasn't been built yet. The dev server is only setup to reload the html and js. It's not going to rebuild the purescript.

> 
> I also found weird that Vite tried to open a page in Chrome, even tho that's not my default browser..
> 

That is weird. I might dig into vite's docs on this...


> Generally I'm not sure it makes sense to involve yet another tool in the pipeline, so might be worth taking a step back and thinking what is it that we're trying to solve.
> ...
> The additional advantage of this would be to avoid having a `package.json` file in the `client-halogen` package, which would be a bit confusing.

I put it in a separate subfolder because its an optional development workflow. The Github CI action won't try to install vite. And most contributors to `spago` need not think about it. 
It's tucked away in the folder its most relevant to. 

> For (1) we are already depending on having esbuild in path, so we could use the --watch flag for building the bundle on every change, so we get some kind of hot reload (more like a warm reload since you'd still have to refresh the page, but we wouldn't need a dev server).
I think esbuild is quick enough for this to be usable, and we could cover (3) as well with this, if we wrap this in a npm script so we don't have to type down the esbuild command all the time.

I will give this a go and see the approaches compare in speed and convenience.

PR #1074, opened 2023-10-12
Draft: Shorten CI times with better npm caching on windows

CharlesTaylor7:
Results Inconclusive, come back later.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,7 +51,6 @@ jobs:
             ${{ runner.os }}-build-
             ${{ runner.os }}-
           path: |
-            node_modules
             ~/.npm
 
       - name: Cache PureScript dependencies
@@ -63,7 +62,7 @@ jobs:
             output
 
       - name: Install NPM dependencies
-        run: npm ci --dd
+        run: npm ci --dd --prefer-offline --config=~/.npm
 
       - name: Link dev dependencies as global dependencies
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,7 +62,9 @@ jobs:
             output
 
       - name: Install NPM dependencies
-        run: npm ci --dd --no-audit --cache .npm
+        run: |
+          npm config set cache .npm
+          npm ci --dd --no-audit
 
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -45,15 +45,14 @@ jobs:
         env:
           cache-name: cache-node-modules
         with:
-          path: ~/.npm
```
CharlesTaylor7, line 48: This is the culprit right here. This does function as a global download cache for linux, and macos, but it was hard to find documentation on it. Every guide I looked at said to cache `node_modules` instead. 
On windows restoring this path restored less than 1 MB, where as for Linux and MacOS they were restoring between 25MB and 30MB.

.github/workflows/build.yml
```diff
@@ -45,15 +45,14 @@ jobs:
...
-          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package.json') }}
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
```
CharlesTaylor7, line 48: Since we're installing from the lock file with `npm ci`, every guide also recommended hashing on the lock file and not the base file. This also has the advantage that tweaks to development scripts and metadata won't impact the build cache.

CharlesTaylor7: In short windows was caching nothing npm related, neither local dependencies nor global ones.

.github/workflows/build.yml
```diff
@@ -45,15 +45,13 @@ jobs:
         env:
           cache-name: cache-node-modules
         with:
-          path: ~/.npm
```
CharlesTaylor7, line 48: https://docs.npmjs.com/cli/v6/using-npm/config#cache

f-f:
@CharlesTaylor7 why close this?

CharlesTaylor7:
> @CharlesTaylor7 why close this?

I was using lots of your GitHub Ci resources and not making progress. 
I started a separate PR against my fork, so that I wouldn't be incurring lots of extra builds and caches for you.

f-f:
Oh I see - thanks!

PR #1075, opened 2023-10-12
Migrate pedantic-packages to package config

### LINKED ISSUE #1049: Migrate `pedantic-packages` from workspace config to package config
As mentioned here: https://github.com/purescript/spago/pull/1045#discussion_r1349488225

JordanMartinez:
### Description of the change

Fixes #1049. Ugh... writing these tests man...

### Checklist:

### DIFF
--- a/src/Spago/Command/Build.purs
+++ b/src/Spago/Command/Build.purs
@@ -6,7 +6,9 @@ module Spago.Command.Build
 
 import Spago.Prelude
 
+import Control.Monad.ST as ST
 import Data.Array as Array
+import Data.Array.ST as STA
 import Data.Map as Map
 import Data.Set as Set
 import Data.Traversable (sequence)
@@ -30,6 +32,7 @@ type BuildEnv a =
   , logOptions :: LogOptions
   , workspace :: Workspace
   , psaCliFlags :: PsaTypes.PsaOutputOptions
+  , pedanticPackages :: Boolean
   | a
   }
 
@@ -46,6 +49,7 @@ run opts = do
   , workspace
   , logOptions
   , psaCliFlags
+  , pedanticPackages
   } <- ask
 
   BuildInfo.writeBuildInfo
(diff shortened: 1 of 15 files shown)


src/Spago/Command/Publish.purs
```diff
@@ -127,7 +128,7 @@ publish _args = do
   let globs = Build.getBuildGlobs { selected: [ selected ], withTests: false, dependencies: allDependencies, depsOnly: false }
   maybeGraph <- Graph.runGraph globs []
   for_ maybeGraph \graph -> do
-    graphCheckErrors <- Graph.toImportErrors selected <$> runSpago (Record.union { graph, selected } env) Graph.checkImports
+    graphCheckErrors <- Graph.toImportErrors selected { reportSrc: true, reportTest: false } <$> runSpago (Record.union { graph, selected } env) Graph.checkImports
```
JordanMartinez, line 131: Previously, tests would cause a publish failure here (I think?) if the workspace had pedantic packages.

In this situation, I assume setting `pedantic_packages` to true is desirable because we're publishing. Thoughts? Or should we be using the package's config here?

test/Prelude.purs
```diff
@@ -106,7 +106,9 @@ check checkers execResult = do
       Left err -> err.stderr
       Right res -> res.stderr
 
-  when false do
+  printStdoutStderr <- liftEffect $ map isJust $ Process.lookupEnv "SPT_DEBUG" -- spago test
+
+  when printStdoutStderr do
```
JordanMartinez, line 111: There's too many times I've wanted to enable this. So, I thought it'd be better if we used an env flag to do this. I chose `SPT_DEBUG` because it was shorter than `SPAGO_TEST_DEBUG`.

src/Spago/Command/Build.purs
```diff
@@ -135,23 +139,27 @@ run opts = do
...
-            Graph.toImportErrors selected <$> runSpago (Record.union { graph, selected } env) Graph.checkImports
-          unless (Array.null errors) do
-            die' errors
+      -- TODO: when the entire workspace is built, 
+      -- here we could go through all the workspace packages and run the check for each
+      -- The complication is that "dependencies" includes all the dependencies for all packages
+      errors <- map Array.fold $ for pedanticPkgs \(Tuple selected options) -> do
+        Graph.toImportErrors selected options <$> runSpago (Record.union { graph, selected } env) Graph.checkImports
+      unless (Array.null errors) do
+        die' errors
```
f-f, line 162: Here we are calling `purs graph` for each package in the project, which is slow. We should be calling it only once with all the packages/dependencies, then make sense of it after the fact.

This will require some shuffling in the Graph module, and I am doing a similar reshuffling in #1069, so might be worth getting that in first

test/Prelude.purs
f-f: Ah yes, brilliant. But let's go for `SPAGO_TEST_DEBUG` as "SPT" is a lil cryptic 😄 

This is also worth noting in the CONTRIBUTORS file

test/Spago/Build/Polyrepo.purs
```diff
@@ -387,7 +388,7 @@ spec = Spec.describe "polyrepo" do
       , spagoYaml: Init.defaultConfig' $ PackageOnly
           { name: mkPackageName "case-four-package-a"
           , dependencies: [ "console", "effect", "prelude" ]
-          , test: Just { moduleMain: "Subpackage.SameName.Test.Main", strict: Nothing, censorTestWarnings: Nothing }
+          , test: Just { moduleMain: "Subpackage.SameName.Test.Main", strict: Nothing, censorTestWarnings: Nothing, pedanticPackages: Nothing, dependencies: Nothing }
```
f-f, line 391: I wonder if it's worth defining a default config for these polyrepo tests that we override?

src/Spago/Command/Build.purs
JordanMartinez: Not quite. If you look a few lines earlier, you'll see a single purs graph call. I reworked this code when we did the topological sort code to separate the purs graph call from the error creation for a single package code.

test/Spago/Build/Polyrepo.purs
JordanMartinez: I definitely need to clean up some of these tests, but I'm not sure if that should be done in this PR.

test/Prelude.purs
JordanMartinez: Fixed.

src/Spago/Command/Build.purs
f-f: Oh I see, sorry for missing that - this is all good then!

src/Spago/Command/Publish.purs
f-f: No here it's good to set it to true - this made me spot another bug though: that `for_` there means that if we have no graph then we'd just skip the check. We shouldn't, and throw an error instead 😄

f-f: It's a two line fix so we can fix it here, unless you'd prefer otherwise

JordanMartinez: What should the error message say if it's not there?

f-f: I guess we could just print the error we get:
https://github.com/purescript/spago/blob/96c8846042251b26c4668297cde5ed037170c965/src/Spago/Purs.purs#L123-L135

f-f (approved):

test/Spago/Build/Polyrepo.purs
JordanMartinez: Opened an issue here: https://github.com/purescript/spago/issues/1079

f-f:
Could we fix it by swapping configs? I.e. compiling everything with a config that doesn't have `pedantic_packages` set, then building, then replacing it with one that has the setting

JordanMartinez:
> Could we fix it by swapping configs? I.e. compiling everything with a config that doesn't have `pedantic_packages` set, then building, then replacing it with one that has the setting

Yeah, that's the approach I'm taking.

PR #1076, opened 2023-10-13
Update run section in readme

### LINKED ISSUE #1050: `spago run` without specifying a package
I have a UX request, instead of needing to run:
`spago run -p foo -m Foo`

I would like to run:
`spago run -m Foo`

The compiled modules in `output` are unique, so specifying the package feels redundant as a user.

I do understand that, it's easier from spago's standpoint if it only has to build the necessary package to get the desired module. 
I'm of the opinion that the improved UX for simple cases, warrants allowing a full workspace build when the requested module hasn't been built yet.

JordanMartinez:
### Description of the change

Fixes #1050

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,9 +364,9 @@ Oof! That's a lot of typing. Fortunately it's possible to configure most of thes
 See [here](#the-configuration-file) for more info about this, but it allows us to instead write:
 
 ```console
-# module defined
-spago run -p package-name
-spago run -p --purs-args "--verbose-errors"
+# The main module can be defined in the configuration file, but
+# purs flags still need to be supplied at the command line
+spago run -p package-name --purs-args "--verbose-errors"
 
 # Args passed to program defined
 $ spago run -p package-name
--- a/README.md
+++ b/README.md
@@ -342,23 +342,34 @@ Multiple commands are possible - they will be run in the order specified:
 $ spago build --before clear --before "notify-send 'Building'"
 ```
 
-If you want to run the program (akin to `pulp run`), just use `run`:
+If you want to run the program (akin to `pulp run`), just use `run`. The commands below are verbose until these values are defined in the configuration file:
 ```console
-# The main module defaults to "Main"
-$ spago run
+$ spago run -p package-name -m Module.Containing.Main
 
-# Or define your own module path to Main
-$ spago run --main ModulePath.To.Main
+# We can pass arguments through to `purs compile`
+$ spago run -p package-name  -m Module.Containing.Main --purs-args "--verbose-errors"
 
-# And pass arguments through to `purs compile`
-$ spago run --main ModulePath.To.Main --purs-args "--verbose-errors"
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -348,23 +348,34 @@ Multiple commands are possible - they will be run in the order specified:
 $ spago build --before clear --before "notify-send 'Building'"
 ```
 
-If you want to run the program (akin to `pulp run`), just use `run`:
+If you want to run the program (akin to `pulp run`), just use `run`. The commands below are verbose until these values are defined in the configuration file:
```
f-f, line 351: I don't think there's a need to mention the verbosity since we'll explain that in a second
```suggestion
If you want to run the program, just use `run`:
```

README.md
```diff
@@ -348,23 +348,34 @@ Multiple commands are possible - they will be run in the order specified:
...
+$ spago run -p package-name  -m Module.Containing.Main --purs-args "--verbose-errors"
+
+# We can pass arguments through to the program being run
+$ spago run -p package-name  -m Module.Containing.Main -- arg1 arg2
+```
 
-# Or define your own module path to Main
-$ spago run --main ModulePath.To.Main
+Oof! That's a lot to run a command. If you configure these parameters in the package's configuration file's `package.run` section, you don't have to supply them at the command line.
+See [here](#the-configuration-file) for more info about this, but it allows us to instead write.
```
f-f, line 363: Some wording to make it smoother to read
```suggestion
Oof! That's a lot of typing. Fortunately it's possible to configure most of these parameters in the `package.run` section of your configuration file, so you don't have to supply them at the command line.

See [here](#the-configuration-file) for more info about this, but it allows us to instead write:
```

README.md
```diff
@@ -348,23 +348,34 @@ Multiple commands are possible - they will be run in the order specified:
...
 
-# And pass arguments through to `purs compile`
-$ spago run --main ModulePath.To.Main --purs-args "--verbose-errors"
+```console
+# module defined
+spago run -p package-name
+spago run -p --purs-args "--verbose-errors"
```
f-f, line 368: The second command here doesn't work, so I think you mean to write this
```suggestion
```console
# The main module can be defined in the configuration file, but
# purs flags still need to be supplied at the command line
spago run -p package-name --purs-args "--verbose-errors"
```

README.md
```diff
@@ -348,23 +348,34 @@ Multiple commands are possible - they will be run in the order specified:
...
 
-# Or pass arguments to the backend, in this case node
-$ spago run -- arg1 arg2
+# Args passed to program defined
+$ spago run -p package-name
```
f-f, line 371: Think it's useful to show an example of the config in this case
```suggestion
# It's possible to even pass arguments from the config, which would look like this:
#
# package:
#   run:
#       main: Main
#       execArgs:
#         - "arg1"
#         - "arg2"
$ spago run -p package-name
```

f-f (approved):

PR #1080, opened 2023-10-14
Spago docs --deps-only

### LINKED ISSUE #1020: Reinstate the `spago docs` command

CharlesTaylor7:
### Description of the change

Resolves #1020 

With this change I believe we're basically at parity with the old spago docs command. 

(There was a --no-search flag in the old spago which skips indexing, but I don't see a strong reason to bring that back)

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -938,7 +938,7 @@ For example to generate ctags for use in your editor:
 $ spago docs --format ctags
 ```
 
-Sometimes you'd like to pull up docs for dependencies even when you have compilation errors in your project. This good use case for the --deps-only flag:
+Sometimes you'd like to pull up docs for dependencies even when you have compilation errors in your project. This is a good use case for the --deps-only flag:
 
 ```console
 $ spago docs --deps-only`
--- a/README.md
+++ b/README.md
@@ -932,11 +932,18 @@ If you wish for the documentation to be opened in browser when generated, you ca
 $ spago docs --open
 ```
 
-To build the documentation as Markdown instead of HTML, or to generate tags for your project,
-you can pass a `format` flag:
+You can customize the output to other formats beyond html. Supported formats include ctags, etags, and markdown.
+For example to generate ctags for use in your editor:
 ```console
 $ spago docs --format ctags
 ```
+
+Sometimes you'd like to pull up docs for dependencies even when you have compilation errors in your project. This good use case for the --deps-only flag:
+
+```console
+$ spago docs --deps-only`
+```
+
 
 ### Alternate backends
 
--- a/bin/src/Flags.purs
+++ b/bin/src/Flags.purs
@@ -264,3 +264,11 @@ sourceMaps =
     ( O.long "source-maps"
         <> O.help "Creates a source map for your bundle"
     )
+
+depsOnly :: Parser Boolean
+depsOnly =
+  O.switch
+    ( O.long "deps-only"
+        <> O.help "Build depedencies only"
+    )
+
--- a/bin/src/Main.purs
+++ b/bin/src/Main.purs
@@ -363,8 +363,7 @@ publishArgsParser =
 
 docsArgsParser :: Parser DocsArgs
 docsArgsParser = Optparse.fromRecord
-  -- TODO: --deps-only
-  { depsOnly: pure false :: Parser Boolean
+  { depsOnly: Flags.depsOnly
   , open: O.switch
       ( O.long "open"
           <> O.short 'o'


CharlesTaylor7:
Technically the feature works, but the console outs give the impression it's building the project, even when isn't:

 ```console
 spago docs --deps-only
Read the package set from the registry

✅ Selecting 6 packages to build:
    docs-search-client-halogen
    docs-search-common
    docs-search-index
    spago
    spago-bin
    spago-core

Writing a new lockfile
Downloading dependencies...
Generating documentation for the project. This might take a while...
Documentation written to: generated-docs/html
Building the search index...
"bower.json" decoding failed for .spago/packages/avar-5.0.0/bower.json: An error occurred while decoding a JSON value:
  Expected value of type 'ROOT."main" - Expected a value of type Array but got String'.
Indexing 1004 modules from 127 packages...
Added 8064 definitions and 9636 type definitions from 127 packages to the search index.
Link: file:///Users/chuck/repos/spago/generated-docs/html/index.html
```

**Comment by CharlesTaylor7** [comment]
spago install also includes this extra terminal output:

```console
spago install
Read the package set from the registry

✅ Selecting 6 packages to build:
    docs-search-client-halogen
    docs-search-common
    docs-search-index
    spago
    spago-bin
    spago-core

Writing a new lockfile
Downloading dependencies...
Building...
           Src   Lib   All
Warnings     0     0     0
Errors       0     0     0

✅ Build succeeded.

```

README.md
```diff
@@ -932,12 +932,19 @@ If you wish for the documentation to be opened in browser when generated, you ca
...
 
-To build the documentation as Markdown instead of HTML, or to generate tags for your project,
-you can pass a `format` flag:
+You can customize the output to other formats beyond html. Supported formats include ctags, etags, and markdown.
+For example to generate ctags for use in your editor:
 ```console
 $ spago docs --format ctags
 ```
 
+Sometimes you'd like to pull up docs for dependencies even when you have compilation errors in your project. This good use case for the --deps-only flag:
```
f-f, line 941:
```suggestion
Sometimes you'd like to pull up docs for dependencies even when you have compilation errors in your project - in such cases it's possible to use the --deps-only flag:
```

f-f:
> There was a --no-search flag in the old spago which skips indexing, but I don't see a strong reason to bring that back

Agreed, don't think it's useful since indexing is pretty quick.

> the console output give the impression it's building the project, even when isn't

If it's not building the project then we should fix that, because we need to make sure the project is build before generating the index, otherwise the docs files will be out of date

> spago install also includes this extra terminal output

What's extra about it? Looks like regular business to me. If you're referring to the fact that it's building the project then sure, `spago install` does that (it's basically a `spago build --deps-only`)

CharlesTaylor7:
> > the console output give the impression it's building the project, even when isn't
> 
> If it's not building the project then we should fix that, because we need to make sure the project is build before generating the index, otherwise the docs files will be out of date

No, I think its working as intended. 
 I just mean that it's building the dependencies of the projects, not the user's local code. To me, the message of "selecting 6 packages to build" sounds looks like its building the entire workspace. There's no indication it's going to stop short of building the application code until it stops.


> > spago install also includes this extra terminal output
> 
> What's extra about it? Looks like regular business to me. If you're referring to the fact that it's building the project then sure, `spago install` does that (it's basically a `spago build --deps-only`)


This is a good analogy. I'm thinking in both these cases it could alternately say something like "Selecting n packages to build (dependencies only)".

README.md
CharlesTaylor7, line 941:
```suggestion
Sometimes you'd like to pull up docs for dependencies even when you have compilation errors in your project. This is a good use case for the --deps-only flag:
```

f-f:
> I just mean that it's building the dependencies of the projects, not the user's local code. To me, the message of "selecting 6 packages to build" sounds looks like its building the entire workspace. There's no indication it's going to stop short of building the application code until it stops.

Ah, I see what you mean now. I think the point of saying that is to note to the user 'look, we found these packages in the workspace for our build', which makes sense in this case too, since we're kicking off a build - even if it's just to build the dependencies of the workspace, because we are going to figure them out from these packages.
So not sure how we could frase this any better

f-f (approved):

PR #1084, opened 2023-10-14
Cleanup code and tests in Polyrepo tests

### LINKED ISSUE #1079: Cleanup Polyrepo test code
There are various places where this code could be cleaned up. The main goal of this issue is to reduce unneeded verbosity/duplication/redundancy where possible.

We need a different framework than what's currently used in the code to
1. add various things to the package's configuration
2. make packages that 
    1. have unique src and test module names
    2. can be depended on in other packages' code

See https://github.com/purescript/spago/pull/1075#discussion_r1357889757 for context.

JordanMartinez:
### Description of the change

Fixes #1079 

### Checklist:


test/Spago/Build/Polyrepo.purs
```diff
@@ -68,97 +71,189 @@ spec = Spec.describe "polyrepo" do
...
-              <>
-                ( packageName # maybe [] \pkgName ->
-                    [ ""
-                    , "packageName :: String"
-                    , "packageName = \"" <> pkgName <> "\""
-                    ]
-                )
-        }
+    mkDependencies :: Array String -> Config.Dependencies
+    mkDependencies = Config.Dependencies <<< Map.fromFoldable <<< map (flip Tuple Nothing <<< mkPackageName)
```
JordanMartinez, line 75: Maybe this should go in `Test/Prelude.purs`?

test/Spago/Build/Polyrepo.purs
```diff
@@ -68,97 +71,189 @@ spec = Spec.describe "polyrepo" do
...
+      Init.defaultConfig' $ PackageOnly $ snd finalConfig
+
+    configAddSrcStrict :: Tuple String Init.DefaultConfigPackageOptions -> Tuple String Init.DefaultConfigPackageOptions
+    configAddSrcStrict = map \r -> r
+      { build = Just
+          { strict: Just true
+          , censorProjectWarnings: r.build >>= _.censorProjectWarnings
+          , pedanticPackages: r.build >>= _.pedanticPackages
+          }
+      }
```
JordanMartinez, line 101: Should these functions go in `Init/Config.purs`?

test/Spago/Build/Polyrepo.purs
f-f: Absolutely!

test/Spago/Build/Polyrepo.purs
f-f: We are using these in tests only so I don't think they should be in the project source tree; at best we might be able to reuse them in other tests outside of the polyrepo stuff, so we can move them to the Test.Prelude or some other common module if we find refactoring opportunities in other tests

f-f (approved):
This is nice 👏

PR #1085, opened 2023-10-14
Halogen dev workflow, with esbuild

CharlesTaylor7:
@f-f Alternative to #1072, based on your feedback: https://github.com/purescript/spago/pull/1072#issuecomment-1762822050

This version avoid pulling in `vite`, and just bundles with ESbuild in watch mode.
Overall, it's probably less maintenance overhead, but you will have to manually refresh the page index.html file after the bundling completes.


package.json
```diff
@@ -30,7 +30,8 @@
     "format": "purs-tidy format-in-place core src bin docs-search",
     "format:check": "purs-tidy check core src bin docs-search",
     "bundle": "spago bundle -p spago-bin",
-    "prepublishOnly": "spago build && ./bin/index.dev.js bundle -p spago-bin && ./bin/index.dev.js bundle -p docs-search-client-halogen"
+    "prepublishOnly": "spago build && ./bin/index.dev.js bundle -p spago-bin && ./bin/index.dev.js bundle -p docs-search-client-halogen",
+    "halogen:dev": "echo \"#!/usr/bin/env node\n\nimport { main } from './output/Docs.Search.App/index.js'; main();\" | esbuild --platform=browser --format=iife --sourcemap --bundle --outfile=generated-docs/html/docs-search-app.js --loader:.node=file --alias:punycode=punycode/ --watch=forever"
```
CharlesTaylor7, line 34: A few notes:
- `spago bundle` only allows `extraArgs` in the spago.yaml configs, not from the command line. So I was forced to use esbuild directly. 
- Esbuild doesn't know its supposed to invoke the `main` function so I copied the same workarounds from `Spago.Command.Docs.` i.e. it needs the --iife flag, and the main function to be invoked via extra stdin input.
- --watch=forever is so that it keeps watching even after stdin stops receiving input. This does the right thing ™️ , it rebuilds when output/ dependencies change. If you run in regular watch mode, you get this error:

```console
[watch] build finished, watching for changes...
[watch] stopped because stdin was closed (use "--watch=forever" to keep watching even after stdin is closed)
```

f-f: > only allows `extraArgs` in the spago.yaml configs, not from the command line

Oh, that looks like an oversight 😄 I think we have a good argument to add this now

f-f:
Windows CI doesn't look happy, but I like the overall looks of this much better. I think we can fix the bundle extraArgs here, so we can use a `bundle` command instead of directly calling esbuild

CharlesTaylor7:
@f-f On Windows, where does spago bundle output go?

The CI is saying that neither `bin/bundle.js` nor `bin/docs-search-app.js` is present even right after calling the bundle commands:

https://github.com/purescript/spago/actions/runs/6541347144/job/17762743250?pr=1085#step:13:2

f-f:
Yeah, that's very confusing - I'll have a look

f-f:
I can't push to your branch (for future PRs you could make branches in this reoo so we can all push to them) but I would try to `ls` the current folder as well

package.json
JordanMartinez: > it needs the --iife flag

Why does it need that?

f-f: > Why does it need that?

To bundle for the browser, [see esbuild docs](https://esbuild.github.io/api/#platform)

JordanMartinez: :thinking: Wasn't there some other reason though about why we couldn't use ESM specifically? I thought there was some library that had an issue, but I can't recall which or why.

CharlesTaylor7:
@f-f Hey, I'm back and trying to debug this.

I found something weird. On Windows only, If you look at the last two workflow runs, you can see the bundle step works when using the installed `spago` package, but does nothing when using the local `./bin/index.dev.js`.

Both are bundling in verbose mode, so there should be lots of console output:

✅ Working:
https://github.com/purescript/spago/actions/runs/6834906185/job/18588114472?pr=1085#step:12:1

❌ Failing with very little console output:
https://github.com/purescript/spago/actions/runs/6834832975/job/18587969681#step:12:1

f-f:
> you can see the bundle step works when using the globally installed spago package, but does nothing when using the freshly built ./bin/index.dev.js

Oh! Of course. Calling an executable to be run with `./some-program` will use the shebang definition on Linux/macOS (which we do include when bundling), but does not work on Windows. We should switch that call from `./bin/index.dev.js` to `node ./bin/index.dev.js`

bin/src/Flags.purs
```diff
@@ -203,6 +203,15 @@ backendArgs =
...
 
+esbuildArgs :: Parser (List String)
+esbuildArgs =
+  O.many $
+    O.strOption
+      ( O.long "esbuild-args"
+          <> O.help "Arguments to pass to ESBuild"
+          <> O.metavar "ARGS"
+      )
+
```
CharlesTaylor7, line 214: @f-f Thoughts on the flag name? 
Extra purs args are under --purs-args. 

But the workflow config presently has a generic field `extra_args` instead of `esbuild_args` for the field.

f-f: Let's go with a more generic `bundler-args`

CharlesTaylor7: Doesn't that imply some false degree of generality?

As soon as a spago user starts relying on this flag to pass extra args, they are relying on the esbuild API. Hypothetically, if spago ever switches from bundling with esbuild to something else, then this user's use case is broken. We might as well name the flag after the underlying tool for clarity. Given we can't actually safely change the bundler without breaking assumptions. 

 `--purs-args` also sets a bit of a precendent, vs. calling it `--compiler-args`.

bin/src/Main.purs
```diff
@@ -719,7 +721,8 @@ mkBundleEnv bundleArgs = do
       ( (Config.parseBundleType =<< bundleArgs.type)
           <|> bundleConf _.type
       )
-  let bundleOptions = { minify, module: entrypoint, outfile, platform, type: bundleType, sourceMaps: bundleArgs.sourceMaps, extraArgs: fromMaybe [] (bundleConf _.extra_args) }
+  let extraArgs = Array.fromFoldable bundleArgs.extraArgs <> fromMaybe [] (bundleConf _.extra_args)
```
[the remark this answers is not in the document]
CharlesTaylor7: Fair enough, everything else on the CLI overrides the config file right?

bin/src/Flags.purs
f-f: I'm thinking of alternate backends when I propose this: I'd like to make `spago bundle` call out to `$BACKEND bundle` if a backend is specified, in which case a esbuild-specific flag is not truthful anymore.

We had a similar situation with the old spago, where we first had `node-args`, then we added `backend-args` and the two did the same thing and it was confusing 😄

bin/src/Main.purs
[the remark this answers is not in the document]
f-f: As far as I know yes!

f-f:
This looks great now - thanks @CharlesTaylor7! 👏👏

f-f (approved):

PR #1092, opened 2023-10-16
Remove src deps from test deps on `--pedantic-packages` report

### LINKED ISSUE #1043: For pedantic builds, limit install recommendations for test deps
https://github.com/purescript/spago/pull/1041#discussion_r1345795122

When you run `spago build -p spago --pedantic-packages` you might see:

```console
Run the following command to install them all:
  spago install -p spago aff-promise ansi console control enums exceptions formatters node-fs-aff routing-duplex unfoldable unicode
  
  ...
  
  Run the following command to install them all:
  spago install --test-deps -p spago console enums identity newtype node-fs-aff quickcheck
```

But we'd prefer it to be something like:
```console
Run the following command to install them all:
  spago install -p spago aff-promise ansi console control enums exceptions formatters node-fs-aff routing-duplex unfoldable unicode
  
  ...
  
  Run the following command to install them all:
  spago install --test-deps -p spago identity newtype quickcheck
```
Basically the duplicated recommendations between the missing project deps,  and the missing test deps, have been deduplicated.

JordanMartinez:
### Description of the change

Fixes #1043

In `checkImports`, moved the main code into a where clause so that I could use the source project's `used` deps rather than their declared ones.

Updated the various tests to pass while adding a new test for pedantic packages that's more exhaustive.

### Checklist:


test/Spago/Build.purs
```diff
@@ -102,36 +104,40 @@ spec = Spec.around withTempDir do
...
             spago [ "build" ] >>= shouldBeSuccess
 
         Spec.it "passing --pedantic-packages CLI flag" \{ spago, fixture } -> do
-          setupSrcUnusedDeps spago
+          setupSrcUnusedDeps spago true
           spago [ "build", "--pedantic-packages" ] >>= shouldBeFailureErr (fixture "check-unused-dependency.txt")
 
         Spec.it "package config has 'pedantic_packages: true'" \{ spago, fixture } -> do
-          setupSrcUnusedDeps spago
+          setupSrcUnusedDeps spago false
```
f-f, line 140:
```suggestion
          setupSrcUnusedDeps spago true
```

I'd love to see a test that has these packages in test dependencies as well, to make sure we do not remove them unconditionally from the suggestion that we give out

JordanMartinez: I didn't quite understand what was meant here, but I think latest commits addresses this point.

f-f: Ah sorry, my comment was misguided. I meant to say that if we happen to have a package - let's say `effect` - both in the test and on the source dependencies, Spago should note both

JordanMartinez: Ah, ok.
Hm... but that test doesn't pass :disappointed:  :smile:

JordanMartinez: Actually, it does pass. I just needed to install the dependencies in `test` as well.

JordanMartinez: See latest commit.

test/Spago/Build.purs
```diff
@@ -99,65 +101,212 @@ spec = Spec.around withTempDir do
...
+                , "import Effect.Console as Console"
+                , "import Control.Alt as Alt"
+                , "main = unit"
+                ]
+              -- get rid of "Compiling ..." messages and other compiler warnings
+              spago [ "build" ] >>= shouldBeSuccess
+
+          Spec.it "passed --pedantic-packages CLI flag" \{ spago, fixture } -> do
+            setupSrcTransitiveTests spago true
+            spago [ "build", "--pedantic-packages" ] >>= shouldBeFailureErr (fixture "check-direct-import-transitive-dependency.txt")
```
f-f, line 142: I'd expect this to complain about the test packages containing superfluous `console` and `effect` (since they are already in the source dependencies) - the flag should turn it on for both src and test targets, since it's a one-off, i.e. the user is explicitly asking for a cleanup pass

test/Spago/Build.purs
```diff
@@ -99,65 +101,212 @@ spec = Spec.around withTempDir do
...
+            setupTestTransitiveTests spago
+            addPedanticPackagesToTest
+            spago [ "build" ] >>= shouldBeFailureErr (fixture "check-direct-import-transitive-dependency-test.txt")
+
+      Spec.describe "fails with warnings about unused dependencies" do
+
+        Spec.describe "in a source package when" do
+
+          let
+            setupSrcUnusedDeps spago installConsoleAndEffect = do
```
f-f, line 189: These names are a bit confusing - maybe a record will help with conveying intent while reading through?
```suggestion
            setupUnusedDepsInSrc spago { installConsoleAndEffectInTest } = do
```

test/Spago/Build.purs
```diff
@@ -99,65 +101,212 @@ spec = Spec.around withTempDir do
...
+              spago [ "init", "--name", "7368613235362d2f444a2b4f56375435646a59726b53586548" ] >>= shouldBeSuccess
+              when installConsoleAndEffect do
+                spago [ "install", "--test-deps", "console", "effect" ] >>= shouldBeSuccess
+              FS.writeTextFile (Path.concat [ "src", "Main.purs" ]) "module Main where\nimport Prelude\nmain = unit"
```
f-f, line 193: I think we could do with a `writeMain :: Array String -> Effect Unit` helper (maybe a `writeTestMain` too?)

test/Spago/Build.purs
```diff
@@ -99,65 +101,212 @@ spec = Spec.around withTempDir do
...
+            spago [ "build", "--pedantic-packages" ] >>= shouldBeFailureErr (fixture "check-unused-dependency.txt")
 
-        Spec.it "package config has 'pedantic_packages: true'" \{ spago, fixture } -> do
-          setupSrcUnusedDeps spago
-          addPedanticPackagesToSrc
-          spago [ "build" ] >>= shouldBeFailureErr (fixture "check-unused-dependency.txt")
+          Spec.it "package config has 'pedantic_packages: true'" \{ spago, fixture } -> do
+            setupSrcUnusedDeps spago false
+            addPedanticPackagesToSrc
+            spago [ "build" ] >>= shouldBeFailureErr (fixture "check-unused-dependency.txt")
```
f-f, line 204: I think we have another issue here:
- the source dependencies have `console` and `effect`
- ...but the sources don't use them
- the test dependencies don't have them
- ...but the test sources _do_ use them

We correctly complain about them being unused (because we have the flag set to true in the package), but since the flag is not set in tests we don't say anything there. 
However removing them would make the build fail, since they are used in tests. Not sure how to disentangle this one.

test/Spago/Build.purs
```diff
@@ -99,65 +101,212 @@ spec = Spec.around withTempDir do
...
+
+        Spec.describe "in a test package when" do
 
-        Spec.it "package's test config has 'pedantic_packages: true' and test code has unused dependencies" \{ spago, fixture } -> do
-          spago [ "init", "--name", "7368613235362d2f444a2b4f56375435646a59726b53586548" ] >>= shouldBeSuccess
-          FS.writeTextFile (Path.concat [ "test", "Test", "Main.purs" ]) "module Test.Main where\nimport Prelude\nmain = unit"
           let
-            modifyPkgTestConfig pedantic = modifyPackageConfig \config ->
+            setupTestUnusedDeps spago = do
```
f-f, line 209: Again trying to make this smoother to read through..
```suggestion
            setupUnusedDepsInTest spago = do
```

f-f (commented):
Didn't go through the implementation yet, just trying to make sense of the tests for now. There's something about how the tests are setup that adds indirection - not sure if we need more abstraction or less there, will sleep on it

test/Spago/Build.purs
JordanMartinez: It did. That's why the `true` arg in `setupSrcTransitiveTests spago true` installs those deps in `test`. Otherwise, I'll get an error about that. In this particular group of tests, we're testing whether source packages alone get that error, which I can't do since `--pedantic-packages` enables that check on the `test` code as well.

test/Spago/Build.purs
JordanMartinez: The options forward are:
1. Stop making `test` depend on all the packages declared by the source package implicitly. Rather, make it explicit. A `test` package only implicitly depends on its source package; that's it. While this simplifies this particular situation, it complicates `spago install` and `spago uninstall` situations. But hey, isn't that what the `--pedantic-packages=autofix` idea is for?
2. Add a `purs graph` call that allows one to see which dependencies in the source package are actually used and have the test package only implicitly depend on those used dependencies. While this resolves this issue, it'll be confusing because a `spago uninstall` can make a future `spago build` not work as (correct me if I'm wrong) `spago build -p foo` build both the source and test packages of `foo`.
3. Add another check for this situation. If source declares unused deps that tests does not declare but which uses, then include it as an error. Since we're already going to be reporting that the source package has unused dependencies and thereby failing the build, we may also state the above issue.

JordanMartinez: Any new thoughts on this? Or are you still thinking about it?

f-f: I'm still thinking about the overall shape of this PR, but I think I also prefer option 3 here

test/Spago/Build.purs
f-f: I think my confusion comes from the fact that it's hard to see what the test is trying to verify - so I suggest we:
- do not generate the outputs via code, but only use fixtures
- this means that we'll have a lot of fixtures, so let's stick them in a separate folder, like `test-fixtures/pedantic`
- then before each test case let's have a few lines summarising why the test should return what it returns

test/Spago/Build.purs
f-f: Yeah, I think option 1 is a possibility but definitely uncomfortable, so I'd like to reach for it only in case we can't make anything else work.

As for Option 2, I am not sure how it resolves the issue - I think `spago uninstall` is supposed to be able to break the build, so that part is fine. The test package already implicitly depends on the packages, the issue is that we're not detecting if we are slating these implicit dependencies for removal.

Option 3 makes sense, and I'd go even further in the source+test packages integration, and add another kind of "error" that we return - on top of "unused" and "undeclared transitive" - which is "unused in source but used in test" (I'm sure we can find a more compact naming 😄). So in the same graph check we'd take the unused packages for the source package, and see if they'd be used in the test package. If yes, we add `spago install --test-deps` instructions for them, which should counterbalance the `spago uninstall` ones.

test/Spago/Build.purs
JordanMartinez: That's... a lot of work. Is there an alternative for making these tests clearer such as comments?

f-f: We can merge and I can do the work if you'd prefer

f-f: I can also pile commits on this branch if you'd like a self-contained package

JordanMartinez: I'm fine with you pushing commits here.

JordanMartinez:
Hoping to get to this sometime this week, but next week might be more likely.

JordanMartinez:
> There's one failing test now (together with a `TODO` in the code), for a case that I think we should catch, related to the source-and-test-dependencies interplay that we discussed above

Could you clarify what the issue here is? It's been a while.

f-f:
> Could you clarify what the issue here is? It's been a while.

Uh, it turns out the implementation is already doing what we think it should do and the test was failing for a bad fixture. I'll sort out the merge conflict and merge this in..

Issue #1094, opened 2023-10-18
Slow to start compilation when many packages are in the workspace


finnhodgkin:
I'm getting long pauses before compilation starts on projects with many workspace packages - above 100 - or if there are many nested dependencies. I'm guessing most projects won't reach the size where this starts to bite, but I'm getting waits of 5-10 seconds on a project with 50-100 packages.

I've created a [minimal reproduction repo](https://github.com/finnhodgkin/spago-dependency-tester) to see if I could spot any obviously lagging parts. The main bottleneck appears to be the transitive dependency fetching, specifically the many `Map.union` calls used to combine the different package deps. Swapping the unions to a single map that's updated less seems to speed things up nicely.

Setup | On master | After tweak
-- | -- | --
500 packages, each dependent on the last | 3m 44s | 19s
200 packages, each dependent on the last | 12s | 1.3s
100 packages, each dependent on the first 10 packages | 17s | 0.8s

All run on an m1 macbook with a very unscientifc timings setup... 

PR to follow that'll probably have a bunch of really obvious gotchas but is working locally and gets through the tests - thought I may as well 🤷

f-f:
Amazing investigation - this is exactly the stress testing that we need before putting this in the hands of everyone. Appreciate this a lot ❤️

f-f:
Fixed in #1095

PR #1095, opened 2023-10-18
Remove expensive union calls during dependency fetching

finnhodgkin:
### Description of the change

#1094

I think any test I add for this is likely to be super flaky because it would be about timings. I'm sure I've missed something obvious 👀 

### Checklist:

### DIFF
--- a/src/Spago/Command/Fetch.purs
+++ b/src/Spago/Command/Fetch.purs
@@ -404,11 +404,6 @@ getTransitiveDepsFromPackageSet packageSet deps = do
     init :: TransitiveDepsResult
     init = { packages: (Map.empty :: Map PackageName Package), errors: mempty }
 
-    -- | Cheap size check to be used in place of Set equality
-    errorSize :: TransitiveDepsResult -> Int
-    errorSize { errors: { notInPackageSet: p, notInIndex: i, cycle: c } } =
-      Set.size p + Set.size i + Set.size c
-
     go :: Set PackageName -> PackageName -> StateT TransitiveDepsResult (Spago (FetchEnv a)) Unit
     go seen dep = do
       -- We stash packages that we encountered along the way in `seen`,
@@ -416,7 +411,7 @@ getTransitiveDepsFromPackageSet packageSet deps = do
       if Set.member dep seen then do
         State.modify_ $ cycleError dep
       else do
-        -- If the package is transitive dependency of some other package that
+        -- If the package is a transitive dependency of some other package that
         -- we already met, then we don't need to do the work again
         alreadyRun <- Map.member dep <$> State.gets _.packages
         when (not alreadyRun)
@@ -431,14 +426,16 @@ getTransitiveDepsFromPackageSet packageSet deps = do
                 Nothing -> State.modify_ $ notInIndexError dep
                 Just dependenciesMap -> do
                   -- Compare errors before and after recursively running transitive deps
-                  errors <- errorSize <$> State.get
+                  errors <- State.gets _.errors
 
                   -- recur here, as we need to get the transitive tree, not just the first level
                   void $ forWithIndex dependenciesMap
                     (\dependency _ -> go (Set.insert dep seen) dependency)
 
-                  errorsAfterTransitiveDeps <- errorSize <$> State.get
+                  -- Errors may have changed after running through the child deps
+                  errorsAfterTransitiveDeps <- State.gets _.errors
 
+                  -- Do not include the package if any child deps fail
                   when (errors == errorsAfterTransitiveDeps) do
                     State.modify_ \st -> st { packages = Map.insert dep package st.packages }
 
--- a/src/Spago/Command/Fetch.purs
+++ b/src/Spago/Command/Fetch.purs
@@ -404,42 +404,41 @@ getTransitiveDepsFromPackageSet packageSet deps = do
     init :: TransitiveDepsResult
     init = { packages: (Map.empty :: Map PackageName Package), errors: mempty }
 
-    mergeResults :: TransitiveDepsResult -> TransitiveDepsResult -> TransitiveDepsResult
-    mergeResults r1 r2 =
-      { packages: Map.union r1.packages r2.packages
-      , errors: r1.errors <> r2.errors
-      }
+    -- | Cheap size check to be used in place of Set equality
+    errorSize :: TransitiveDepsResult -> Int
+    errorSize { errors: { notInPackageSet: p, notInIndex: i, cycle: c } } =
+      Set.size p + Set.size i + Set.size c
 
-    go :: Set PackageName -> PackageName -> StateT (Map PackageName (Map PackageName Package)) (Spago (FetchEnv a)) TransitiveDepsResult
-    go seen dep =
-      if (Set.member dep seen) then do
-        pure (init { errors { cycle = Set.singleton dep } })
+    go :: Set PackageName -> PackageName -> StateT TransitiveDepsResult (Spago (FetchEnv a)) Unit
+    go seen dep = do
+      if Set.member dep seen then do
+        State.modify_ $ cycleError dep
       else do
-        cache <- State.get
-        case Map.lookup dep cache of
-          Just allDeps ->
-            pure (init { packages = allDeps })
-          Nothing ->
-            -- First look for the package in the set to get a version number out,
-            -- then use that version to look it up in the index and get the dependencies
-            case Map.lookup dep packageSet of
-              Nothing -> pure (init { errors { notInPackageSet = Set.singleton dep } })
-              Just package -> do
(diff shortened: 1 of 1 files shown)


f-f:
> I think any test I add for this is likely to be super flaky because it would be about timings

The only two requirements we have is that it keeps being correct and not slower 😄

src/Spago/Command/Fetch.purs
```diff
@@ -404,42 +404,41 @@ getTransitiveDepsFromPackageSet packageSet deps = do
...
-      }
-
-    go :: Set PackageName -> PackageName -> StateT (Map PackageName (Map PackageName Package)) (Spago (FetchEnv a)) TransitiveDepsResult
-    go seen dep =
-      if (Set.member dep seen) then do
-        pure (init { errors { cycle = Set.singleton dep } })
+    -- | Cheap size check to be used in place of Set equality
+    errorSize :: TransitiveDepsResult -> Int
+    errorSize { errors: { notInPackageSet: p, notInIndex: i, cycle: c } } =
+      Set.size p + Set.size i + Set.size c
```
f-f, line 410: Could you expand (in the docstring) on the reasoning for why this exists?

src/Spago/Command/Fetch.purs
```diff
@@ -404,42 +404,41 @@ getTransitiveDepsFromPackageSet packageSet deps = do
...
+
+    go :: Set PackageName -> PackageName -> StateT TransitiveDepsResult (Spago (FetchEnv a)) Unit
+    go seen dep = do
+      if Set.member dep seen then do
```
f-f, line 414: I think this little chunk of code will benefit a lot from proficuous commenting
```suggestion
      -- We stash packages that we encountered along the way in `seen`,
      -- so if we see it again we have a cycle
      if Set.member dep seen then do
```

src/Spago/Command/Fetch.purs
```diff
@@ -404,42 +404,41 @@ getTransitiveDepsFromPackageSet packageSet deps = do
...
-                  Just dependenciesMap -> do
-                    -- recur here, as we need to get the transitive tree, not just the first level
-                    { packages: childDeps, errors } <-
-                      for (Map.toUnfoldable dependenciesMap :: Array (Tuple PackageName Range)) (\(Tuple d _) -> go (Set.insert dep seen) d)
-                        >>= (pure <<< foldl mergeResults init)
-                    let allDeps = Map.insert dep package childDeps
-                    when (Set.isEmpty (errors.cycle <> errors.notInIndex <> errors.notInPackageSet)) do
-                      State.modify_ $ Map.insert dep allDeps
-                    pure { packages: allDeps, errors }
+        alreadyRun <- Map.member dep <$> State.gets _.packages
```
f-f, line 417: (I am adding comments as I read through this, feel free to correct any misunderstandings I have)
```suggestion
        -- If the package is transitive dependency of some other package that
        -- we already met, then we don't need to do the work again
        alreadyRun <- Map.member dep <$> State.gets _.packages
```

src/Spago/Command/Fetch.purs
```diff
@@ -404,42 +404,41 @@ getTransitiveDepsFromPackageSet packageSet deps = do
...
+        when (not alreadyRun)
+          case Map.lookup dep packageSet of
```
f-f, line 419:
```suggestion
          -- If we need to compute the dependencies from scratch instead, we first look
          -- in the package set to get a version number out, then use that version to
          -- look it up in the index and get the dependencies
          case Map.lookup dep packageSet of
```

src/Spago/Command/Fetch.purs
```diff
@@ -404,42 +404,41 @@ getTransitiveDepsFromPackageSet packageSet deps = do
...
+            Nothing -> State.modify_ $ notInPackageSetError dep
+            Just package -> do
+              maybeDeps <- State.lift $ memoisedGetPackageDependencies dep package
+              case maybeDeps of
+                Nothing -> State.modify_ $ notInIndexError dep
+                Just dependenciesMap -> do
+                  -- Compare errors before and after recursively running transitive deps
```
f-f, line 426: I think this is the bit I'm struggling to get - we are comparing the amount of errors because if we don't add any errors then that means that the subtree is "clean" (i.e. error-free) and we can move forward?

f-f (commented):
I left a few comments, but to sum this up: my understanding of this is that we are basically replacing the `Map.union` with a better usage of the `StateT` - does this make sense?

src/Spago/Command/Fetch.purs
finnhodgkin: Yeah exactly that. I felt kinda gross writing it because it's so much like classic mutation.

src/Spago/Command/Fetch.purs
finnhodgkin: On second thought this is almost definitely preemptive optimisation. Probably actually makes sense to just check equality of the whole errors record. 

That way if new errors are added in the future there won't be a worry about having to update this. That or use heterogeneous to get the size of all error `Set`s in the record (I just didn't want to add an extra dep to spago).

finnhodgkin:
> my understanding of this is that we are basically replacing the Map.union with a better usage of the StateT - does this make sense?

Well put :)

src/Spago/Command/Fetch.purs
finnhodgkin: I'll add a comment or replace with equality this evening :+1:

src/Spago/Command/Fetch.purs
finnhodgkin: > when (Set.isEmpty (errors.cycle <> errors.notInIndex <> errors.notInPackageSet)) do

That was the equivalent before ☝️

JordanMartinez:
LGTM.

@f-f It might be helpful to add a benchmark using a similarly-large repo that enables everything (e.g. warning censoring, pedantic packages, etc.) and then seeing how long it takes to run `spago build` and `spago build -p package-that-depends-on-all-others`. It doesn't necessarily need to be checked all the time, but it may be helpful to see how much of a change in performance a PR touching internals causes.

f-f:
Eh, I have the feeling that if CI is not checking it we'd probably forget to run it when it matters 😄

We had the same exact issue as this one in the Haskell implementation, and it was quickly rewritten so be fast because it was just annoying. (my point here being that bottlenecks like this will be evident quickly - e.g. I was already aware this bit of code was slow but was planning to route around it with #917)

We could add just a normal test that builds 50 packages just to make sure it won't get too slow, but I wouldn't bother about that even, since after we ship #917 you'd run this code only when generating the lock file rather than on every run

src/Spago/Command/Fetch.purs
f-f: Let's try with Set equality and see how slow it gets.

If that doesn't work then my next favourite option would be to replace the record that sorts errors in different buckets with just a List/Array/Set of `TransitiveDepsError` (which would be a sum type of all the errors we could get). Then we could check for the size of that.

src/Spago/Command/Fetch.purs
```diff
@@ -404,42 +404,45 @@ getTransitiveDepsFromPackageSet packageSet deps = do
...
+                  errors <- State.gets _.errors
+
+                  -- recur here, as we need to get the transitive tree, not just the first level
+                  void $ forWithIndex dependenciesMap
+                    (\dependency _ -> go (Set.insert dep seen) dependency)
+
+                  -- Errors may have changed after running through the child deps
+                  errorsAfterTransitiveDeps <- State.gets _.errors
+
+                  -- Do not include the package if any child deps fail
```
f-f, line 438: This makes me wonder: are we repeating work unnecessarily? If we have errors while resolving dependencies for the package A then we don't include it in the state, but then the next time (in another dependency branch) that we encounter it then we will try to resolve these again?

finnhodgkin: It's necessary for the circular dependencies test. The expectation in the test is for both package a and package b to be included in the error, without this condition only one of them is.

f-f: I think I was referring to something else, but I reread the implementation a couple more times and I'm convinced it is correct 😄

finnhodgkin:
Another chunk of time can be shaved off the `readWorkspace` step by parallelising the git ignore check. 

In my tests it moved a build from 3s to 2s, but it's way less impactful than the deps change. Want me to pop that it into this PR as, open another or just ignore it for now?

f-f (approved):
Thanks @finnhodgkin! This is fantastic 🙂

f-f:
> In my tests it moved a build from 3s to 2s, but it's way less impactful than the deps change. Want me to pop that it into this PR as, open another or just ignore it for now?

Oh I'm intrigued, didn't occur to me to parralelise that one! I think we can merge this as is, so new PR it is 🙂

PR #1098, opened 2023-10-19
Add `ls path`

### LINKED ISSUE #1082: Add `spago ls paths` command
We should add a command to display the [common paths used by Spago](https://github.com/purescript/spago/blob/master/src/Spago/Paths.purs):
- the global cache
- the local cache
- the registry repo
- the index
- the database

JordanMartinez:
### Description of the change

Fixes #1082. 

Does this need an flag for each path value?

### Checklist:

### DIFF
--- a/src/Spago/Command/Ls.purs
+++ b/src/Spago/Command/Ls.purs
@@ -3,6 +3,7 @@ module Spago.Command.Ls
   , listPackages
   , listPackageSet
   , LsEnv(..)
+  , LsPathsArgs
   , LsDepsArgs
   , LsPackagesArgs
   ) where
@@ -10,6 +11,7 @@ module Spago.Command.Ls
 import Spago.Prelude
 
 import Data.Codec.Argonaut as CA
+import Data.Codec.Argonaut.Common as CAC
 import Data.Codec.Argonaut.Record as CAR
 import Data.Foldable (elem, traverse_)
 import Data.Map (filterKeys)
@@ -40,6 +42,10 @@ type LsDepsOpts =
   , transitive :: Boolean
   }
 
+type LsPathsArgs =
+  { json :: Boolean
+  }
+
 type LsSetEnv =
   { dependencies :: Fetch.PackageTransitiveDeps
   , logOptions :: LogOptions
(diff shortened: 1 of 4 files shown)


src/Spago/Command/Ls.purs
```diff
@@ -45,6 +53,18 @@ type LsEnv =
   , selected :: WorkspacePackage
   }
 
+listPaths :: Spago { logOptions :: LogOptions } Unit
+listPaths = do
+  logDebug "Running `listPaths`"
+  traverse_ logInfo
```
f-f, line 59: `logInfo` goes to stderr, which we use for logging. Program output should go to `stdout`.

We already have facilities for this (the `output` function), see e.g. how the other `ls` commands or the `registry‘ commands do it.
It's quite fancy at this point 😄

We might want a table here, as it's the default output mode for the other ls commands too

src/Spago/Command/Ls.purs
```diff
@@ -45,6 +53,18 @@ type LsEnv =
...
+    [ "Global cache path: " <> Paths.globalCachePath
+    , "Global registry path: " <> Paths.registryPath
+    , "Global registry index path: " <> Paths.registryIndexPath
+    , "Global package sets path: " <> Paths.packageSetsPath
+    , "Local cache path: " <> Paths.localCachePath
+    , "Local cache packages path: " <> Paths.localCachePackagesPath
+    ]
```
f-f, line 66: Could we add the db path here as well? We'll need to move it from `Spago.Db.databasePath` to here

src/Spago/Command/Ls.purs
```diff
@@ -45,6 +53,18 @@ type LsEnv =
   , selected :: WorkspacePackage
   }
 
+listPaths :: Spago { logOptions :: LogOptions } Unit
+listPaths = do
```
f-f, line 57:
```suggestion
listPaths { json } = do
```

We don't want flags for each path, but we do want a flag to output Json, as with the other ls commands

JordanMartinez:
Latest commit addressed feedback.

f-f (approved):
This is great, thanks Jordan!

PR #1102, opened 2023-10-20
Update README.md with notes about shell completions when installing w…

postsolar:
…ith tools other than NPM.

### Description of the change

Add a note to README which explains the issue which occurs when generating shell completions scripts for a Spago installation done by PNPM or other non-NPM package managers.

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1041,10 +1041,10 @@ or alternatively if you don't want to edit your `~/.bashrc`:
 spago --bash-completion-script $(which spago) >> ~/.bash_completion
 ```
 
-*Note*: if you installed Spago not with NPM, but with PNPM or some other package manager,
-this package manager might have bundled your installation and your package name in the script may end up being incorrect.
-For example, when installed with PNPM, the resulting script will reference incorrect package `bundle.js` instead of `spago`.
-If you're using something other than NPM, verify the referenced package name in the completions script.
+> [!NOTE]\
+> If you installed Spago not with NPM, but with PNPM or some other package manager, this package manager might have bundled your installation and your package name in the script may end up being incorrect.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1041,6 +1041,11 @@ or alternatively if you don't want to edit your `~/.bashrc`:
 spago --bash-completion-script $(which spago) >> ~/.bash_completion
 ```
 
+*Note*: if you installed Spago not with NPM, but with PNPM or some other package manager,
+this package manager might have bundled your installation and your package name in the script may end up being incorrect.
+For example, when installed with PNPM, the resulting script will reference incorrect package `bundle.js` instead of `spago`.
+If you're using something other than NPM, verify the referenced package name in the completions script.
+
```
f-f, line 1048:
```suggestion
> [!NOTE]\
> If you installed Spago not with NPM, but with PNPM or some other package manager, this package manager might have bundled your installation and your package name in the script may end up being incorrect.
> For example, when installed with PNPM, the resulting script will reference incorrect package `bundle.js` instead of `spago`.
> If you're using something other than NPM, verify the referenced package name in the completions script.

```

README.md
```diff
@@ -1068,6 +1073,7 @@ compinit
 
 *Note*: you might need to call this multiple times for it to work.
 
+*Note*: see the note in the Bash section above when installing Spago with a package manager other than NPM.
```
f-f, line 1076:
```suggestion
> [!NOTE]\
> You might need to call this multiple times for it to work.

> [!NOTE]\
> See the note in the Bash section above when installing Spago with a package manager other than NPM.
```

f-f (approved):
Thanks!

f-f:
I updated the notes to use the GitHub syntax for displaying nice notes 🙂

PR #1104, opened 2023-10-23
Update flake and lockfile, add spago and purs-backend-es to shell

thomashoneyman:
### Description of the change

This minor internal-only update does a couple things:

1. Updates the flake to the latest purescript-overlay (giving access to spago 0.93.18, among other things)
2. Adds spago to the dev shell. Spago is used in the package.json scripts but we don't actually specify a version anywhere; now, the latest spago from purescript-overlay will be used.
3. Adds purs-backend-es to the dev shell. Right now the optimizer isn't used, but I've added it so it's easy to integrate into e.g. the publish workflow at some time; it provides noticeable speed improvements so it would be nice to publish an optimized spago.
4. Updates the spago.lock file. Something must be funky about the lockfile generation, as this was out-of-date regarding the updated ordered-collections version from a few days ago.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -20,6 +20,7 @@
             purs
             purs-tidy
             purs-backend-es
+            spago-unstable
 
             nodejs
             esbuild
--- a/flake.nix
+++ b/flake.nix
@@ -15,11 +15,12 @@
         };
       in {
         devShells.default = pkgs.mkShell {
+          name = "spago";
           buildInputs = with pkgs; [
-            purs-bin.purs-0_15_10
-            # TODO: we need to get purescript-overlay to track the latest versions
-            # spago-unstable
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -15,11 +15,13 @@
         };
       in {
         devShells.default = pkgs.mkShell {
+          name = "spago";
           buildInputs = with pkgs; [
-            purs-bin.purs-0_15_10
-            # TODO: we need to get purescript-overlay to track the latest versions
-            # spago-unstable
-            purs-tidy-bin.purs-tidy-0_10_0
+            purs
```
thomashoneyman, line 20: This will provide the latest stable. Right now that's 0.15.10.

f-f (approved):
Thanks! And yes, lockfiles are broken until we fix #917 😄

PR #1105, opened 2023-10-23
Add timings to verbose output

f-f:
This PR adds to the verbose output the amount of milliseconds elapsed from the beginning of the execution, so that it's easier to find slow bits that can be sped up.

cc @JordanMartinez @finnhodgkin 

It looks something like this (bikeshedding on how they should look like is welcome):

image

### DIFF
--- a/core/src/Log.purs
+++ b/core/src/Log.purs
@@ -32,9 +32,15 @@ import Data.Array as Array
 import Data.Array.NonEmpty (NonEmptyArray)
 import Data.Array.NonEmpty (toArray) as NEA
 import Data.Codec.Argonaut (JsonCodec)
+import Data.DateTime.Instant (Instant)
+import Data.DateTime.Instant as Instant
 import Data.Either (Either(..))
+import Data.Int as Int
 import Data.Maybe (Maybe(..), fromMaybe)
+import Data.Newtype (unwrap)
 import Data.String as String
+import Data.String.Utils as Strings
+import Data.Time.Duration (Milliseconds)
 import Data.Traversable (traverse)
 import Dodo (Doc, print, twoSpaces)
 import Dodo (indent, break) as DodoExport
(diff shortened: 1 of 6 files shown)


core/src/Log.purs
```diff
@@ -100,6 +111,13 @@ log { content, level } = do
   case logOptions.verbosity, level of
     LogQuiet, _ -> pure unit
     LogNormal, LogDebug -> pure unit
+    LogVerbose, _ -> do
+      now <- Effect.liftEffect $ Now.now
+      let
+        (timeDiff :: Milliseconds) = Instant.diff now logOptions.startingTime
+        millisDoc = Ansi.foreground Ansi.White $ Ansi.bold $ Log.text $ Strings.padStart 8 $ show $ Int.round $ unwrap timeDiff
```
finnhodgkin, line 118: This is so much nicer than the million `time`/`timeEnd`s I was dropping in before!

Given that the times just ascend is the padding necessary? I'd find this hard to parse:
```
[80ms] ...
[802323ms] ...
[802ms] ...
``` 

But not so much this:
```
[4ms] ...
[10ms] ...
[75ms] ...
[80ms] ...
[2000ms] ...
[32322ms] ...
```

Maybe it's more noticeably jumpy at the start. Could drop it down to `4` so the nudge only happens after longer intervals?

```suggestion
        millisDoc = Ansi.foreground Ansi.White $ Ansi.bold $ Log.text $ Strings.padStart 4 $ show $ Int.round $ unwrap timeDiff
```

f-f: > But not so much this

My eye finds that equally hard to parse, and I figured out because I started with what you're suggesting here, and ended up with the padding 😄

f-f: I am unsure about the square brackets though. I just wanted some separation with the rest of the log line, I guess a dash or something could work as well

JordanMartinez: I don't think I care one way or another whether there's brackets or hyphens used.

I do like the current padded usage.

Issue #1048, opened 2023-10-07
When `spago run`, if the target cli app manipulates console interface it may not terminate properly


edwardw:
Just use the example from [`purescript-node-readline`](https://github.com/purescript-node/purescript-node-readline/blob/master/test/Main.purs) repo:

```purescript
module Main where

import Prelude

import Effect (Effect)
import Effect.Console (log)
import Node.EventEmitter (on_)
import Node.ReadLine (close, createConsoleInterface, lineH, noCompletion, prompt, setPrompt)

main :: Effect Unit
main = do
  interface <- createConsoleInterface noCompletion
  setPrompt "(type 'quit' to stop)\n> " interface
  prompt interface
  interface # on_ lineH \s ->
    if s == "quit" then close interface
    else do
      log $ "You typed: " <> s
      prompt interface
```

If `spago bundle` then `node index.js`, after typing quit the app terminates. But if `spago run` directly, it hangs. It terminates only after pressing `ctrl-d`. With `spago run --verbose`, it can be deduced that the hang must have happened here: [src/Spago/Cmd.purs#L59](https://github.com/purescript/spago/blob/d3174e2dbf2a7e9d5ddb97e7bc304702b5d9cb7f/src/Spago/Cmd.purs#L59).

This happened with `spago` version `0.93.9`. I also checkout out this repo and built a fresh version from head. It also happened there.

edwardw:
Also dug a little bit around node documentation. It seems there must be some reason there exist three flavor of stream piping API `stream.pipe`, `stream.pipeline` and the experimental `stream.compose`. But better leave that to you node experts.

JordanMartinez:
Hm... This is probably a bug with `node-execa`. Spago is calling `subprocess.stdin.shareParentProcessStdin`, so the parent stdin is being shared with the `subprocess stdin`.

f-f:
We're not using the latest execa version - @JordanMartinez do you think that "just" upgrading to latest might fix this?

JordanMartinez:
I'm not sure. I'll have to see what goes on in `execa`.

JordanMartinez:
So, after I updated `Test/Prelude.purs`' `spago'` to use `spawn` rather than `exec`, I added the following test. Note: the `node-readline` code below is v8.1.0, not v7.0.0 in the opening thread.:
```purs

    Spec.itOnly "can run interactive programs" \{ spago, spago' } -> do
      spago [ "init" ] >>= shouldBeSuccess
[...]

PR #1106, opened 2023-10-23
Manually order config codecs

thomashoneyman:
### Description of the change

Currently all JSON codecs use the `Codec.Argonaut.Record` module with its record sugar for defining objects. While convenient, this method will always order fields alphabetically in JSON, which isn't what we want in a config file. Instead we should manually order these fields. As an example, a fresh init of a config file will place the `dependencies` key before the package's `name` key; they really ought to go the other way around.

This PR switches to a manual encoding of the config formats.

### Notes

I noticed that we generally reuse the YAML fields directly, such as `exec_args` being typed in PureScript as `type ... = { exec_args :: ... }`. If we want to be more purescript-y we can rename these fields in the codecs so the YAML is `exec_args` but the PureScript is `execArgs`.

The reason this particularly caught my attention is that we generally use snake case in YAML _except_ for `execArgs`, which is camel case. Thought it may have been an oversight. But maybe it's intentional.

I'm also seeing some odd test failures but I'll report on them once the suite runs on this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [![build](https://github.com/purescript/spago/actions/workflows/build.yml/badge.svg)](https://github.com/purescript/spago/actions/workflows/build.yml)
 [![Maintainer: f-f](https://img.shields.io/badge/maintainer-f%2d-f-teal.svg)](http://github.com/f-f)
 
-*(IPA: /ˈspaɡo/)*
+_(IPA: /ˈspaɡo/)_
 
 PureScript package manager and build tool.
 
@@ -22,23 +22,24 @@ PureScript package manager and build tool.
 > This new Spago is still in alpha, so while most of it works well, there will be some rough edges here and there. Please report them if you find any!
 
 The recommended installation method for Windows, Linux and macOS is `npm` (see the latest releases on npm
-  [here][spago-npm]):
+[here][spago-npm]):
 
 ```
 npm install -g spago@next
 ```
 
 Other installation methods available:
+
 - With Nix, using [purescript-overlay]
 
 **General notes:**
+
 - The assumption is that you already installed the [PureScript compiler][purescript].
   If not, get it with `npm install -g purescript`, or the recommended method for your OS.
 - You might have issues with `npm` and Docker (e.g. getting the message "Downloading the spago binary failed.." etc)
   You have two options:
   - either **do not run npm as root**, because it doesn't work well with binaries. Use it as a nonprivileged user.
   - or use `--unsafe-perm`: `npm install -g --unsafe-perm spago@next`
-
 
 ## Super quick tutorial
 
@@ -63,6 +64,7 @@ This last command will create a few files:
 ```
 
 If you have a look at the `spago.yaml` file, you'll see that it contains two sections:
+
 - [the `workspace` section](#the-workspace), which details the configuration for the _dependencies_ of the project as a whole (which can be a monorepo, and contain more than one package), and other general configuration settings.
   In this sample project, the only configuration needed is the [package set](#whats-a-package-set) version from which all the dependencies will be chosen. See [here](#querying-package-sets) for more info about how to query the package sets.
 - [the `package` section](#whats-a-package), that is about the configuration of the package at hand, such as its name, dependencies, and so on.
@@ -76,6 +78,7 @@ $ spago run
 ```
 
 This will:
+
 - download and compile the necessary dependencies (equivalent to `spago install`)
 - compile this sample project in the `output/` directory (equivalent to `spago build`).\
   You can take a look at the content of `output/Main/index.js` to see what kind of JavaScript has been generated from your new `Main.purs` file
@@ -95,6 +98,7 @@ $ node .
 Great! If you read unitl here you should be set to go write some PureScript without worrying too much about the build 😊
 
 Where to go from here? There are a few places you should check out:
+
 - see [the "How to achieve X"](#how-do-i) section for practical advice without too much explanation
 - see instead the [Concepts and Explanations](#concepts-and-explanations) section for more in-depth explanations about the concepts that power Spago, such as [package sets](#whats-a-package-set), or [the Workspace](#the-workspace).
 
@@ -153,10 +157,10 @@ Where to go from here? There are a few places you should check out:
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-
 ## Design goals and reasons
 
 Our main design goals are:
+
 - **Great UX**: a good build system just does what's most expected and gets out of the way so you can focus on actually thinking about the software itself, instead of spending your time configuring the build.
 - **Minimal dependencies**: users should not be expected to install a myriad of tools on their system to support various workflows. Spago only expects `git` and `purs` to be installed.
 - **Reproducible builds**: we exploit [package sets](#whats-a-package-set) and [lock files](#the-lock-file) to make your build reproducible, so that if your project builds today it will also build tomorrow and every day after that.
@@ -169,13 +173,13 @@ Some tools that inspired `spago` are: [Rust's Cargo][cargo], [Haskell's Stack][s
 We'd love your help, and welcome PRs and contributions!
 
 Some ideas for getting started:
+
 - [Build and run `spago`](CONTRIBUTING.md#developing-spago)
 - [Help us fix bugs and build features](https://github.com/purescript/spago/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22+label%3A%22defect%22)
 - Help us improve our documentation
 - Help us [log bugs and open issues][new-issue]
 
 For more details see the [`CONTRIBUTING.md`][contributing]
-
 
 ## How do I...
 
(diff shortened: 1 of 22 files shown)


core/src/Config.purs
```diff
@@ -57,17 +56,18 @@ import Registry.Range as Range
...
 
 configCodec :: JsonCodec Config
-configCodec = CAR.object "Config"
-  { package: CAR.optional packageConfigCodec
-  , workspace: CAR.optional workspaceConfigCodec
-  }
+configCodec = CA.object "Config"
+  $ CA.recordPropOptional (Proxy :: _ "package") packageConfigCodec
+  $ CA.recordPropOptional (Proxy :: _ "workspace") workspaceConfigCodec
+  $ CA.record
```
thomashoneyman, line 70: I generally stuck to the order I saw defined in the type synonyms. But if you want a different order for any fields just lmk and I can change it.

f-f:
> The reason this particularly caught my attention is that we generally use snake case in YAML except for execArgs, which is camel case. Thought it may have been an oversight. But maybe it's intentional.

It's an oversight and should be fixed 😄

(Unless we'd rather switch everything to camelCase. I don't have a strong preference)

thomashoneyman:
One test failure:

I had some trouble tracking down where in the code this is exactly happening, but for some reason the test section below is not respecting the codec ordering and is always ordering `dependencies` before `main`. This is a departure from the rest of the fixtures, which behave correctly.

```
  ✗ adds test dependencies to the config file when the test section does not exist:
  "package:\n  name: aaa\n  dependencies:\n    - console\n    - effect\n    - prelude\n  test:\n    dependencies:\n      - foreign\n    main: Test.Main\nworkspace:\n  package_set:\n    registry: 29.3.0\n  extra_packages: {}\n"
≠
"package:\n  name: aaa\n  dependencies:\n    - console\n    - effect\n    - prelude\n  test:\n    main: Test.Main\n    dependencies:\n      - foreign\nworkspace:\n  package_set:\n    registry: 29.3.0\n  extra_packages: {}\n"
```

core/src/Config.purs
```diff
@@ -101,24 +101,24 @@ type PublishConfig =
...
   }
 
 runConfigCodec :: JsonCodec RunConfig
-runConfigCodec = CAR.object "RunConfig"
-  { main: CAR.optional CA.string
-  , execArgs: CAR.optional (CA.array CA.string)
-  }
+runConfigCodec = CA.object "RunConfig"
+  $ CA.recordPropOptional (Proxy :: _ "main") CA.string
+  $ CA.recordPropOptional (Proxy :: _ "execArgs") (CA.array CA.string)
```
f-f, line 120:
```suggestion
  $ CA.recordPropOptional (Proxy :: _ "exec_args") (CA.array CA.string)
```

core/src/Config.purs
```diff
@@ -130,25 +130,25 @@ type TestConfig =
...
-  { main: CA.string
-  , execArgs: CAR.optional (CA.array CA.string)
-  , dependencies: dependenciesCodec
-  , censor_test_warnings: CAR.optional censorBuildWarningsCodec
-  , strict: CAR.optional CA.boolean
-  , pedantic_packages: CAR.optional CA.boolean
-  }
+testConfigCodec = CA.object "TestConfig"
+  $ CA.recordProp (Proxy :: _ "main") CA.string
+  $ CA.recordPropOptional (Proxy :: _ "execArgs") (CA.array CA.string)
```
f-f, line 135:
```suggestion
  $ CA.recordPropOptional (Proxy :: _ "exec_args") (CA.array CA.string)
```

f-f (approved):
Sooo pretty 🙂

This is great - now that we encode the fields manually I'm up for changing the records to be CamelCase if you'd prefer that as well

f-f:
> I had some trouble tracking down where in the code this is exactly happening

My bet is on `Config.js` - we manually manipulate the yaml there and maybe there's some automatic sorting going on (IIRC there's an option to disable it)

f-f:
Hmm, `Yaml.stringify` has a `sortMapEntries` which is the option I'm thinking about, but it defaults to false and we are not currently overwriting it

src/Spago/Config.js
```diff
@@ -12,7 +12,7 @@ export function addPackagesToConfigImpl(doc, isTest, newPkgs) {
 
   const deps = (() => {
     if (isTest) {
-      const test = getOrElse(pkg, "test", doc.createNode({ dependencies: [], main: "Test.Main" }));
+      const test = getOrElse(pkg, "test", doc.createNode({ main: "Test.Main", dependencies: [] }));
```
thomashoneyman, line 15: This was the culprit.

README.md
```diff
@@ -5,7 +5,7 @@
 [![build](https://github.com/purescript/spago/actions/workflows/build.yml/badge.svg)](https://github.com/purescript/spago/actions/workflows/build.yml)
 [![Maintainer: f-f](https://img.shields.io/badge/maintainer-f%2d-f-teal.svg)](http://github.com/f-f)
 
-*(IPA: /ˈspaɡo/)*
+_(IPA: /ˈspaɡo/)_
```
thomashoneyman, line 8: Sorry, I updated this file for the use of `execArgs` -> `exec_args` and, as I use Prettier, it was auto-formatted. If you want me to revert the changes to this file then I can (though I personally agree with the changes made).

thomashoneyman:
> I'm up for changing the records to be CamelCase if you'd prefer that as well

I don't care one way or another, just wanted to check. If we did, it'd look something like this:

```purs
type MyObject = { blahBlah :: String }

codec = Profunctor.dimap to from $ CA.Record.object "MyObject"
  { blah_blah: CA.string
  }
  where
  to = Record.rename (Proxy :: _ "blah_blah") (Proxy :: _ "blahBlah")
  from = Record.rename (Proxy :: _ "blahBlah") (Proxy :: _ "blah_blah")
```

f-f:
> If we did, it'd look something like this

Oh I see - I thought it would come for free with the manual encoding, but instead it seems a little tedious so let's keep things as they are

bin/src/Main.purs
```diff
@@ -123,7 +123,7 @@ type RunArgs =
   , pedanticPackages :: Boolean
   , pursArgs :: List String
   , backendArgs :: List String
-  , execArgs :: Maybe (Array String)
+  , exec_args :: Maybe (Array String)
```
f-f, line 126: The rest of the fields here are camel case, so I'd rather not rename this

bin/src/Flags.purs
```diff
@@ -183,8 +183,8 @@ pursArgs =
               )
       )
 
-execArgs :: Parser (Maybe (Array String))
-execArgs =
+exec_args :: Parser (Maybe (Array String))
+exec_args =
```
f-f, line 187: No need to rename the flag either I think

core/src/Config.purs
```diff
@@ -101,54 +101,54 @@ type PublishConfig =
...
+  $ CA.recordProp (Proxy :: _ "license") License.codec
+  $ CA.recordPropOptional (Proxy :: _ "location") Location.codec
+  $ CA.recordPropOptional (Proxy :: _ "include") (CA.array CA.string)
+  $ CA.recordPropOptional (Proxy :: _ "exclude") (CA.array CA.string)
+  $ CA.record
 
 type RunConfig =
   { main :: Maybe String
-  , execArgs :: Maybe (Array String)
+  , exec_args :: Maybe (Array String)
```
f-f, line 114: These (in RunConfig and TestConfig) are the only snake cases we need, it should go in camel case in the rest of the codebase

README.md
f-f: Yeah it's fine, these changes are fairly innocent

thomashoneyman:
Sorry about that! Renamed just the two now.

thomashoneyman:
Not sure about the macOS failures — are these flaky?

```
spago » run
  ✓︎ works at all
Running test in /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/spago-nodejs/7368613235362d427376674a41584638796e4d6861586e4252684f4a716550465a746f6c6b2f6b4d446441756771472b66733d
  ✗ can pass stdin to the application:

  test timed out after 90000ms
Running test in /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/spago-nodejs/7368613235362d4b6d43784c6a36554c784d55486352483632527964766a384634537a504d4c7a304851424232674f5a30513d
  ✗ can pass arguments to the application:

  test timed out after 90000ms
Running test in /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/spago-nodejs/7368613235362d384d30555832524948586a585539307857306a7659615166456c4e326b41676e73696c5145502b797167553d
```

f-f:
They are flaky, we think they are related to #1048

f-f (approved):

thomashoneyman:
Also now a Windows failure, this time I believe because this line is not expected in the fixture:

```
Refreshing the Registry Index...\n
```

Is there a way to control this, perhaps by running all tests in `--offline` mode?

thomashoneyman:
I think this is done as far as I'm concerned, so feel free to merge away if you are OK ignoring the tests!

PR #1111, opened 2023-11-09
Add uninstall command

### LINKED ISSUE #1091: Add `uninstall` command
See https://github.com/purescript/spago/issues/1042#issuecomment-1762918583

Blocks #1042

JordanMartinez:
### Description of the change

Fixes #1091.

### Checklist:


bin/src/Main.purs
```diff
@@ -536,6 +554,10 @@ main = do
             env' <- runSpago env (mkBuildEnv buildArgs dependencies)
             let options = { depsOnly: true, pursArgs: List.toUnfoldable args.pursArgs, jsonErrors: false }
             runSpago env' (Build.run options)
+          Uninstall { packagesToRemove, selectedPackage, testDeps } -> do
+            { env, fetchOpts } <- mkFetchEnv offline { packages: packagesToRemove, selectedPackage, ensureRanges: false, testDeps: false }
```
f-f, line 558: The `packages` here are additional packages to add to the config, which is the opposite of what we want to achieve here 😄 
```suggestion
            { env, fetchOpts } <- mkFetchEnv offline { packages: [], selectedPackage, ensureRanges: false, testDeps: false }
```

bin/src/Main.purs
```diff
@@ -536,6 +554,10 @@ main = do
...
+            let options = { testDeps, dependenciesToRemove: Set.fromFoldable fetchOpts.packages }
```
f-f, line 559: Then I guess we should parse the package names from `packagesToRemove` instead of getting them from `fetchOpts`

src/Spago/Command/Uninstall.purs
```diff
@@ -0,0 +1,129 @@
...
+     , onError :: String -> m Unit
+     }
+  -> m Unit
+editSpagoYaml { configPath, modifyConfig, onError } = do
+  content <- liftAff $ FS.readYamlDocFile Core.configCodec configPath
+  case content of
+    Left err ->
+      onError err
+    Right { yaml: config } ->
+      liftAff $ FS.writeYamlFile Core.configCodec configPath $ modifyConfig config
```
f-f, line 129: This piece of code was fine for tests, but it's not good to use for normal business - rewriting the config like that loses comments and extra fields the config might contain.

We should instead manipulate the `Yaml.Doc`, and use the same setup [as it's done in the Config module](https://github.com/purescript/spago/blob/23697f4f7e93cfcdf7d71b0833f9591588747948/src/Spago/Config.purs#L521), where we get the `Doc` from the workspace, and apply an FFI function that does the yaml manipulation while preserving comments.

bin/src/Main.purs
JordanMartinez: > ...are additional packages to add to the config

AFAICT via [`mkFetchEnv`](https://github.com/purescript/spago/blob/master/bin/src/Main.purs#L880-L897), that's not true. Rather, we just validate there that the strings we got from the CLI can be parsed into `PackageName` values. The resulting `packageNames` value is never used anywhere else and is just stored in `fetchOpts`. And for context, the only reason why I'm using `mkFetchEnv` instead of doing this outside of that is because I need a `workspace` value.

src/Spago/Command/Uninstall.purs
JordanMartinez: Ah, ok. Let me fix that.

bin/src/Main.purs
f-f: Ah yeah sorry - I got confused about `mkFetchEnv` vs `Fetch.run`, all good

f-f (approved):
Looks great!

PR #1113, opened 2023-11-09
Fix `bundle-app` when non-`node` platform

### LINKED ISSUE #1112: Node shebang in minified and bundled browser app code
It appears that there is a node shebang being output in our minified code when bundling our app for the browser. This is causing issues for a few of the users of our application, because security software is identifying our bundled/minified code as a shell script.

Should the shebang be there, or is this a mistake?
https://github.com/purescript/spago/blob/2f272e6d7db00645600318e327207d78a0b97c84/src/Spago/Command/Bundle.purs#L62C57-L62C57

JordanMartinez:
### Description of the change

Fixes #1112 

### Checklist:


f-f (approved):
👏👏

PR #1115, opened 2023-11-11
Fix: Error message when `purs` is not on path

CharlesTaylor7:


src/Spago/Cmd.purs
```diff
@@ -118,7 +118,7 @@ getExecutable command =
         Right r -> pure { cmd: cmd1, output: r.stdout }
         Left err' -> do
           let cmd2 = mkCmd command Nothing
-          logDebug [ "Failed to find purs.cmd. Trying with just purs...", show err' ]
+          logDebug [ "Failed to find ", command, ".cmd. Trying with just purs...", show err' ]
```
f-f, line 121: I think we'll want this change merged or this log will be very confusing 😄

PR #1130, opened 2023-12-03
Use multiple top-level values for BuildInfo.purs

### LINKED ISSUE #1128: Design of the `BuildInfo.purs` file
There's three questions I have.

#### Question 1: 1 record vs multiple values

In https://github.com/purescript/spago/pull/1126#issuecomment-1834665101, I asked why we currently output the `BuildInfo.purs` file's content as a single top-level record with all of the data there rather than outputting each field as its own top-level value.

In other words, why this
```purs
module Spago.Generated.BuildInfo where

buildInfo :: { packages :: <recordType>, pursVersion :: String, spagoVersion :: String }
buildInfo =
  { packages: { "foo": "1.2.3", "bar": "1.2.3" }
  , pursVersion: "1.2.3"
  , spagoVersion: "1.2.3"
  }
```
instead of this?
```purs
module Spago.Generated.BuildInfo where

packages :: <recordType>
packages =
  { "foo": "1.2.3"
  , "bar": "1.2.3"
  }

pursVersion :: String
pursVersion = "1.2.3"

spagoVersion :: String
spagoVersion = "1.2.3"
```

To me the latter makes more sense because then one can use such values via `BuildInfo.packages.foo` rather than `BuildInfo.buildInfo.packages.foo`.

#### Question 2: Adding inline directives

Related to the above. Since more and more people are using `purs-backend-es`, wouldn't it make sense to add inline directives for this value? That way, if some code should only be run when compiled with `purs >= 1.2.3`, the resulting code would optimize away that check? As of right now, I don't think it is optimized away.

#### Question 3: Publishing

What happens if a user happens to publish a library that uses the `Spago.Generated.BuildInfo` module? Wouldn't that cause problems for package sets and/or anyone using that library due to getting a duplicate module error in both situations? Are there any checks in the Registry to prevent this from happening?

JordanMartinez:
### Description of the change

Somewhat dependent on #1126 (adds tests). Fixes #1128.

### Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,14 +72,14 @@ jobs:
       - name: Build the project
         run: spago build
 
-      - name: Bootstrap executable
-        run: node ./bin/index.dev.js bundle -p spago-bin
+      # - name: Bootstrap executable
+      #   run: node ./bin/index.dev.js bundle -p spago-bin
 
-      - name: Bundle docs-search client
-        run: node ./bin/bundle.js bundle -p docs-search-client-halogen
+      # - name: Bundle docs-search client
+      #   run: node ./bin/bundle.js bundle -p docs-search-client-halogen
 
-      - name: Run tests
-        run: node ./bin/bundle.js test
+      # - name: Run tests
+      #   run: node ./bin/bundle.js test
 
       - name: Check formatting (Linux only)
         if: matrix.os == 'ubuntu-latest'
--- a/bin/src/Main.purs
+++ b/bin/src/Main.purs
@@ -500,7 +500,7 @@ main = do
     printVersion = do
       logOptions <- mkLogOptions startingTime { noColor: false, quiet: false, verbose: false, offline: Offline }
       runSpago { logOptions } do
-        logInfo BuildInfo.buildInfo.spagoVersion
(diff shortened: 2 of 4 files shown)


JordanMartinez:
Should these changes be done in #1126?

JordanMartinez:
Helps to push to the right branch...

JordanMartinez:
@f-f Looks like we got a chicken-and-egg problem. I need to reference the new BuildInfo file to get the bootstrap build step to work, but doing so fails to build the initial code to run the tests.

f-f:
Right - I think the correct procedure is:
- we merge this
- I cut a new release
- then push a commit that re-enables the bootstrap

@JordanMartinez please double check this 😄

JordanMartinez:
That sounds correct to me.

f-f (approved):

f-f:
Done: https://github.com/purescript/spago/actions/runs/7090846490

f-f:
Tests fail though: https://github.com/purescript/spago/actions/runs/7090846490/job/19298707004

f-f:
Ah I see - this is because I fixed the output of `spago --version` to (correctly) go to stdout instead of stderr. Patch incoming

PR #1138, opened 2023-12-10
Support lockfiles, take 2

### LINKED ISSUE #917: Support spago.lock files

### LINKED ISSUE #1134: Don't require a checkout of the registry repo when building with a lockfile
In Nix environments it isn't possible to hit the network during a build unless you have a hash of the entity you are downloading — so as to prove what you downloaded is exactly what you expected and did not change.

In these environments it is still possible to download dependencies for a Spago project because the lockfile contains a hash for every entry. However, if you then run `spago build --offline` the build will fail because Spago wants the registry and registry-index repositories present.

This is problematic because we can't clone these repositories in the offline environment — we'd need to hash the registry and registry-index and keep those hashes up to date so as to download and place them where Spago wants for the build.

But the registry and regsitry-index don't contain any information beyond what is already captured in the lockfile — at least for performing a build (you would still need them for things like `spago install`, but you'd never do that in an offline environment).

My request is that `spago build`, `spago test`, and other commands that do not require information beyond what is in the lockfile also do not require the registry and registry-index to be present, since those are difficult to procure in an offline environment like CI. If that is a bit thorny, then at least I'd like to relax the repos being present requirement when using the `--offline` flag in a build.

f-f:
This PR fixes #917 (and fixes #1134 as well while we're at it), adding lockfile support for all builds (solver and package sets), with no opt-out, i.e. all spago invocations that imply a build plan will now output a lockfile, or use the existing one to save work.

We also add a `--pure` flag to skip updating the lockfile even if it's deemed out of date.

Note: this is a breaking change for the lockfile format, but right now they are so broken that no one should be using them 🙂


spago.lock
```diff
@@ -38,6 +39,11 @@ workspace:
       test_dependencies: []
     docs-search-common:
       path: docs-search/common
+      needed_by:
+        - spago-bin
+        - spago
+        - docs-search-index
+        - docs-search-client-halogen
```
f-f, line 46: Both the workspace packages and the plan packages now include a `needed_by` key, so we don't even spend time computing a plan (which takes time), and instead allows for a blanket approach of "read everything in, filter for what we need, and that's our plan"

thomashoneyman:
Before I dive in, does this also address #1134?

src/Spago/Command/Fetch.purs
```diff
@@ -278,6 +240,75 @@ run { packages: packagesToInstall, ensureRanges, isTest } = do
...
+              pure { version, integrity, dependencies: packageDependencies }
+          updatePackage $ FromRegistry registryVersion
+        LocalPackage { path } -> do
+          updatePackage $ FromPath { path, dependencies: packageDependencies }
+
+  let
+    toArray :: forall k v. Map k v -> Array (Tuple k v)
+    toArray = Map.toUnfoldable
+  ({ packages, workspacePackages } :: LockfileBuilderResult) <-
+    Array.foldM processPackage
```
f-f, line 297: We used to just map the packages into the lockfile data structure, but we now use a fold to go through them so we can keep track of which workspace package needs which other package, which makes it dead easy to figure out the build plan afterwards

src/Spago/Command/Fetch.purs
```diff
@@ -341,16 +372,47 @@ type TransitiveDepsResult =
...
   { workspace } <- ask
-  case workspace.packageSet of
-    Registry extraPackages -> do
-      plan <- getTransitiveDepsFromRegistry depsRanges extraPackages
-      logDebug $ "Got a plan from the Solver: " <> printJson (Internal.Codec.packageMap Version.codec) plan
-      pure (map RegistryVersion plan)
-    PackageSet set -> getTransitiveDepsFromPackageSet set (Array.fromFoldable $ Map.keys depsRanges)
+  case workspace.packageSet.lockfile of
+    -- If we have a lockfile we can just use that - we don't even need build a plan, we can just filter the packages
+    -- marked as needed by the workspace package that we are processing, as we just dumped the plan in the lockfile.
```
f-f, line 381: Computing the build plan when we have a lockfile is basically "filter by `needed_by`"

src/Spago/Config.purs
```diff
@@ -410,6 +391,23 @@ rootPackageToWorkspacePackage { rootPackage, workspaceDoc } = do
   hasTests <- liftEffect $ FS.exists "test"
   pure { path: "./", doc: workspaceDoc, package: rootPackage, hasTests }
 
+workspacePackageToLockfilePackage :: WorkspacePackage -> Tuple PackageName Lock.WorkspaceLockPackage
+workspacePackageToLockfilePackage { path, package } = Tuple package.name
+  { path
+  , dependencies: package.dependencies
+  , test_dependencies: foldMap _.dependencies package.test
+  , needed_by: mempty -- Note: this is filled in later
+  }
```
f-f, line 400: These functions feel like they belong to the `Lock` module, but that would create a cycle, so they are stuck here. I think it's fine, not too much code

src/Spago/Config.purs
```diff
@@ -410,6 +391,23 @@ rootPackageToWorkspacePackage { rootPackage, workspaceDoc } = do
...
+
+shouldComputeNewLockfile :: { workspace :: Core.WorkspaceConfig, workspacePackages :: Map PackageName WorkspacePackage } -> Lock.WorkspaceLock -> Boolean
+shouldComputeNewLockfile { workspace, workspacePackages } workspaceLock =
+  -- the workspace packages should exactly match, except for the needed_by field, which is filled in during build plan construction
+  (map (workspacePackageToLockfilePackage >>> snd) workspacePackages == (map (_ { needed_by = mempty }) workspaceLock.packages))
+    -- and the extra packages should exactly match
+    && (fromMaybe Map.empty workspace.extra_packages == workspaceLock.extra_packages)
+    -- and the package set address needs to match - we have no way to match the package set contents at this point, so we let it be
+    && (workspace.package_set == map _.address workspaceLock.package_set)
```
f-f, line 409: This is the bit of logic that computes when the lockfile needs a refresh - it's a bit involved because some things are generated and we don't have them at the point where we read the lockfile in (like, we stash the whole package set in there), but hopefully it's commented enough to make sense

f-f: And dumping the package set in there is important so that we can then:
1. skip hashing it, since we just have the content
2. use it for commands like `spago ls`, where we can use just the lockfile as well

f-f:
@thomashoneyman CI fails without that patch, so I pushed the fix for #1134 as well!

spago.lock
thomashoneyman: I'm not sure I'm following this — if we list the dependencies for each package (including workspace ones), what is the advantage of needed_by?

In other words what is the difference between a) building a workspace package by looking up each of its dependencies in turn vs. b) build it by filtering by dependencies that list it in their needed_by? Aren't they equivalent?

thomashoneyman: For example, lockfiles like `pnpm`'s list dependencies at their resolved version for every package:
https://lfx.rushstack.io/pages/concepts/pnpm_lockfile

f-f: > if we list the dependencies for each package (including workspace ones), what is the advantage of needed_by?

The list of dependencies is just one layer of the tree, and you'd have to keep doing lookups until you reach the bottom of the tree (i.e. packages with no dependencies). Having a `needed_by` allows for storing the whole plan in there, instead of a single layer.

spago.lock
```diff
@@ -203,7 +218,460 @@ workspace:
         - tuples
       test_dependencies: []
   package_set:
-    registry: 20.0.1
+    address:
+      registry: 20.0.1
+    compiler: ">=0.15.7 <0.16.0"
+    content:
```
natefaubion, line 224: A bit of a drive-by comment, but why does the lockfile need to contain a full copy of the package set?

f-f: I answered this in [this comment above](https://github.com/purescript/spago/pull/1138#discussion_r1421788407), let's keep the discussion in that thread

spago.lock
thomashoneyman: I see. Would the same be achieved by listing the full resolutions for packages instead of just their explicit listed dependencies? That way a build plan for e.g. the `spago-bin` package is already stored as its list of resolutions.

f-f: So you'd have a `resolutions` key for each workspace package listing its build plan?

f-f: Makes sense, let me try it out

bin/src/Main.purs
```diff
@@ -528,7 +547,7 @@ main = do
             logInfo "Set up a new Spago project."
             logInfo "Try running `spago run`"
           Fetch args -> do
-            { env, fetchOpts } <- mkFetchEnv offline args
+            { env, fetchOpts } <- mkFetchEnv offline (Record.insert (Proxy :: _ "isRepl") false args)
```
thomashoneyman, line 550: Any reason for the choice to use `insert` here vs. `merge` on line 562? Arbitrary?

src/Spago/Command/Build.purs
```diff
@@ -42,7 +42,7 @@ type BuildOptions =
   , jsonErrors :: Boolean
   }
 
-run :: forall a. BuildOptions -> Spago (BuildEnv a) Unit
+run :: BuildOptions -> Spago (BuildEnv _) Unit
```
thomashoneyman, line 45: What's the reason to wildcard this? Won't this produce warnings?

thomashoneyman: Same question in the other modules, like `Docs` below.

src/Spago/Command/Publish.purs
```diff
@@ -105,14 +100,12 @@ publish _args = do
   -- As first thing we run a build to make sure the package compiles at all
   runSpago
     -- We explicitly list the env fields because `Record.merge` didn't compile.
-    { getManifestFromIndex: env.getManifestFromIndex
-    , getMetadata: env.getMetadata
-    , workspace: env.workspace { selected = Just selected }
-    , logOptions: env.logOptions
+    -- FIXME: I think we can make it work
```
thomashoneyman, line 103: FIXME?

src/Spago/Registry.purs
```diff
@@ -21,60 +40,201 @@ import Spago.Git as Git
...
   , db :: Db.Db
   | a
+  )
+
+type PreRegistryEnv a = Record (PreRegistryEnvRow a)
+
+type RegistryEnvRow a = PreRegistryEnvRow
+  ( getRegistry :: Spago (PreRegistryEnv ()) RegistryFunctions
+  | a
+  )
```
thomashoneyman, line 57: I'm not sure I follow why we have the "pre-registry" row and then the "registry" row with the extra `getRegistry` function. Would you mind explaining it a little?

src/Spago/Config.purs
```diff
@@ -149,8 +148,8 @@ data Package
 
 -- | Reads all the configurations in the tree and builds up the Map of local
 -- | packages to be integrated in the package set
-readWorkspace :: forall a. Maybe PackageName -> Spago (Git.GitEnv a) Workspace
-readWorkspace maybeSelectedPackage = do
+readWorkspace :: Maybe PackageName -> Boolean -> Spago (Registry.RegistryEnv _) Workspace
```
thomashoneyman, line 151: I'm always a little leery of booleans as arguments; what do you think of either making this a record:

```purs
{ selectedPackage :: Maybe PackageName
, pureBuild :: Boolean
}
```

or a data type

```purs
data PureBuild = NotPure | Pure
```

thomashoneyman:
On light review this looks quite good; I like the introduction of the locking mechanism for access to refreshing the registry. Did you consider debouncing it as well such that we only refresh the registry at most once every N seconds? Maybe that's a premature optimization, but it's something I had to do in the registry so it came to mind.

My main comment has to do with the `needed_by` key vs. just storing resolutions directly as I find the latter a little easier to work with, conceptually. Once you make a decision on that I can review again and test this out.

src/Spago/Repl.purs
```diff
@@ -6,21 +6,21 @@ import Spago.Prelude
...
 supportPackage packageSet = do
-  { getMetadata, logOptions } <- ask
-  case packageSet of
-    PackageSet packages -> pure $ Map.filterWithKey (\k _v -> k == supportPackageName) packages
+  case packageSet.buildType of
+    PackageSetBuild _info packages -> pure $ Map.filterWithKey (\k _v -> k == supportPackageName) packages
     -- TODO: we should look in the "other" packages first
-    Registry _other -> do
-      maybeMetadata <- runSpago { logOptions } (getMetadata supportPackageName)
+    RegistrySolverBuild _other -> do
```
thomashoneyman, line 22: A quick thought on solver vs. package set builds: should `spago init` require you to specify 'app' or 'lib' as an init type such that libraries use ranges off the bat and don't use a package set?

bin/src/Main.purs
f-f: I tend to use `insert` if I'm only dealing with one key, could try to swap it with merge and see if that works (all the record things are quite finicky when it comes to envs, for some reason)

src/Spago/Command/Build.purs
f-f: It will produce warnings, but we should filter them out - I added all the wildcards after chatting with @mikesol (on Discord, [here](https://discord.com/channels/864614189094928394/867149806178664478/1183383103817924698)) about fitting all of our Envs together, and found that using wildcards gets us better inference - I had a few puzzling errors just go away once I swapped them in. (Mike has [a whole blog post about this](https://dev.to/mikesol/metapolymorphic-providers-160p))

src/Spago/Command/Publish.purs
f-f: I was hoping to fix this one too, but I can't find a good way to make it work. So I'll think about it some more or take out this FIXME

src/Spago/Registry.purs
f-f: Yeah, the compiler would try to infer an infinite type if these would run in the same env, so I had to separate them

src/Spago/Config.purs
```diff
@@ -130,9 +124,14 @@ derive newtype instance Eq RemotePackageSet
 
 type PackageMap = Map PackageName Package
 
-data PackageSet
-  = PackageSet PackageMap
-  | Registry PackageMap
+data BuildType
+  = RegistrySolverBuild PackageMap
+  | PackageSetBuild PackageSetInfo PackageMap
```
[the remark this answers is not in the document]
f-f: > Presumably this means you are either building with the solver, ie. you use dependency ranges, or you're building with a package set, ie. you don't use ranges. What happens in the hybrid cases?

There's no hybrid case. If you have a package set then you are building with it, if you don't have one then you are using the solver.

We throw away the ranges when we get a build plan with the package set - see here, we have a `Map PackageName Range`, and call `Map.keys`:
https://github.com/purescript/spago/blob/d7128c76f36a8c27810ff8d8380bd1dff85d9295/src/Spago/Command/Fetch.purs#L399

If the project has a package set the ranges only matter when publishing, where we build with _both_ the package set and the solver (so the ranges _are_ checked, but not very often)

src/Spago/Config.purs
f-f: Sure, either works

src/Spago/Repl.purs
f-f: > should spago init require you to specify 'app' or 'lib' as an init type such that libraries use ranges off the bat and don't use a package set?

We force users that publish their package to specify ranges, so I don't see a reason to enforce this upfront.

f-f:
> Did you consider debouncing it as well such that we only refresh the registry at most once every N seconds? Maybe that's a premature optimization, but it's something I had to do in the registry so it came to mind.

Not a premature optimization, it was in fact the first optimization we had around this, and got to it pretty quickly - it's pretty annoying to have to wait for a `git pull` so often.

We debounce Registry pulls so that they happen every 15 minutes at most, see here:
https://github.com/purescript/spago/blob/d7128c76f36a8c27810ff8d8380bd1dff85d9295/src/Spago/Registry.purs#L286-L309

src/Spago/Repl.purs
f-f: Oh, if you mean "allow users to init a project without a package set", then we already have it: `spago init --use-solver`

bin/src/Main.purs
thomashoneyman: I don't prefer either, it just looked like you were using `merge` with one field below and I didn't know if there was a reason to choose one or another that emerged from your env-wrangling :)

src/Spago/Repl.purs
f-f: The memory of trying to get a Bower project to work and stepping through whole pages or solver errors in bright red ink is still fresh in me - sure, our solver is much more sensible than Bower's, but I feel the main reason the majority of the community is using package sets today is that they are just less complex to deal with and users can concentrate on actually writing the thing instead of dealing with builds. (as a user I would like that to be my experience at least, and I'm even in that small slice of the demographic that _likes_ dealing with builds!)

So yeah, I would prefer to keep package sets as default because they are easy, and have solver builds as the "yes yes I know what I'm about to get into" opt-in option.

f-f:
@thomashoneyman ready for a look again

thomashoneyman:
Looks like the lockfile may need to be regenerated, according to this CI failure?

thomashoneyman:
Looks pretty good! I did find that if I install a package I always seem to get a message about not having a lockfile, even though I do have one.

```
➜ spago install -p bin aff-bus

✅ Selecting package to build: bin

Adding 1 packages to the config in bin/spago.yaml
No lockfile found, generating one...
Lockfile written to spago.lock. Please commit this file.
Downloading dependencies...
Building...
           Src   Lib   All
Warnings     0     0     0
Errors       0     0     0

✅ Build succeeded.
```

thomashoneyman:
Also, it seems that if I run `spago build` in the case my lockfile is incorrect (out of date, missing, etc.) then my workspaces are built. Then, if I run `spago build` again I expect a no-op, but instead my project sources and anything in extra_packages seem to get recompiled a second time. Is there some sort of cache issue there?

f-f:
> Looks like the lockfile may need to be regenerated, according to this CI failure?

Eh, the patch includes a change that moves from a fatal error to a warning - I published a new version (0.93.22), but CI doesn't seem to be picking that one up?

f-f:
> Also, it seems that if I run spago build in the case my lockfile is incorrect (out of date, missing, etc.) then my workspaces are built. Then, if I run spago build again I expect a no-op, but instead my project sources and anything in extra_packages seem to get recompiled a second time. Is there some sort of cache issue there?

I've seen this one happen, but since a long time, so I don't think the lockfile is involved

f-f:
> I did find that if I install a package it is added to my spago.yaml config, but it isn't added to the lockfile and I always get a message that I'm missing a lockfile even though I do have one.

> The lockfile is still not updated and lists aff-bus as one of my package's dependencies even though that's not true, as reflected in the spago.yaml

I'll have a look at these - we have a test that should test this, but it's likely not thorough enough

thomashoneyman:
It doesn't look like 0.93.22 ever made it to NPM:
https://registry.npmjs.org/spago

f-f:
> It doesn't look like 0.93.22 ever made it to NPM

Ah that must be it then! I'll push a new one once I manage to patch the holes you found 😄

f-f:
@thomashoneyman I should have addressed everything now - pushed 0.93.22 with the latest changes, and added a test that checks the lockfile when installing/uninstalling (the scenarios you tried). The patches also fixed up a few other fixtures, which I am happy about

thomashoneyman (approved):

Issue #1139, opened 2023-12-13
Mixed case configuration for publish location


natefaubion:
It looks like the configuration consistently uses snake case everywhere, but there's still some camel case for "location".
<img width="296" alt="Screen Shot 2023-12-13 at 11 24 42 AM" src="https://github.com/purescript/spago/assets/144058/8fc9670e-9082-4b9c-aae3-786e8098c62b">

thomashoneyman:
I agree that we should choose one case or another and standardize on it. I believe the choice so far has been snake case.

I suspect the reason `location` uses camel case is because Spago's yaml parsing first converts to JSON and then reuses the Registry json codecs to parse the result. JSON always uses camel case, so any registry type found in the spago.yaml file is going to end up using camel case.

It seems there are a few ways forward:

1. Accept mixed case in the spago.yaml (I think this is a bad idea because you can't easily reason about what format a given field will take)
2. Standardize on snake case (We would need to convert camel case to snake case and vice versa when using registry types)
3. Standardize on camel case (Registry types would then work as-is).

I think most yaml I've seen in the wild uses camel case and/or dash case, hence why we've done that in Spago. But it seems that there really isn't a canonical YAML format:
https://stackoverflow.com/questions/22771226/what-is-the-canonical-yaml-naming-style

I am personally fine with using camel case throughout the spago.yaml files; it's certainly the easiest option especially since we always operate on the yaml via JSON and are reusing JSON codecs from the registry. But it's really @f-f's call.

JordanMartinez:
I'd prefer camelCase. All those `_` are annoying to type.

f-f:
I have no preference as long as it's consistent

thomashoneyman:
Yes, the registry will have to support both, but only for the 'publish' section of the configuration since that's all it parses.

flip111:
I'd prefer snake case, but i don't mind too much either.

So what will be the final decision?

nate: snake
thomas: snake
jordan: camal
f-f: undecided

Is it like this so far?

f-f:
As Thomas noted, there's no real convention for YAML and it's about the surrounding conventions - PureScript uses CamelCase so that's likely to be the most natural choice.

I'd say we should try out how CamelCase reads and go from there.

thomashoneyman:
I don’t think Nate expressed a preference. I think camel is the most reasonable choice given the JSON round tripping and PureScript as the surrounding context.

PR #1154, opened 2023-12-25
🎅 🐫 - Switch to camel case in config

flip111:
https://github.com/purescript/spago/issues/1139

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ In **all** cases, you'll want to switch to the new Registry package sets, so rep
 
 ```yaml
 workspace:
-  package_set:
+  packageSet:
     url: https://raw.githubusercontent.com/purescript/package-sets/psc-0.15.10-20230919/packages.json
 ```
 
@@ -221,7 +221,7 @@ workspace:
 
 ```yaml
 workspace:
-  package_set:
+  packageSet:
     registry: 41.2.0
 ```
 
@@ -386,7 +386,7 @@ spago run -p package-name --purs-args "--verbose-errors"
 # package:
 #   run:
 #       main: Main
-#       exec_args:
+#       execArgs:
 #         - "arg1"
 #         - "arg2"
 $ spago run -p package-name
@@ -481,14 +481,14 @@ in there, but thankfully I figure how to fix it. So I clone it locally and add m
 
 Now if I want to test this version in my current project, how can I tell `spago` to do it?
 
-There's a section of the `spago.yaml` file just for that, called `extra_packages`.
+There's a section of the `spago.yaml` file just for that, called `extraPackages`.
 
 In this case we override the package with its local copy, which should have a `spago.yaml` - our `workspace` will look something like this:
 
 ```yaml
 workspace:
     registry: 41.2.0
-  extra_packages:
+  extraPackages:
     aff:
       path: ../my-purescript-aff
 ```
@@ -520,7 +520,7 @@ In this case, we can just change the override to point to some commit of our for
 ```yaml
 workspace:
     registry: 41.2.0
-  extra_packages:
+  extraPackages:
     aff:
       git: https://github.com/my-user/purescript-aff.git
       ref: aaa0aca7a77af368caa221a2a06d6be2079d32da
@@ -535,14 +535,14 @@ workspace:
 > [!IMPORTANT]\
 > You still need to `spago install my-new-package` after adding it to the package set, or Spago will not know that you want to use it as a dependency!
 
-If a package is not in the upstream package set, you can add it exactly in the same way, by adding it to `extra_packages`.
+If a package is not in the upstream package set, you can add it exactly in the same way, by adding it to `extraPackages`.
 
 E.g. if we want to add the `facebook` package:
 
 ```yaml
 workspace:
     registry: 41.2.0
-  extra_packages:
+  extraPackages:
     facebook:
       git: https://github.com/Unisay/purescript-facebook.git
       ref: v0.3.0 # branch, tag, or commit hash
@@ -557,7 +557,7 @@ As you might expect, this works also in the case of adding local packages:
 ```yaml
 workspace:
     registry: 41.2.0
-  extra_packages:
+  extraPackages:
     facebook:
       path: ../my-purescript-facebook
 ```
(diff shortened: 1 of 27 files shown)


f-f:
Heh, CI fails because Spago can't build itself anymore - I think we should accept both formats for now, and keep the config in the old format until we have a new version out

flip111:
I think it's a bit too much for the codebase to support two versions of the config. Keeping things backward compatible for alpha status software is not important imo. Isn't there another way too bootstrap the CI build once?

f-f:
We have already bootstrapped in the past so that's not the issue - we are indeed in alpha, but at this point there are quite some packages using this version of spago, and this breaks 100% of them.
We can save some of that inconvenience by supporting the old config as well for a while, e.g. until beta.
This is not wastes effort because we'll eventually have to be able to support multiple versions of the config, once things get more stable.

flip111:
Maybe if it can be either one or the other version of the config. I don't see it working out with mixed versions (differences per property field)

f-f:
> I don't see it working out with mixed versions (differences per property field).

We wouldn't mix versions per field, we'd either accept the current config or the new one, no mix.

> About breaking packages, they can update their yaml file to the new format and it should be good to go no?

The error that spago is throwing is not telling folks that they should upgrade to the new format. Adding such an error message would need detecting the old format, at which point we might as well just accept it.

flip111:
Is there a yaml printer available? We could parse the old format and do the upgrade for them

f-f:
Automatically upgrading the config format is a good idea, but it might take quite a few lines - we have a "yaml printer", but editing the configuration is not just a matter of "write the new content to replace the old content": once we parse the yaml into a PureScript type, we lose all the metadata that was contained into the original file, e.g. comments.
So we need to read the yaml file and keep around the rich representation that contains all this metadata, and whenever we want to edit the configuration we need to reflect back the edits into this format.
As an example, we do this for `spago install some-package`, here:
https://github.com/purescript/spago/blob/b450892d69aa3013861fde6c73299d2c17aba3dc/src/Spago/Config.purs#L507-L510

The above calls [a piece of FFI](https://github.com/purescript/spago/blob/b450892d69aa3013861fde6c73299d2c17aba3dc/src/Spago/Config.js#L10) that traverses the config in the format that the JS `yaml` library uses, and adds the right things in the right place.
If we want to automatically upgrade people's configs (which I think it's desirable) we would use a similar approach.

JordanMartinez:
> The above calls [a piece of FFI](https://github.com/purescript/spago/blob/b450892d69aa3013861fde6c73299d2c17aba3dc/src/Spago/Config.js#L10) that traverses the config in the format that the JS yaml library uses, and adds the right things in the right place.
> If we want to automatically upgrade people's configs (which I think it's desirable) we would use a similar approach.

Is that the last thing blocking this PR (aside from resolving any merge conflicts with `master`)?

flip111:
We need this auto-upgrade the yaml file to camel case and then it's good to go

JordanMartinez:
Are you going to work on that? Or do you need help with knowing how to do that?

f-f:
Yes I think that once we recognise the old config and auto-upgrade it then we don't need to do much else here - we'll want to add a few tests to ensure that we don't regress on parsing old configs and are always able to port them.

JordanMartinez:
I came up with a solution for auto-migrating the case situation.

The below JS code (which allowed for faster iteration in a dynamic context) works using a top-down approach by:
1. getting the old value at a given path
2. deleting that value in the doc
3. adding it back in under a new name

Due to mutating the value, there's one difficulty with this approach. Let's specifically rename `package` and `package.name` by uppercasing them. When we hit the `package` key, we make the following change:
```diff
-package:
-  name: foo
-  description: bar
+PACKAGE:
+  name: foo
+  description: bar
```
What was previously located at `package.name` is now located at `PACKAGE.name`. So, we need to account for this in Step 1 above when getting the original value. Thus, I am currently storing the parent path and the final key part separately so that I can easily map between the two. The below script illustrates this.

`getIn`/`deleteIn`/`addIn` use an array of paths into the document to update that part of the document. `getIn` is slightly different in that one needs to additionally pass in `true`, so that it doesn't unwrap any scalar values.

<details>

<summary>Script</summary>

```js
import Yaml from "yaml";

const doc = Yaml.parseDocument(`
package:
    name: "foo"
    bar: 1
workspace:
    something: 1
`);

const paths =
    [ { parent: [], name: "package" }
    , { parent: ["package"], name: "name" }
    ];

const migrated = doc.clone();
const renameFn = (x) => x.toUpperCase();
paths.forEach(path => {
    const oldFullPath = (() => {
        const p = path.parent.slice().map(renameFn);
        p.push(path.name);
        return p;
    })();
    const newFullPath = (() => {
        const p = path.parent.slice();
        p.push(path.name);
        return p.map(renameFn);
    })();
    if (migrated.hasIn(oldFullPath)) {
        console.log(`\nFound a value at path '${oldFullPath.join(".")}'`);
        const value = migrated.getIn(oldFullPath, true);
        console.log(`Got value`);
        console.log(value);
        console.log("Deleting old value");
        migrated.deleteIn(oldFullPath);
        console.log("Readding original value");
        migrated.addIn(newFullPath, value);
        console.log("Doc is now...");
        console.log(migrated.toJSON());
    } else {
        console.log(`\nNo value found at path ${path.join(".")}`);
    }
});

console.log("\nFinal doc");
console.log(migrated.toJSON());
```

</details>

Output:
```
Found a value at path 'package'
Got value
YAMLMap {
  items: [
    Pair { key: [Scalar], value: [Scalar] },
    Pair { key: [Scalar], value: [Scalar] }
  ],
  range: [ 14, 37, 37 ]
}
Deleting old value
Readding original value
Doc is now...
{ workspace: { something: 1 }, PACKAGE: { name: 'foo', bar: 1 } }

Found a value at path 'PACKAGE.name'
Got value
Scalar {
  value: 'foo',
  range: [ 20, 25, 26 ],
  source: 'foo',
  type: 'QUOTE_DOUBLE'
}
Deleting old value
Readding original value
Doc is now...
{ workspace: { something: 1 }, PACKAGE: { bar: 1, NAME: 'foo' } }

Final doc
{ workspace: { something: 1 }, PACKAGE: { bar: 1, NAME: 'foo' } }
```

Not sure how this should deal with Arrays yet, but we can get the config's keys using type-level programming to produce an array of paths

f-f:
I had a look at #1168 and had a chat with Jordan about the patch - we are both not very happy with the typeclass-based solution - it's very generic but quite noisy and fiddly.

I think a more lightweight approach would be to have a `configV1Codec :: JsonCodec ConfigV1` that we use to read the old config, together with a `foreign migrateV1Config :: YamlDoc ConfigV1 -> YamlDoc Config` function (from `Config.js`) that just does the low-level manipulation of the config.
This is a pattern that we can apply for all the config migrations from here on and, while it's not as guaranteed to be correct for all cases, it's quite simple to handle.

f-f:
Superseded by #1202

Issue #1136, opened 2023-12-07
No packages found due to `spago.yml` instead of `spago.yaml`


JordanMartinez:
I ran `spago init` and then proceeded to manually setup a multi-package repo because no such command exists in spago right now. In doing so, I used the file name `spago.yml` rather than `spago.yaml`. So, `spago build` errored with no packages found. Since `*.yml` AFAIK is a valid yaml file extension, it may be worth notifying the user if a `spago.yml` file exists that they need to use `spago.yaml` instead.

f-f:
The main reason why we have refused to deal with both extensions so far is that it's just a lot of corner cases. E.g. the situation where one has both a `spago.yaml` and a `spago.yml` is nonsensical. So I think we should try to see if a `spago.yml` is there, and tell the user to move it to the `.yaml` extension.

flip111:
Would be good to look for `spago.yml` and then tell the user to rename this file, or offer to rename it

flip111:
The error was on `build` not on `init`. So it could happen on a lot of other commands too that you expect it to put on a wrongly named config file. Should the check go here https://github.com/purescript/spago/blob/master/bin/src/Main.purs#L519 instead of adding the check individually for each command?

f-f:
No, the check would go around the place where we try to read the config(s), so we're able to return a meaningful error (i.e. pointing to the actual path that we were trying to read, etc)

So somewhere around here: https://github.com/purescript/spago/blob/6fe57d88040c5031441b298c4f16a42bf270974a/src/Spago/Config.purs#L157-L159

f-f:
Fixed in #1155

PR #1155, opened 2023-12-25
🎅 Detect yml instead of yaml

flip111:
https://github.com/purescript/spago/issues/1136


core/src/Config.purs
```diff
@@ -481,7 +481,13 @@ readConfig :: forall a. FilePath -> Spago (LogEnv a) (Either String { doc :: Yam
 readConfig path = do
   logDebug $ "Reading config from " <> path
   FS.exists path >>= case _ of
-    false -> pure $ Left $ case path of
-      "spago.yaml" -> "Did not find " <> path <> " Run `spago init` to initialise a new project."
-      _ -> "Did not find " <> path
+    false -> case path of
+      "spago.yaml" -> do
+        hasYml <- FS.exists "spago.yml"
```
klntsky, line 486: It would be better to search for `spago.yml` in the same directory where the config file (`path`) is located.

core/src/Config.purs
```diff
@@ -481,7 +481,13 @@ readConfig :: forall a. FilePath -> Spago (LogEnv a) (Either String { doc :: Yam
...
+        pure $ Left $
+          if hasYml then
+            "Spago's default configuration file must be called spago.yaml The file spago.yml was found, rename it if you will, or run `spago init` to initialise a new project."
+          else
+            "Did not find spago.yaml Run `spago init` to initialise a new project."
+      _ -> pure $ Left $ "Did not find " <> path
```
klntsky, line 492: These two cases (lines 491-492) are functionally the same. Use pattern matching on two variables or `case` with guards to combine them

core/src/Config.purs
```diff
@@ -481,7 +481,13 @@ readConfig :: forall a. FilePath -> Spago (LogEnv a) (Either String { doc :: Yam
...
   FS.exists path >>= case _ of
-    false -> pure $ Left $ case path of
-      "spago.yaml" -> "Did not find " <> path <> " Run `spago init` to initialise a new project."
-      _ -> "Did not find " <> path
+    false -> case path of
+      "spago.yaml" -> do
+        hasYml <- FS.exists "spago.yml"
+        pure $ Left $
+          if hasYml then
+            "Spago's default configuration file must be called spago.yaml The file spago.yml was found, rename it if you will, or run `spago init` to initialise a new project."
```
klntsky, line 489:
```suggestion
            "Spago's default configuration file must be called spago.yaml, not spago.yml. Rename it or run `spago init` to initialize a new project."
```

f-f (commented):
We should add a few tests with fixtures that verify that we are returning the right error messages at the right time

core/src/Config.purs
f-f: Spago tends to use the British spellings (`initialise`) over the american ones 😄

core/src/Config.purs
f-f: Indeed, we use this function also when looking for workspace packages, where it's possible that the user has used the `.yml` extension.

I would use [`String.stripSuffix`](https://pursuit.purescript.org/packages/purescript-strings/6.0.1/docs/Data.String#v:stripSuffix) to detect if the config we're looking for has a `.yml` extension, instead of matching only on the root config

src/Spago/Config.purs
```diff
@@ -155,7 +155,7 @@ readWorkspace { maybeSelectedPackage, pureBuild } = do
   -- First try to read the config in the root. It _has_ to contain a workspace
   -- configuration, or we fail early.
   { workspace, package: maybePackage, workspaceDoc } <- Core.readConfig "spago.yaml" >>= case _ of
-    Left err -> die [ "Couldn't parse Spago config, error:\n  " <> err, "The configuration file help can be found here https://github.com/purescript/spago#the-configuration-file" ]
+    Left err -> die [ "Couldn't parse Spago config, error:\n  " <> err, ". The configuration file help can be found here https://github.com/purescript/spago#the-configuration-file" ]
```
f-f, line 158: Why this change? The new string will go in a new line, so this is going to look weird. Let's add a test with a fixture to verify how it looks like

flip111: This should be reverted

core/src/Config.purs
```diff
@@ -481,7 +483,21 @@ readConfig :: forall a. FilePath -> Spago (LogEnv a) (Either String { doc :: Yam
 readConfig path = do
   logDebug $ "Reading config from " <> path
   FS.exists path >>= case _ of
-    false -> pure $ Left $ case path of
-      "spago.yaml" -> "Did not find " <> path <> " Run `spago init` to initialise a new project."
-      _ -> "Did not find " <> path
+    false -> do
+      yml <- case map (_ <> ".yml") $ String.stripSuffix (String.Pattern ".yaml") path of
+        Nothing -> pure Nothing
+        Just y -> do
```
klntsky, line 489: `traverse`

core/src/Config.purs
```diff
@@ -481,7 +483,21 @@ readConfig :: forall a. FilePath -> Spago (LogEnv a) (Either String { doc :: Yam
...
+          hasYml <- FS.exists y
+          pure $
+            if hasYml then
+              Just y
+            else
+              Nothing
```
klntsky, line 495:
```suggestion
            pure $ guard hasYml *> pure y
```

core/src/Config.purs
```diff
@@ -481,7 +483,21 @@ readConfig :: forall a. FilePath -> Spago (LogEnv a) (Either String { doc :: Yam
...
+      let
+        message = case yml of
+          Nothing -> "Did not find " <> path
+          Just y -> "Did not find " <> path <> ". Spago's configuration files should end with .yaml. Try renaming " <> y
+      pure $ Left $ case path of
+        "spago.yaml" -> message <> "Run `spago init` to initialize a new project."
```
klntsky, line 501:
```suggestion
        "spago.yaml" -> message <> " Run `spago init` to initialize a new project."
```

core/src/Config.purs
f-f: I find `guard` harder to read than the if/else above, let's keep that one

JordanMartinez:
CI may fail because I'm not sure what other tests my be affected by these 'pretty print' changes.

core/src/Log.purs
```diff
@@ -98,10 +98,14 @@ instance Loggable String where
...
-instance Loggable a => Loggable (Array a) where
+instance Loggable (Array Docc) where
+  toDoc = Log.lines
+else instance Loggable a => Loggable (Array a) where
   toDoc = Log.lines <<< map toDoc
 
-instance Loggable a => Loggable (NonEmptyArray a) where
+instance Loggable (NonEmptyArray Docc) where
+  toDoc = Log.lines <<< NEA.toArray
+else instance Loggable a => Loggable (NonEmptyArray a) where
```
f-f, line 108: Why these instance chains? They seem unnecessary

src/Spago/Command/Fetch.purs
```diff
@@ -377,7 +378,12 @@ getPackageDependencies packageName package = case package of
       Right { yaml: { package: Just { dependencies: (Dependencies deps) } } } -> do
         pure (Just (map (fromMaybe Config.widestRange) deps))
       Right _ -> die [ "Read valid configuration from " <> configLocation, "However, there was no `package` section to be read." ]
-      Left err -> die [ "Could not read config at " <> configLocation, "Error: " <> err ]
+      Left errLines -> die
+        [ Log.text $ "Could not read config at " <> configLocation
+        , Log.text "Error: "
+        -- type annotation needed here to clarify that this is `Doc GraphicsParam`, not `Doc a`
+        , Log.indent $ Log.lines $ map Log.text errLines :: Docc
+        ]
```
f-f, line 386: The rest of the codebase uses `toDoc` rather than things like `Dodo.text`, I think it works better for inference?

Also note that we are re-exporting `indent` from the Prelude, so most of the times you don't need to import Dodo at all

core/src/Config.purs
```diff
@@ -479,25 +479,30 @@ legacyPackageSetEntryCodec = CA.object "LegacyPackageSetEntry"
...
+        "spago.yaml", Nothing -> 
+          [ "Did not find `" <> path <> "`. Run `spago init` to initialize a new project." ]
+        "spago.yaml", Just y ->
+          [ "Did not find `" <> path <> "`. Spago's configuration files must end with `.yaml`, not `.yml`. "
+          , "Try renaming `" <> y  <> "` to `" <> path <> "` or run `spago init` to initialize a new project."
+          ]
+        _, Nothing ->
+          ["Did not find `" <> path <> "`."]
+        _, Just y -> 
+            [ "Did not find `" <> path <> "`. Spago's configuration files must end with `.yaml`, not `.yml`. "
```
flip111: This duplicates code, previously it was possible to have only one source line with this message

src/Spago/Command/Fetch.purs
JordanMartinez: Ah, that's what I needed to do. Thanks. This change was largely just to fix the inference stuff.

core/src/Log.purs
JordanMartinez: It just removes an unnecessary `map` when we know what the `a` is. I've removed them.

core/src/Config.purs
JordanMartinez: Correct. But...

The message (per your original code) was outputting:
> Try renaming spago.yml Run `spago init` to initialize a new project.

I thought it would be better to put an `or` in between those two sentences when the context made sense. In some cases, one is likely the correct choice. And rather than write something weird like:
```
  fold $ Array.catMaybes
    [ Just ....
    , if cond1 then Just "Try renaming spago.yaml to..." else Nothing
    , if cond2 then Just "Run spago init" else Nothing
    , if cond3 then Just "Try ... or run ..." else Nothing
    ]
```
I thought it would be more readable to just duplicate the code. This isn't a performance-critical part of the codebase, so I don't think the duplication is harmful.

f-f: Yeah I agree that this change is good

src/Spago/Config.purs
```diff
@@ -188,7 +193,7 @@ readWorkspace { maybeSelectedPackage, pureBuild } = do
       -- We try to figure out if this package has tests - look for test sources
       hasTests <- FS.exists (Path.concat [ Path.dirname path, "test" ])
       pure $ case maybeConfig of
-        Left e -> Left $ "Could not read config at path " <> path <> "\nError was: " <> e
+        Left eLines -> Left $ "Could not read config at path " <> path <> "\nError was: " <> Array.intercalate "\n" eLines
```
f-f, line 196: The `intercalate` here is a bit of a smell because it unnecessarily cares about formatting the error message returned from above - we should return a `Docc` instead of a `Array String` from the `readConfig` function

f-f (commented):
One small thing then this is good to go

f-f (approved):

Issue #1110, opened 2023-11-01
Publish rebuilds with another version


flip111:
![Screenshot from 2023-11-01 22-21-15](https://github.com/purescript/spago/assets/2244480/be382f3d-7816-4983-865f-d7aed179f4ae)

normal `npx spago build` works fine

f-f:
The `publish` command is rebuilding your package using the Registry solver, which other users will use when interacting with your package.

As you noted in the screenshot, the version bounds in the configuration allow for the version that is being used in the rebuild.

I believe the logic we are using for inferring the bounds is correct (assuming that all the patch version in a minor release are not going to introduce breakage), and I believe the SemVer of that package is not correct since it introduced breakage.

You should edit your configuration to restrict the bounds from using the breaking version.

flip111:
> I believe the logic we are using for inferring the bounds is correct

I also think that is correct. But the rebuild is failing. To rebuild version 0.4.2 should be downloaded. Possibly we should even consider to rebuild in a temporary directory not to mess with the main cache

> I believe the SemVer of that package is not correct since it introduced breakage

Which package? arrays-extra or foldable-traversable-extra ?

> You should edit your configuration to restrict the bounds from using the breaking version.

This is not the problem because in the new version that function is also there https://pursuit.purescript.org/packages/purescript-arrays-extra/0.4.2/docs/Data.Array.Extra.First#v:modifyOrSnoc the problem is is that the source file is not available for the compiler

f-f:
Ah, I misread the error in the screenshot - indeed everything else is correct, except that Spago is not installing the missing packages. We should be fetching them before issuing a rebuild.

flip111:
Should [this](https://github.com/purescript/spago/blob/master/src/Spago/Command/Fetch.purs#L78) be called from [here](https://github.com/purescript/spago/blob/master/src/Spago/Command/Publish.purs#L339-L340)?

f-f:
Probably? That would be the first thing I would try. Then I would [add a test](https://github.com/purescript/spago/blob/master/test/Spago/Publish.purs) and see how it goes. (tip for tests: use `itOnly` to only run a single test instead of having to run the whole test suite)

flip111:
`Fetch.run` is only called from Main so far. And it doesn't take a package name as direct argument. Probably it's there in some reader/state monad. That's why i didn't know if it would be a good idea to call run directly or rather if it needs to be refactored in separate functions.

f-f:
[...]

PR #1157, opened 2023-12-25
🎅 Fetch before rebuild

flip111:
Fix https://github.com/purescript/spago/issues/1110


f-f:
I don't think this is fixing the issue - the dependencies from the build plan are not passed in anywhere, and these are the ones we need to fetch.

A good first step for a fix would be to write a test that replicates the issue in #1110. After we get CI to fail because of that we can then write the correct patch.

flip111:
I'm not a fan of ttd i rather write the patch first if i go ahead with this. Should `Fetch.run` called multiple times for each dependency or can i pass all the deps at once?

f-f:
This is not about TTD, but rather about replicating the failure in a standard environment (CI). If we can't replicate we can't fix.

flip111:
Can the registry be queried as to what is the latest version of a package?

f-f:
Why would we need that? We'll need to build with whatever versions the solver returns, which is not always going to be the latest one.

flip111:
As new versions get added to the registry, the solver might resolve to something else. I expect the test to start breaking.

f-f:
> I expect the test to start breaking.

It's entirely possible to have a test that doesn't break over time - we'll need to set the version bounds to be different than what the package set contains.

flip111:
In #1110 i don't think it was relevant that an outdated package `0.4.1` was available in the cache. The problem should just reproduce when removing any package from the cache which then should be refetched before rebuild. That the problem occurs due to an upgraded version of a dependency is not relevant to test the solution.

flip111:
The test passes but i don't think it works correctly. Not sure what is happening

f-f:
This has been discussed on Discord, but I'll report here as well:
Spago uses a global cache and I think that's ultimately the reason why the new test is not working - you are deleting the package from the local cache, but Spago will find it in the global cache so it just copies it back there.
The user-facing command to get the location of the global cache is `spago ls paths`, but in tests you can use the `Spago.Paths` module

flip111:
Superseed by https://github.com/purescript/spago/pull/1283

PR #1161, opened 2024-01-12
Fix missing git dependencies when building monorepos

f-f:
Our build at work uncovered a bug that seems to trigger only in very specific condition.

This PR contains:
- a test that replicates the bug (and the first commit here shows that CI breaks on it)
- the fix for said bug (that makes the test suite happy again)

I suspect the bug was likely introduced as a side effect of #1018 and exposed by the work in #1138, and it's due to improper glob handling during the build: we are constructing the globs for the whole build when querying the graph, and reusing the same globs for the build. However, since #1138 we do not fetch all the dependencies prior to building, so when building a single package in a monorepo, we could stumble on dependencies that have not been fetched yet (because the fetch step will only fetch the dependencies that we need at the moment).
This could in principle happen with any kind of dependency that we fetch from the internet, but doesn't reliably happen with Registry packages, because we heavily cache them. Fetching a git repo provides a reliable repro instead.

The fix in this patch prepares two sets of globs, one for `purs graph` and one for `purs compile`. This seems to fix things, but I'm not entirely convinced it's the most sensible thing we could do. Maybe we should always fetch all dependencies instead, even when building a single package in a monorepo? Something to ponder, I think the current fix works fine for now.

### DIFF
--- a/src/Spago/Command/Build.purs
+++ b/src/Spago/Command/Build.purs
@@ -101,12 +101,6 @@ run opts = do
     selectedPackages = case workspace.selected of
       Just p -> NEA.singleton p
       Nothing -> Config.getWorkspacePackages workspace.packageSet
-    graphGlobs = getBuildGlobs
-      { dependencies: allDependencies
-      , depsOnly: opts.depsOnly
-      , withTests: true
-      , selected: selectedPackages
-      }
     globs = getBuildGlobs
       { dependencies: case workspace.selected of
           Just p -> unsafeFromJust $ Map.lookup p.package.name dependencies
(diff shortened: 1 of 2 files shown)


f-f:
Yes indeed - I just stumbled on this but with `--pedantic-packages` instead of `--strict`. I'll put together a new test and look for a fix

f-f:
@JordanMartinez I got rid of the different set of globs, and just restricted to the package globs if we are building with a single package. I also updated the test accordingly, so this should be good for another look

JordanMartinez (approved):

PR #1164, opened 2024-01-13
Document the bundler-args argument

### LINKED ISSUE #1163: Support extra esbuild argument for GJS
Hello, when running purescript with https://gjs.guide/ , the FFI needs such binding:

```js
import GLib from 'gi://GLib';
```

Which makes esbuild fail with:

```
✘ [ERROR] Could not resolve "gi://GLib"

    output/GLib.Variant/foreign.js:4:17:
      4 │ import GLib from 'gi://GLib';
        ╵                  ~~~~~~~~~~~

  You can mark the path "gi://GLib" as external to exclude it from the bundle, which will remove
  this error.
```

To make this work, we need to pass the following arguments: `"--external:gi://*"  "--external:resource://*"`.
Would it be possible to do that with spago, e.g. with a "ESBUILD_EXTRA_ARGS" env or something?

TristanCacqueray:
Fixes #1163

### DIFF
--- a/README.md
+++ b/README.md
@@ -952,19 +952,14 @@ $ node -e "import('./index.js').then(m => console.log(m.main))"
 [Function]
 ```
 
-If the build fails with `✘ [ERROR] Could not resolve "better-sqlite3"`, you can pass extra arguments to esbuild with:
-
-- the `bundler-args` argument, for example: `--bundler-args "--external:better-sqlite3"`.
-
-… or by using the following configuration:
-
-```yaml
-package:
-  bundle:
-    extra_args:
-      - "--external:better-sqlite3"
-```
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -952,6 +952,19 @@ $ node -e "import('./index.js').then(m => console.log(m.main))"
...
+
+… or by using the following configuration:
+
+```yaml
+package:
+  bundle:
+    extra_args:
+      - "--external:better-sqlite3"
+```
+
```
f-f, line 967: Slight rewording to make it more general and more consistent formatting:
```suggestion
Spago does not wrap the entirety of the bundler's API (esbuild for JS builds), so it's possible to pass arguments through to it. E.g. to exclude an NPM package from the bundle you can pass the `--external` flag to esbuild:
- either through the command line, with the `--bundler-args` flag, i.e. `--bundler-args "--external:better-sqlite3"`.
- or by adding it to the configuration file:
  ```yaml
  package:
    bundle:
      extra_args:
        - "--external:better-sqlite3"
  ```

```

f-f (approved):
Thanks!

Issue #1165, opened 2024-01-14
Issue warnings when the configuration has unrecognised fields


f-f:
As noted e.g. in #1159, the fact that we just ignore fields that we don't parse means that it's possible to get the names of the fields wrong and not know about it.

I am not sure how the fix for this one would look like - I think either we'll end up doing some low-level parsing with the `yaml` library in [Config.js](https://github.com/purescript/spago/blob/68b644644faf38f2d2ccbee7b02758de60a8f1c7/src/Spago/Config.js), or hopefully we can do something [in the Codec at the PS level](https://github.com/purescript/spago/blob/68b644644faf38f2d2ccbee7b02758de60a8f1c7/core/src/Config.purs#L66)

thomashoneyman:
The relevant registry PR is merged, but #1272 isn't, so I'm reopening this.

PR #1169, opened 2024-01-18
Fix readme bug on specifying test dependencies

flip111:
Config doesn't work with `tests` and also because of this https://github.com/purescript/spago/issues/1165 it's not clear


f-f (approved):
Thanks!

PR #1186, opened 2024-02-09
Fix issue with git clone: tags would get pulled twice due to the lockfile storing hashes

f-f:
We added support for lockfiles in #1138, but that introduces a series of quirks that we're uncovering over time - today's one is that if one pulls a git repo as a dependency, we store the commit hash in the lockfile, regardless of the hash/branch/tag that was used in the config.
This caused Spago to fetch a repo twice then specifying a tag, first using the tag as git ref, then the second time with the commit hash, after reading the lockfile and noticing that no repo was available with that commit hash.

This patch fixes it by first cloning the repo, then reading the commit hash, and eventually copying the folder to what would be fetched when the commit hash was specified.


JordanMartinez (approved):
Looks fine to me.

PR #1188, opened 2024-02-26
Fixing parsing of execArgs flag

imsaravana369:
### Description of the change

TODO

### Checklist:

- [ ] Added the change to the "Unreleased" section of the changelog (where to add this?)
- [ x] Added some example of the new feature to the `README`
- [x] Added a test for the contribution (if applicable) (will write one)

### DIFF
--- a/bin/src/Flags.purs
+++ b/bin/src/Flags.purs
@@ -191,15 +191,13 @@ pursArgs =
       )
 
 execArgs :: Parser (Maybe (Array String))
-execArgs =
-  OT.optional
-    $ Array.fromFoldable
-    <$> O.many
-      ( O.strArgument
-          ( O.help "Arguments to pass to the running script"
-              <> O.metavar "ARGS"
-          )
-      )
+execArgs = (\s -> if List.null s then Nothing else (Just $ Array.fromFoldable s))
+  <$> O.many
+    ( O.strArgument
+        ( O.help "Arguments to pass to the running script"
+            <> O.metavar "ARGS"
+        )
+    )
 
 backendArgs :: Parser (List String)
 backendArgs =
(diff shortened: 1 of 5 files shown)


bin/src/Flags.purs
```diff
@@ -191,9 +191,7 @@ pursArgs =
       )
 
 execArgs :: Parser (Maybe (Array String))
-execArgs =
-  OT.optional
-    $ Array.fromFoldable
+execArgs = (\s -> if List.null s then Nothing else (Just $ Array.fromFoldable s))
```
JordanMartinez, line 194: Just a thought, but why not just use `Array String` as the result here rather than `Maybe (Array String)`? `[]` is essentially the same as `Nothing`.

JordanMartinez: Ah, just saw the issue. `Maybe` indicates whether this should override the configuration file's value.

f-f:
Do we have any tests for this flag? If not, we should add one

bin/src/Flags.purs
imsaravana369: yeah I was also thinking that whether it would be better if we append both of them, that i have my default args in my spago.yaml, still I can pass extra arguments explicitly while doing `spago run`... but it's possible that it might cause confusion if users are not aware of this behaviour...

imsaravana369:
> Do we have any tests for this flag? If not, we should add one

yeah will add tests

bin/src/Flags.purs
f-f: Yes, the general convention in Spago is that any flag provided on the command line overrides whatever is in the config, so we shouldn't use both.

imsaravana369: got it.

imsaravana369:
> Do we have any tests for this flag? If not, we should add one

Added testcases.

f-f (approved):
Thanks!

PR #1190, opened 2024-03-01
Document how to change global cache dir location

### LINKED ISSUE #751: Document how to override the location of the global cache

JordanMartinez:
### Description of the change

Fixes #751.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1495,6 +1495,16 @@ When using package sets it is disabled by default - since we already get a stabl
 
 You can enable it manually by adding a `lock: true` field to the `workspace` section of your `spago.yaml` file, and that will keep it on regardless of which solving mode you're using.
 
+### File System Paths used in Spago
+
+Run `spago ls paths` to see all paths used by Spago. But in general, Spago utilizes two main directories for every project:
+- the global cache directory
+- the local cache directory
+
+The global cache directory is located at `${XDG_CACHE_HOME}/spago-nodejs`. It's location can be changed by configuring the [`XDG_CACHE_HOME`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) environment variable.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1495,6 +1495,16 @@ When using package sets it is disabled by default - since we already get a stabl
 
 You can enable it manually by adding a `lock: true` field to the `workspace` section of your `spago.yaml` file, and that will keep it on regardless of which solving mode you're using.
 
+### File System Paths used in Spago
+
+Run `spago ls paths` to see all paths used by Spago. But in general, Spago utilizes two main directories for every project:
+- the global cache directory
+- the local cache directory
+
+The global cache directory is located at `${XDG_CACHE_HOME}/spago-nodejs`. It's location can be changed by configuring the [`XDG_CACHE_HOME`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) environment variable.
```
f-f, line 1504: This is not the whole story - as I noted [in this comment](https://github.com/purescript/spago/issues/751#issuecomment-1772844736) there are a few more details:
- on Linux, I had to move the `XDG_RUNTIME_DIR` on a NixOS system for things to work, as the `XDG_CACHE_DIR` was not defined. However, I couldn't find any reference to this fact either in the XDG spec nor in the [source of env-paths](https://github.com/sindresorhus/env-paths)
- the location mentioned here only applies on Linux. [macOS and Windows locations are different](https://github.com/sindresorhus/env-paths?tab=readme-ov-file#pathscache)

f-f (approved):
Thanks!

PR #1194, opened 2024-03-07
Trimming the prefix_by message while censoring warning

imsaravana369:
One of the ways to censor warnings is by using `by_prefix`. Some warning messages have additional leading whitespaces so the filtering is not happening correctly, so this PR fixes this by doing the filtering after trimming the text.

TODO

### Checklist:

### DIFF
--- a/test/Spago/Build.purs
+++ b/test/Spago/Build.purs
@@ -2,6 +2,7 @@ module Test.Spago.Build where
 
 import Test.Prelude
 
+import Data.Foldable (fold)
 import Node.FS.Aff as FSA
 import Node.Path as Path
 import Node.Platform as Platform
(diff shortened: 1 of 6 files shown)


imsaravana369:
Should we give support for regex also?

f-f:
I'm not sure what this is addressing - do you have a concrete example?

imsaravana369:
So I have this in my `spago.yaml` 
```
    censor_project_warnings:
        - by_prefix: "The Name"
```
to censor a warning like,
```Name a1 was introduced but not used.
  in value declaration main
  ```
  
  but the issue is somehow the warning  message is padded with leading whitespace, so I had to change `by_prefix` in such a way it has the leading whitespace(matching the error message).

**Comment by f-f** [comment]
Ah that makes sense!
I don't think we should support regexes as of now.

Could you add a test for this?

**Comment by imsaravana369** [comment]
yes sure will add tests

**Comment by imsaravana369** [comment]
Added testcase

File: test/Spago/Build.purs [hunk 1]
```diff
@@ -100,6 +100,23 @@ spec = Spec.around withTempDir do
...
+          { src : fixture "censor-warnings.purs" 
+          , dst : srcMain
+          } 
+       FSA.unlink spagoYaml 
+       FS.copyFile 
+        { src : fixture "censor-warnings.yaml"
+        , dst : spagoYaml
+        }
+       spago [ "build" ] >>= shouldBeSuccess
+
```
f-f, line 119: A few notes:
- if you move the fixture files into a directory that is structures as a spago project, then you can use `FS.copyTree` to just copy the project in the test directory, so you don't have to call `init` and do this file juggling. See [the fixtures for the monorepo tests for an example](https://github.com/purescript/spago/tree/master/test-fixtures/monorepo)
- the test is not checking that things are properly censored, just asserting that the build succeeds. We should check that we are not getting these warnings - see this test in the monorepo suite that is doing a similar thing:

https://github.com/purescript/spago/blob/fdaa7e63b1357cc7309db75a18d99421756de13b/test/Spago/Build/Monorepo.purs#L124-L135

imsaravana369: okay this one looks neat, will do this👍

imsaravana369: Made the changes in the test cases, can you check once?

f-f (approved):
Thanks!

PR #1199, opened 2024-03-12
Fix output truncation with json-errors and build failure. Fix #1160

### LINKED ISSUE #1160: json is cut off
In php i run

```php
<?php

exec('spago build --json-errors --quiet', $output);
```

The json line is cut off after 65536 characters. I don't believe it's a php problem because when i use `cat` on a file with more characters the line is longer. When i run spago directly in the terminal the problem does not manifest. I believe it has something to do with the way nodejs writes to the output buffer.

When you google for `nodejs stdout cut off` quite a few old but i guess still relevant github issues pop-up. Some write actions can be asynchronous and are not guaranteed to finalize the nodejs process terminates.

nwolverson:
Problematic situation is when --json-errors is passed (output is one very long line printed out all at once), there are many warnings (json > 65536 bytes) and there is an error (spago should exit with an error code). Set the error code and allow control flow to naturally exit to allow output to be flushed.

### Description of the change

TODO

### Checklist:


nwolverson:
Not clear why the test fails on windows only, I suspect something about trailing newline (I noticed there's some munging of that in the tests setup, and the output being single-line)

test/Spago/Build.purs
```diff
@@ -161,6 +161,22 @@ spec = Spec.around withTempDir do
...
+        srcWarn = Path.concat [ "src", "Warn.purs" ]
+      FSA.unlink srcMain
+      FS.copyFile
+        { src: fixture "build-json-truncated-main.purs"
+        , dst: srcMain
+        }
+      FS.copyFile
+        { src: fixture "build-json-truncated-warn.purs"
+        , dst: srcWarn
+        }
```
f-f, line 177: If you move the fixtures into their own `build-json-truncated` directory and structure it like a spago project then you can just use `copyTree`, see e.g. the monorepo tests:
https://github.com/purescript/spago/blob/ebb99a02dacd119e8b650855bfae308798c46c07/test/Spago/Build/Monorepo.purs#L117-L120

src/Spago/Command/Build.purs
```diff
@@ -137,26 +138,32 @@ run opts = do
...
+    unless (Array.null pedanticPkgs || opts.depsOnly) do
+      logInfo $ "Looking for unused and undeclared transitive dependencies..."
+      eitherGraph <- Graph.runGraph globs opts.pursArgs
+      graph <- either die pure eitherGraph
+      env <- ask
+      checkResults <- map Array.fold $ for pedanticPkgs \(Tuple selected options) -> do
+        Graph.toImportErrors selected options
+          <$> runSpago (Record.union { selected, workspacePackages: selectedPackages } env) (Graph.checkImports graph)
+      unless (Array.null checkResults) do
+        die $ Graph.formatImportErrors checkResults
```
f-f, line 165: Should this be a `prepareToDie` as well?

f-f:
And about the Windows failure in CI: the fixture contains a `"filename":"src/Main.purs"`, which becomes a `"filename":"src\\Main.purs"` on Windows.

When fixtures contain paths we just keep a different one around just for Windows, e.g. see the tests for `spago sources`:
https://github.com/purescript/spago/blob/ebb99a02dacd119e8b650855bfae308798c46c07/test/Spago/Sources.purs#L36-L38

src/Spago/Command/Build.purs
nwolverson: Yes, this could be, I just didn't think hard enough to understand and restructure this code. I guess the types will make it clear enough

src/Spago/Command/Build.purs
```diff
@@ -153,17 +151,20 @@ run opts = do
         let reportTest = pedanticPackages || (fromMaybe false $ p.package.test >>= _.pedantic_packages)
         Alternative.guard (reportSrc || reportTest)
         pure $ Tuple p { reportSrc, reportTest }
-    unless (Array.null pedanticPkgs || opts.depsOnly) do
+    if Array.null pedanticPkgs || opts.depsOnly then do
```
nwolverson, line 154: well this clearly wasn't it

f-f: I guess it should be the other way around 😄 
```
if Array.null pedanticPkgs || opts.depsOnly then do
  pure true
else do
```

nwolverson: yeah that's what I meant by the comment but I still have 2 spago docs tests failing

f-f: Linux CI is happy!

f-f (approved):

PR #1202, opened 2024-03-15
Switch config to camelCase and support automatic migrations

### LINKED ISSUE #1139: Mixed case configuration for publish location

f-f:
This PR moves the config format to be consistently camelCase, to fix #1139 

As we figured in #1154, this is a big change for the few projects out there that already use the new spago, so we needed to (a) be backwards compatible, and (b) support automatic migration of the config.

The previous attempt to support this was #1168, but that got convoluted enough that we agreed on a simpler approach - here we manipulate the YAML doc entirely on the JS side, and write it to disk if anything was changed, and the user requested it (otherwise we just warn about the config format being out of date)


src/Spago/Config.js
```diff
@@ -99,3 +99,25 @@ export function addRangesToConfigImpl(doc, rangesMap) {
...
+          hasChanged = true;
+          return match.charAt(1).toUpperCase();
+        });
+      }
+    }
+  });
+  if (hasChanged) {
+    return doc;
+  }
+}
```
JordanMartinez, line 123: So, this effectively disables snake_case entirely from the config forever. Moreover, even if someone decided to add a superfluous key to the YAML file in that particular case convention, this script will force it to be camel-case, even if that's not what the user wants. I think that's fine.

JordanMartinez (approved):

src/Spago/Config.js
f-f: > if someone decided to add a superfluous key to the YAML file in that particular case convention, this script will force it to be camel-case, even if that's not what the user wants

Once we find a way to go at #1165 then I think we could effectively disallow superfluous fields, they only seem to cause confusion (see the comment I left in the Config module)

Issue #1182, opened 2024-01-26
Globbing the whole workspace could be faster


Blugatroff:
`spago build` takes surprisingly long to read it's configuration files.

When using my somewhat older laptop i noticed, that an invocation of `spago build` on my smallish AdventOfCode project took 3.5 seconds, even though everything had already been compiled.
I tried investigating what took so long by inserting `logDebug` and using the `--verbose` flag.

I noticed that running `purs` isn't actually the single most expensive step (took 800ms), instead [globbing the workspace](https://github.com/purescript/spago/blob/bbe37b6cd497aa544bd0761fa7a56a5f5d002a87/src/Spago/Config.purs#L172) for every `spago.yaml` and [parsing `spago.lock`](https://github.com/purescript/spago/blob/bbe37b6cd497aa544bd0761fa7a56a5f5d002a87/src/Spago/Config.purs#L253) takes quite long.
I usually expect a package manager to finish very quickly when there is no actual work to be done, even on slightly older hardware.
The old spago just takes 1.5s on the same project and device, so is this just the cost of writing spago in a language that compiles to javascript, or can we reduce this startup time? Or perhaps some of these steps can be skipped in certain cases?

How I measured:
`time node ~/clones/spago/bin/bundle.js build --verbose` shows:
```
[      25ms] CWD: /home/jasper/projects/pure-advent
[      50ms] Global cache: "/home/jasper/.cache/spago-nodejs"
[      52ms] Local cache: "/home/jasper/projects/pure-advent/.spago"
[     112ms] DB: Connecting to database at /home/jasper/.cache/spago-nodejs/spago.v1.sqlite
[     122ms] Reading Spago workspace configuration...
[     122ms] Reading config from spago.yaml
[     164ms] Before Glob.match **/spago.yaml                    I Added this logDebug
[     698ms] After Glob.match **/spago.yaml                       Took about 550ms
[     701ms] Selecting package aoc from ./
[     702ms] Before FS.readYamlFile                             I Added this logDebug
[    1540ms] After FS.readYamlFile                               Took about 800ms
[    1541ms] Lockfile is up to date, using it
[    1542ms] Found lockfile, using the package set from there
[    1548ms]
✅ Selecting package to build: aoc

[    1549ms] Package path: ./
[    1550ms] Requested to install these packages: []
[    1556ms] Downloading dependencies...
[    1619ms] Building...
[    1655ms] Running command:
[...]

PR #1209, opened 2024-04-02
Faster globbing by respecting root .gitignore.

Blugatroff:
Addresses part of #1182 by implementing this [snippet](https://github.com/sindresorhus/globby/issues/50#issuecomment-467897064).

I'm not sure, whether this is the right place for the change, would it have been better to implement this [purescript/registry-dev](https://github.com/purescript/registry-dev) since that's where Glob.match' is implemented? Although then again, Glob.match' is just a wrapper for fast-glob.

On my laptop and repo mentioned in #1182 this change reduces the time spent globbing for spago.yaml from 600ms to 30ms :)

### Description of the change
Instead of globbing everything for **/spago.yaml and then filtering the matches against .gitignore files, use every entry in the root .gitignore as a negative match (ignore) while globbing.

The resulting matches still need to be checked seperately using `Git.isIgnored`, because the new function only considers the root .gitignore file, .gitignore files in subdirectories are ignored.

### Checklist:

### DIFF
--- a/src/Spago/Config.purs
+++ b/src/Spago/Config.purs
@@ -163,6 +163,31 @@ type ReadWorkspaceOptions =
   , migrateConfig :: Boolean
   }
 
+-- | Same as `Glob.match'` but if there is a .gitignore file in the same directory,
+-- | then the `ignore` option will be filled accordingly.
+-- | This function does not respect any .gitignore files in subdirectories.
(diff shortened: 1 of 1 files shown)


f-f (approved):
This is great! Thanks @Blugatroff 👏

Issue #1215, opened 2024-04-15
Publish packages with dependencies


CGenie:
Hello,

Suppose I'm developing 2 libraries at once and one of them depends on the other. After publishing the first one (without the dependency), it seems I can't publish the second one until a new package set is published the next day. I can't publish a package with `extraPackages` in it and there yet isn't a new `packageSet.registry` version that I can put in the second package.

Or am I missing something here?

f-f:
You can add the newly published library to the `extraPackages` just fine if you refer to the registry version instead of pointing to the url of the repo, e.g.
```
workspace:
  extraPackages:
    newly-published-library: 0.0.1
```

CGenie:
OK this works, thanks!

Though this leaves the package YAML file in a somewhat intermediate definition where some of the packages are specified in `dependencies` and some in `extraPackages`. And there seems no way to fix that. Suppose the registry catches up with a new version. Then I can edit my library and move `newly-published-library` to dependencies but then I need to release a new version of that library. So I could have waited one day and made fewer releases anyways...

f-f:
I don't understand - you're supposed to have `newly-published-library` in your dependencies all along if your code depends on it, and the registry even forces you to include ranges for it.
The Registry entirely ignores the `extraPackages` key, and it only goes off the list of `dependencies` that your package has.
Spago first, and the Registry later, should fail the publish if your package doesn't include all the packages it needs to compile the source.

CGenie:
Ah, ok, I guess that's fine then, thanks for the explanation.

f-f:
You're welcome. I think we should be documenting this better - would you like to put together a small paragraph about this?

CGenie:
Can this be merged?

PR #1216, opened 2024-04-19
[README] add info about depending on a freshly published library

CGenie:
### Description of the change

Update `README` with info about depending on a freshly published library.

Fix https://github.com/purescript/spago/issues/1215

### Checklist:


README.md
```diff
@@ -563,6 +563,39 @@ workspace:
...
+this form. Spago will only not let you publish packages which use
+unpublished libraries, i.e. things like
+```yaml
+workspace:
+  extraPackages:
+    newly-published-library:
+	  git: http://.....
+	  ref: ....
+```
+
```
f-f, line 598: I wouldn't really say that "you don't need to add the library to the dependencies": if the library is _not_ a direct dependency of the package to publish (but it's brought in transitively by another package), then this statement is correct but misleading, if it's a direct dependency then I can see it being quite confusing as it might drive people to remove the direct dependency thinking that having the package in the `extraPackages` is enough to include it in the build plan.

The point about git packages is good, but I'd format it as a note to make it stand out:
```suggestion
> [!NOTE]\
> This only works when the package you add to `extraPackages` has been published to the registry. Adding a git dependency will produce an error, as publishing to the Registry only admits build plans that only contain packages coming from the Registry.
```

README.md
```diff
@@ -1055,6 +1055,39 @@ $ spago publish
...
```
f-f, line 1090: This point is already made in the note above

```suggestion
```

f-f (approved):
Thanks!

PR #1220, opened 2024-04-25
Remove registry-foreign dependency and use the new Spago.Glob module instead.

Blugatroff:
### Description of the change
Follow up of #1210

I replaced the only remaining use of Registry.Foreign.FastGlob.match with Spago.Glob.gitignoringGlob.

After running tests i noticed, that `compileGlob` is called with paths into the .spago directory, which is ignored by a .gitignore file, which means `gitignoringGlob` never found a match.
Before using a .gitignore file for pruning the search, `gitignoringGlob` now checks that none of the `includePatterns` would be excluded.

`gitignorePatternToMicromatch` also had a bug, `foo`in a gitignore should translate to a micromatch `**/foo/**` to properly match stuff like `foo/bla/blob`.

### Checklist:

### DIFF
--- a/src/Spago/Glob.purs
+++ b/src/Spago/Glob.purs
@@ -4,6 +4,7 @@ import Spago.Prelude
 
 import Data.Array as Array
 import Data.String as String
+import Data.Foldable (any)
 import Effect.Aff as Aff
 import Effect.Ref as Ref
 import Node.FS.Sync as SyncFS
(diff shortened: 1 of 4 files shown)


src/Spago/Glob.purs
```diff
@@ -93,11 +94,19 @@ fsWalk cwd ignorePatterns includePatterns = Aff.makeAff \cb -> do
...
+            -- thing we are asked to look for?
+            -- For example, the .gitignore might have the pattern ".spago" but
+            -- `includePatterns` contains a glob like .spago/p/foo-3.1.4/**/*.purs
+            let anyIncludePatternWouldBeIgnored = any matcherForThisGitignore includePatterns
+
+            -- In such a case, do not use this .gitignore for pruning the search.
+            when (not anyIncludePatternWouldBeIgnored) do
+              -- Instead of composing the matcher functions, we could also keep a growing array of
+              -- patterns and regenerate the matcher on every append. I don't know which option is
+              -- more performant, but composing functions is more conventient.
```
f-f, line 107:
```suggestion
              -- more performant, but composing functions is more convenient.
```

f-f (approved):
Looks great, thanks!

f-f:
@Blugatroff ah, Windows CI is not happy. I guess this is about paths containing `\` instead of the `/` we match on 😄

PR #1222, opened 2024-05-11
fix: .gitignored dirs should be excluded from workspace, pedanticPackages should handle gitignore patterns with leading slash

cakekindel:
### Description of the change
Fixes a bug with the globbing changes introduced in `56a36d5` where a `.gitignore` pattern of `/.spago` causes spago to (incorrectly) ignore `.spago` when building the module dependency graph.

This is felt when building with `--pedantic-packages`, but I imagine it probably causes some other issues, too.

This is caused by some subtle micromatch behavior differences compared with the previous globbing strategy:

```purescript
--== 56a36d5 src/Spago/Glob.purs L93 ==--
try (SyncFS.readTextFile UTF8 gitignorePath) >>= case _ of
  Left _ -> pure unit
  Right gitignore -> do
    let { ignore, patterns } = gitignoreToMicromatchPatterns base gitignore
    -- when .gitignore is `/.spago`, patterns is `[".spago"]`.
    
    let matcherForThisGitignore = micromatch { ignore } patterns
    let anyIncludePatternWouldBeIgnored = any matcherForThisGitignore includePatterns
    -- when patterns is `[".spago"]`
    -- and includePatterns is `[".spago/p/aff-1.0.0/**/*.purs"]`
    -- then this returns `false` (because `minimatch('foo/**/*')('foo')` returns `false`)

    when (not anyIncludePatternWouldBeIgnored) do
      -- We add all .gitignore patterns to `fsWalk`
      let addMatcher currentMatcher = or [ currentMatcher, matcherForThisGitignore ]
      void $ Ref.modify addMatcher ignoreMatcherRef
```

Reproduce with
```sh
> npm i -g spago@0.93.29
> mkdir gitignore-repro
> cd gitignore-repro
> spago init
> spago build --pedantic-packages
# succeeds
> cat .gitignore | awk '{print "/" $0}' > .gitignore.new; mv .gitignore.new .gitignore
> spago build --pedantic-packages
Sources for package 'gitignore-repro' declares unused dependencies - please remove them from the project config:
  - console
  - effect
  - prelude

These errors can be fixed by running the below command(s):
spago uninstall -p gitignore-repro console effect prelude
```

### Checklist:


cakekindel:
[b0bc0de](https://github.com/purescript/spago/pull/1222/commits/b0bc0dedbdf661f7fc5091b2923d99195e9dd486)

If this change is contentious for any reason I can revert it - something that's been bothering me about bleeding spago is that it walks gitignored dirs looking for workspace members, including `node_modules` and errors when there are .gitignored read-protected dirs (ex. a docker container volume)

f-f (commented):
The change looks good - could you add the repro as a test?
It should go in [the Pedantic suite](https://github.com/purescript/spago/blob/master/test/Spago/Build/Pedantic.purs), but should use the [new `copyTree` structure](https://github.com/purescript/spago/blob/82d99ab015e3852e25ab9f8ae36bbce2bf338cdd/test/Spago/Build/Monorepo.purs#L118) from the more recent tests (it's shorter and nicer to work with, we haven't managed to refactor the old tests yet)

f-f:
> something that's been bothering me about bleeding spago is that it walks gitignored dirs looking for workspace members, including node_modules and errors when there are .gitignored read-protected dirs (ex. a docker container volume)

Do you think there's a way to improve on this while keeping good performance?

cakekindel:
Sorry! To clarify, the partial gitignore change both _significantly_ improves performance when large node_modules is present and aligns with the behavior I'd expect of not looking for source packages in gitignored dirs.

If scanning gitignored dirs for packages is desired though I can revert that change :)

f-f (approved):
Thanks for the clarifications! This looks great now 🙂

cakekindel:
@f-f Gentle bump on this, I'm not sure why actions thinks the macos task has been running for 16 days when it has clearly finished :stuck_out_tongue_closed_eyes:

f-f:
Thanks for the ping! I was waiting for CI to be done then forgot to check on it, sorry 🥲

Issue #1086, opened 2023-10-15
Registry package name limitations are also applied to local packages


finnhodgkin:
Legacy spago didn't have length validation on local package names, whereas spago@next has a strict 50-character limit.

Unfortunately in the project I'm migrating we have a bunch of generated libraries with really long names like `oa-gql-enums-author-field-response-versions-select-column`. It wouldn't be the end of the world if we had to cut the names down or merge the generated libraries, but it feels like an arbitrary limit for unpublished local packages.

Seems to be caused by the spago.yaml being read using the parser from registry-dev:

[Spago config parser](https://github.com/purescript/spago/blob/master/core/src/Config.purs#L85C28)

[Registry name parser](https://github.com/purescript/registry-dev/blob/master/lib/src/PackageName.purs#L119)

Other than this the migration was going great, I'm really loving the new workspace features! 🎉 

Minimal example:

```
package:
  dependencies:
    - console
    - prelude
  name: a-really-long-package-name-that-is-53-characters-long
  test:
    dependencies: []
    main: Test.Main
workspace:
  extra_packages: {}
  package_set:
    registry: 43.0.0
```

Results in: 

> Reading Spago workspace configuration...
\
❌ Couldn't parse Spago config, error:
  An error occurred while decoding a JSON value:
  Under 'Config':
  At object key package:
  Under 'PackageConfig':
  At object key name:
  Under 'PackageName':
  Unexpected value "a-really-long-package-name-that-is-53-characters-long".
Run `spago init` to initialise a new project.

finnhodgkin:
Happy to make a PR some time this week if it's an unintentional change. The 50 character limit for published packages feels legit but I could break the parser into local and remote versions.

f-f:
Thanks for reporting! I had no suspicion whatsoever that this would ever be an issue, and yet 😄 

I am a bit hesitant to split the types that Spago uses from the Registry ones, especially the `PackageName`. I think we picked the 50-char limit somewhat arbitrarily, so I guess moving the post would not be a bit deal - what do you think @thomashoneyman?

f-f:
@finnhodgkin what's the longest package name you have?

finnhodgkin:
150 characters 🙈

thomashoneyman:
I think the limit was chosen arbitrarily and could be raised. But it seems to me the more correct thing to do would be to allow local package names to be arbitrarily long.

[...]

PR #1224, opened 2024-05-29
Use registry lib with 150 char limit on package names

roryc89:
addresses https://github.com/purescript/spago/issues/1086

### Description of the change

TODO

### Checklist:


f-f (approved):
Thanks!

Issue #1229, opened 2024-06-13
Fail to install spago@next on macOS Sonoma 14.5


FranklinChen:
```
$ npm i -g spago@next
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm error code 1
npm error path /opt/homebrew/lib/node_modules/spago/node_modules/better-sqlite3
npm error command failed
npm error command sh -c prebuild-install || node-gyp rebuild --release
npm error TOUCH ba23eeee118cd63e16015df367567cb043fed872.intermediate
npm error   ACTION deps_sqlite3_gyp_locate_sqlite3_target_copy_builtin_sqlite3 ba23eeee118cd63e16015df367567cb043fed872.intermediate
npm error   TOUCH Release/obj.target/deps/locate_sqlite3.stamp
npm error   CC(target) Release/obj.target/sqlite3/gen/sqlite3/sqlite3.o
npm error   LIBTOOL-STATIC Release/sqlite3.a
npm error   CXX(target) Release/obj.target/better_sqlite3/src/better_sqlite3.o
npm error rm ba23eeee118cd63e16015df367567cb043fed872.intermediate
npm error prebuild-install warn install No prebuilt binaries found (target=22.3.0 runtime=node arch=arm64 libc= platform=darwin)
npm error gyp info it worked if it ends with ok
npm error gyp info using node-gyp@10.1.0
npm error gyp info using node@22.3.0 | darwin | arm64
npm error gyp info find Python using Python version 3.12.3 found at "/opt/homebrew/opt/python@3.12/bin/python3.12"
npm error gyp info spawn /opt/homebrew/opt/python@3.12/bin/python3.12
npm error gyp info spawn args [
npm error gyp info spawn args '/opt/homebrew/lib/node_modules/npm/node_modules/node-gyp/gyp/gyp_main.py',
npm error gyp info spawn args 'binding.gyp',
npm error gyp info spawn args '-f',
npm error gyp info spawn args 'make',
npm error gyp info spawn args '-I',
npm error gyp info spawn args '/opt/homebrew/lib/node_modules/spago/node_modules/better-sqlite3/build/config.gypi',
npm error gyp info spawn args '-I',
[...]

PR #1230, opened 2024-06-13
Update dependencies in package.json.

FranklinChen:
Address #1229 by simply updating all dependencies to the latest, and changing just one source file that used an old version of `glob`.


f-f (approved):
Excellent, thank you! 👏

PR #1234, opened 2024-06-17
fix: gitignoringGlob stack safety + perf

### LINKED ISSUE #1231: gitignoringGlob exceeding call stack - add explicit package globs to workspace?
On a very large monorepo project we find that `Glob.gitignoringGlob` exceeds the maximum callstack. 

We can work around this by increasing the maximum stack size on node but we would rather not have to do this.

Another possible solution would be to allow explicit setting of globs to find packages and to not examine `.gitignore` files.

This would also allow generated, gitignored packages to be part of a workspace.

cakekindel:
### Description of the change

fixes #1231 

Before, we were eagerly turning string glob patterns to matching functions `String -> Boolean`, and adding new patterns with `||`. The instance `(HeytingAlgebra b) => HeytingAlgebra (a -> b)` isn't stack safe, causing sufficiently large gitignore files to overflow the call stack.

Now, we accumulate the patterns themselves and build them into micromatch patterns when we evaluate them. The actual call to `micromatch` costs practically nothing, so this flat-out wins time & space cost (as well as fixing the stack-safety issue)

```purs
-- Instead of composing the matcher functions, we could also keep a growing array of
-- patterns and regenerate the matcher on every append. I don't know which option is
-- more performant, but composing functions is more convenient.
```
Turns out we've learned which option is more performant :sweat_smile:

### Checklist:

### DIFF
--- a/src/Spago/Glob.purs
+++ b/src/Spago/Glob.purs
@@ -2,9 +2,12 @@ module Spago.Glob (gitignoringGlob) where
 
 import Spago.Prelude
 
+import Control.Alternative (guard)
+import Control.Monad.Maybe.Trans (MaybeT(..), runMaybeT)
+import Control.Monad.Trans.Class (lift)
 import Data.Array as Array
 import Data.Filterable (filter)
-import Data.Foldable (any)
+import Data.Foldable (any, fold)
 import Data.String as String
 import Data.String.CodePoints as String.CodePoint
 import Effect.Aff as Aff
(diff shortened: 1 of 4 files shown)


test/Spago/Glob.purs
```diff
@@ -66,3 +73,9 @@ spec = Spec.around globTmpDir do
         FS.writeTextFile (Path.concat [p, ".gitignore"]) """/fruits\n/src"""
         a <- Glob.gitignoringGlob p ["fruits/apple/**"]
         a `Assert.shouldEqual` ["fruits/apple"]
+
+      Spec.it "is stacksafe" \p -> do
```
cakekindel, line 77: fails on `f0f7901` (tip of `master` at the moment)

src/Spago/Glob.purs
```diff
@@ -2,21 +2,28 @@ module Spago.Glob (gitignoringGlob) where
...
+import Data.Traversable (traverse_)
 import Effect.Aff as Aff
 import Effect.Ref as Ref
 import Node.FS.Sync as SyncFS
 import Node.Path as Path
 import Record as Record
 import Type.Proxy (Proxy(..))
 
-type MicroMatchOptions = { ignore :: Array String }
+type MicroMatchOptions = { ignore :: Array String, include :: Array String }
```
cakekindel, line 21: this change allows us to accumulate with `append`ing arrays instead of `||`ing functions.

Fun fact for those who may not be aware, if all of a record's fields are `Semigroup`, the record is `Semigroup` so you can `<>` these :)

cakekindel:
is micromatch not cross-platform? :anger:

f-f:
I am very confused as well - I went through the code and all the paths seem to be handled properly, so I am not sure why the failures

cakekindel:
I'd imagine it's because there's an assumption that `foo/**` will match both `foo and foo/a`, which i'm guessing doesn't hold on windows :confused:

f-f (approved):
Looks great, thanks @cakekindel! 👏

PR #1235, opened 2024-06-24
Use type applications for Proxy

f-f:


f-f:
@cakekindel I merged the previous PR as it looked like it passed tests, but it doesn't - there was a stray `Spec.focus` that was preventing previous tests from running - we still have the issue with windows

cakekindel:
Oh my goodness I'm so sorry

f-f:
No worries, it's completely okay! 🙂 Do you have an idea on how to patch it up? If not I'll try to find a block of time sometime in the next days to look at it

PR #1238, opened 2024-06-26
Clone from local cache when possible

### LINKED ISSUE #1208: `spago install` `git clone`s for same monorepo package multiple times
having same package, with only `subdir` is different

```yaml
    deku:
      git: https://github.com/mikesol/purescript-deku.git
      subdir: deku
      ref: main
    deku-core:
      git: https://github.com/mikesol/purescript-deku.git
      subdir: deku-core
      ref: main
    deku-css:
      git: https://github.com/mikesol/purescript-deku.git
      subdir: deku-css
      ref: main
    deku-dom:
      git: https://github.com/mikesol/purescript-deku.git
      subdir: deku-dom
      ref: main
```

it will 

```
 ~/projects/deku-documentation $ spago build --pedantic-packages

✅ Selecting package to build: deku-documentation

Cloning https://github.com/mikesol/purescript-deku.git
Cloning https://github.com/mikesol/purescript-deku.git
Cloning https://github.com/mikesol/purescript-deku.git
Cloning https://github.com/mikesol/purescript-deku.git
Downloading dependencies...
```

multiple times. Very long

Expected: `uniq` before cloning?

------

Full spago.yaml

```
package:
  name: deku-documentation
  publish:
    version: 0.9.24
    license: "Apache-2.0"
  dependencies:
    - free
    - deku
    - deku-core
    - deku-dom
    - yoga-fetch
    - yoga-json
    - affjax-web
    - affjax
    - profunctor-lenses
    - string-parsers
    - tidy-codegen
    - aff
    - routing-duplex
    - random
    - argonaut-core
    - fetch
    - routing
    - filterable
    - qualified-do
    - canvas
    - uint
    - arraybuffer
    - avar
    - web-pointerevents
    - web-touchevents
    - stringutils
workspace:
  backend:
    cmd: purs-backend-es
    args:
      - build
  package_set:
    registry: 50.4.0
  # yq '.workspace.extra_packages | keys[]' spago.yaml | grep -v "deku" | xargs -I {} rm -rdf ".spago/p/{}"
  extra_packages:
    hyrule:
      # git: https://github.com/mikesol/purescript-hyrule.git
      # ref: master
      git: https://github.com/srghma/purescript-hyrule.git
      ref: hyrule-master
    bolson:
      # git: https://github.com/mikesol/purescript-bolson.git
      git: https://github.com/srghma/purescript-bolson.git
      ref: main
    deku:
      git: https://github.com/mikesol/purescript-deku.git
      subdir: deku
      ref: main
    deku-core:
      git: https://github.com/mikesol/purescript-deku.git
      subdir: deku-core
      ref: main
    deku-css:
      git: https://github.com/mikesol/purescript-deku.git
      subdir: deku-css
      ref: main
    deku-dom:
      git: https://github.com/mikesol/purescript-deku.git
      subdir: deku-dom
      ref: main
    dodo-printer:
      # git: https://github.com/natefaubion/purescript-dodo-printer.git
      git: https://github.com/srghma/purescript-dodo-printer.git
      ref: master
    tidy:
      # git: https://github.com/natefaubion/purescript-tidy.git
      git: https://github.com/srghma/purescript-tidy.git
      ref: main
      subdir: lib
    tidy-codegen:
      # git: https://github.com/natefaubion/purescript-tidy-codegen.git
      git: https://github.com/srghma/purescript-tidy-codegen.git
      ref: main
```

ysangkok:
### Description of the change

Fixes #1208.

### Checklist:

### DIFF
--- a/src/Spago/Command/Fetch.purs
+++ b/src/Spago/Command/Fetch.purs
@@ -185,11 +185,16 @@ run { packages: packagesRequestedToInstall, ensureRanges, isTest, isRepl } = do
     -- then for every package we have we try to download it, and copy it in the local cache
     logInfo "Downloading dependencies..."
 
-    parallelise $ (flip map) (Map.toUnfoldable depsToFetch :: Array (Tuple PackageName Package)) \(Tuple name package) -> do
+    gitReposCloned :: Ref (Map String String) <- liftEffect $ Ref.new mempty
+    for_ (Map.toUnfoldable depsToFetch :: Array (Tuple PackageName Package)) \(Tuple name package) -> do
       let localPackageLocation = Config.getPackageLocation name package
       -- first of all, we check if we have the package in the local cache. If so, we don't even do the work
       unlessM (FS.exists localPackageLocation) case package of
-        GitPackage gitPackage -> getGitPackageInLocalCache name gitPackage
+        GitPackage gitPackage -> do
+          currentMap <- liftEffect $ Ref.read gitReposCloned
+          res <- getGitPackageInLocalCache name gitPackage currentMap
+          liftEffect $ Ref.modify_ (Map.insert gitPackage.ref localPackageLocation) gitReposCloned
+          pure res
         RegistryVersion v -> do
           -- if the version comes from the registry then we have a longer list of things to do
           let versionString = Registry.Version.print v
@@ -286,7 +291,7 @@ writeNewLockfile reason allTransitiveDeps = do
     processPackage :: LockfileBuilderResult -> Tuple PackageName (Tuple PackageName Package) -> Spago (FetchEnv a) LockfileBuilderResult
     processPackage result (Tuple workspacePackageName (Tuple dependencyName dependencyPackage)) = do
       (packageDependencies :: Array PackageName) <- (Array.fromFoldable <<< Map.keys <<< fromMaybe Map.empty)
-        <$> getPackageDependencies dependencyName dependencyPackage
+        <$> getPackageDependencies dependencyName dependencyPackage Map.empty
       let
         updatePackage r package = (updateWorkspacePackage r)
           { packages = Map.insert dependencyName package r.packages }
@@ -357,12 +362,13 @@ writeNewLockfile reason allTransitiveDeps = do
 toAllDependencies :: PackageTransitiveDeps -> PackageMap
 toAllDependencies = foldl (Map.unionWith (\l _ -> l)) Map.empty
 
-getGitPackageInLocalCache :: forall a. PackageName -> GitPackage -> Spago (Git.GitEnv a) Unit
-getGitPackageInLocalCache name package = do
+getGitPackageInLocalCache :: forall a. PackageName -> GitPackage -> Map String String -> Spago (Git.GitEnv a) Unit
+getGitPackageInLocalCache name package gitReposCloned = do
   let localPackageLocation = Config.getPackageLocation name (GitPackage package)
+      cloneFrom = Map.lookup package.git gitReposCloned
   tempDir <- mkTemp' (Just $ printJson Config.gitPackageCodec package)
   logDebug $ "Cloning repo in " <> tempDir
-  Git.fetchRepo package tempDir >>= case _ of
+  Git.fetchRepo package cloneFrom tempDir >>= case _ of
     Left err -> die err
     Right _ -> do
       logDebug $ "Repo cloned. Moving to " <> localPackageLocation
@@ -383,8 +389,8 @@ getGitPackageInLocalCache name package = do
             FS.mkdirp $ Path.concat [ Paths.localCachePackagesPath, PackageName.print name ]
             FS.copyTree { src: localPackageLocation, dst: commitHashLocation }
 
-getPackageDependencies :: forall a. PackageName -> Package -> Spago (FetchEnv a) (Maybe (Map PackageName Range))
-getPackageDependencies packageName package = case package of
+getPackageDependencies :: forall a. PackageName -> Package -> Map String String -> Spago (FetchEnv a) (Maybe (Map PackageName Range))
+getPackageDependencies packageName package gitReposCloned = case package of
   RegistryVersion v -> do
     maybeManifest <- Registry.getManifestFromIndex packageName v
     pure $ map (_.dependencies <<< unwrap) maybeManifest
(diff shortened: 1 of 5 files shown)


src/Spago/Command/Fetch.purs
```diff
@@ -286,7 +291,7 @@ writeNewLockfile reason allTransitiveDeps = do
     processPackage :: LockfileBuilderResult -> Tuple PackageName (Tuple PackageName Package) -> Spago (FetchEnv a) LockfileBuilderResult
     processPackage result (Tuple workspacePackageName (Tuple dependencyName dependencyPackage)) = do
       (packageDependencies :: Array PackageName) <- (Array.fromFoldable <<< Map.keys <<< fromMaybe Map.empty)
-        <$> getPackageDependencies dependencyName dependencyPackage
+        <$> getPackageDependencies dependencyName dependencyPackage Map.empty
```
ysangkok, line 294: It seems feasible to populate the map here, but OTOH it doesn't seem like it would save much time since the lock file is generated rarely.

ysangkok:
I pushed a commit to improve efficiency when an existing repo is updated. While developing this commit, I had some concerns:

* In the case that you are switching to an unrelated branch, and you have local     
  unstaged changes, the checkout will fail but the rebase will still run. Spago     
  will autostash, which IMO is a bit confusing, cause you might want to start       
  french if you switch to an unrelated branch. But I suppose this isn't too bad                                                                               
  since support for local unstaged changes is a really experimental feature?        
                                                                                    
* In the case that you are switching to a differnt branch, and you have local       
  commits, they will be left behind because the checkout will just succeed. I       
  suppose this is ok too, since the user shouldn't rely on their custom commits     
  surviving a branch change. But it seems inconsistent with the autostashing.       
                                                                                    
* In the case where the user is using a short-hash which becomes ambiguous, I       
  am relying on rebase to fail with a message like                                  
                                                                                    
      % git rebase b756                                                               
      error: short object ID b756 is ambiguous                                        
      hint: The candidates are:                                                       
      hint:   b7565f941 commit 2012-09-29 - Extend the unpack command for the .cabal file updating    
      hint:   b75671b0b commit 2024-03-11 - Merge pull request #9792 from haskell/mergify/bp/3.12/pr-9768    
      fatal: invalid upstream 'b756'                                                  
                                                                                    
* In the case where the user is 'reverting' an extra-dep using a short hash, it     
  will not actually switch to the new commit if there are unstaged changes. The     
  result is that the new commit is passed to rebase, which will put the 'new'       
  (undesired) commits on the old (desired) commit. This seems to be existing        
  behaviour, so I am not changing it. And it shouldn't affect many since it         
  relies on there being unstaged changes.

f-f:
My understanding of this patch is that we'd only use the cache if we're cloning the same repo multiple times during the current execution.
I think this is subpar: the alternative I propose would be to put the repo in the global cache, and copy it over if we find that we are cloning the same ref.
This would be strictly better for tags and commits, but would require some care for branches, because we wouldn't be able to cache those.
We wouldn't need the checkout logic either I think?

The flow would be:
- if we need to clone a repo, look in the global cache, see if we have the same ref cached
- if it's not cached, then clone it in a temp dir
  - if the ref is a branch (which we [can figure out with a git command](https://stackoverflow.com/questions/18222634/given-a-git-refname-can-i-detect-whether-its-a-hash-tag-or-branch)), then we can't cache, and we'd move it to the `.spago` dir
  - if the ref is a tag or a commit, then we'd move it to the global cache, and then copy it over to the `.spago` dir
 - if it's cached, we just copy it over to the `.spago` dir

ysangkok:
I think the possibility of specifying a branch is making everything much more complicated. Especially because of the `autostash` and `rebase` flags that the codebase currently contains, which makes it look like rogue local commits and changes should survive. Is that a feature we really have to retain?

I think it would be a lot simpler if we started out only handling the case where a full 40 character SHA ref is specified. What would you think if I developed a patch for only that strategy, using your proposal with the global cache? Then it should be easier to test, since we no longer have to handle the case of the upstream branch moving. And we wouldn't handle autostash and rebasing either.

I am also a bit confused about all the places where `getPackageDependencies` is used. If we went with this reduced version of the feature, would it be compatible with the current interface of `getPackageDependencies`? That would make things a lot easier.

Also, would you propose having the global cache filled in a separate IO action, and then later functions would rely on that cache being filled? I tried such an approach in [this commit](https://github.com/ysangkok/spago/commit/89635b31c73e3ec299d954e87d356e13d7c40903), but it was buggy and the call graph was confusing to me, I couldn't figure out the invariants of where the cache was populated, and where it wasn't. And I don't really know how we'd make an API that somehow aligns the global cache with the type system, which would be nice.

ysangkok:
If we only supported refs as 40-character shas, effectively all the code might be contained to `fetchRepo`. That would then look in the global cache and always copy to the destination directory when passed a sha ref, which exists in the cache.

f-f:
> I think the possibility of specifying a branch is making everything much more complicated. Especially because of the autostash and rebase flags that the codebase currently contains, which makes it look like rogue local commits and changes should survive. Is that a feature we really have to retain?

I don't understand how the autostash and rebase flags are causing issues here - Spago's job ends when the source ends up in the right place of the `.spago` folder (i.e. the local cache).
Whatever happens after that (i.e. people editing the repos once they are sitting in the local cache) is the user's business. To my understanding that feature/flow is used quite a lot.
We should be cloning the repo exactly once though, so again, I am missing a concrete scenarion in which these flags are problematic

> I think it would be a lot simpler if we started out only handling the case where a full 40 character SHA ref is specified. What would you think if I developed a patch for only that strategy, using your proposal with the global cache?

People will want to use tags - it's good UX to be able to see which tag one is using. Why is supporting branches and tags an issue? We can keep tags in the global cache, and we can figure out that something is a branch (and not cache it) once we clone it.

> Also, would you propose having the global cache filled in a separate IO action, and then later functions would rely on that cache being filled?

What do you mean by that? It's easy to check if something is in the global cache, a `FS.exists` will do. Or are you referring to the [thundering herd problem](https://en.wikipedia.org/wiki/Thundering_herd_problem)?
We use an [`AVar`](https://pursuit.purescript.org/packages/purescript-avar/5.0.0/docs/Effect.Aff.AVar) for that when fetching the Registry repos:
https://github.com/purescript/spago/blob/66b5f8deb70efcdc12bd435e78a688ed9010ec3d/src/Spago/Registry.purs#L109-L115

> I am also a bit confused about all the places where `getPackageDependencies` is used.

Why is that an issue? This patch should pretty much be self-contained inside `getGitPackageInLocalCache`, which `getPackageDependencies` calls. No details should leak outside of `getGitPackageInLocalCache`: in there we'd check if the package is in global cache, if it's not then we clone it and eventually copy it over there! (as I detailed in my previous comment)

ysangkok:
Thanks for your comments, I tried to adapt to them in #1240. So I am closing this.

PR #1240, opened 2024-06-30
Introduce global git cache

ysangkok:
### Description of the change

Another attempt at #1238. This introduces a global git cache which is used for
non-branches. That means that if you have a tag/commit specified on multiple
dependencies, it will only be cloned once.

### Checklist:


src/Spago/Config.purs
```diff
@@ -464,6 +465,14 @@ shouldComputeNewLockfile { workspace, workspacePackages } workspaceLock =
           _ -> false
       )
 
+getGlobalGitCacheForRepo :: Core.GitPackage -> FilePath
```
f-f, line 468: This function should be moved in the `where` block of `getGitPackageInLocalCache`, since it's only used there

src/Spago/Git.purs
```diff
@@ -40,14 +41,19 @@ runGit args cwd = ExceptT do
     Right r -> Right r.stdout
     Left r -> Left r.stderr
 
-fetchRepo :: forall a b. { git :: String, ref :: String | a } -> FilePath -> Spago (GitEnv b) (Either (Array String) Unit)
+determineShouldCache :: forall b. String -> FilePath -> Spago (GitEnv b) Boolean
+determineShouldCache ref path = do
```
f-f, line 45: It's fine to name things after their purpose, but I'd lean towards being more general here, for the sake of the casual reader (or me in a few months) which might struggle to see how a random git command is tied to our cache. I'd also link the StackOverflow answer for more context:

```suggestion
-- See https://stackoverflow.com/questions/18222634
isBranch :: forall b. { path :: FilePath, ref :: String } -> Spago (GitEnv b) Boolean
isBranch { ref, path } = do
```

src/Spago/Git.purs
```diff
@@ -40,14 +41,19 @@ runGit args cwd = ExceptT do
...
+  showRef <- Except.runExceptT $ runGit_ [ "show-ref", "--verify", "refs/heads/" <> ref ] (Just path)
+  pure $ isLeft showRef
```
f-f, line 47: Now that the function is called `isBranch`, we need to return true if the command succeeds:
```suggestion
  pure $ isRight showRef
```

src/Spago/Git.purs
```diff
@@ -40,14 +41,19 @@ runGit args cwd = ExceptT do
...
+
+fetchRepo :: forall a b. { git :: String, ref :: String | a } -> FilePath -> Spago (GitEnv b) (Either (Array String) Boolean)
 fetchRepo { git, ref } path = do
   repoExists <- FS.exists path
   { offline } <- ask
   case offline, repoExists of
     Offline, true -> do
       logDebug $ "Found " <> git <> " locally, skipping fetch because we are offline"
-      pure $ Right unit
+      Right <$> determineShouldCache ref path
```
f-f, line 56: `fetchRepo` is used throughout the whole project and not only when fetching dependencies. We shouldn't run this check here if possible

src/Spago/Config.purs
```diff
@@ -464,6 +465,14 @@ shouldComputeNewLockfile { workspace, workspacePackages } workspaceLock =
...
+getGlobalGitCacheForRepo package =
+  Path.concat
+  [ Paths.globalCachePath
+  , "git"
+  , fileSystemCharEscape package.git <> fileSystemCharEscape package.ref
```
f-f, line 473: No reason to concat - we can nest them more:
```suggestion
  , fileSystemCharEscape package.git
  , fileSystemCharEscape package.ref
```

src/Spago/Command/Fetch.purs
```diff
@@ -357,31 +357,67 @@ writeNewLockfile reason allTransitiveDeps = do
...
-  Git.fetchRepo package tempDir >>= case _ of
-    Left err -> die err
-    Right _ -> do
+  let gitCacheLocation = Config.getGlobalGitCacheForRepo package
+      localPackageLocation = Config.getPackageLocation name (GitPackage package)
+  existsGitCache <- FS.exists gitCacheLocation
+  Tuple copyOrMove sourceDir <-
+    if existsGitCache
+      then do
+        shouldCache <- Git.determineShouldCache package.ref gitCacheLocation
```
f-f, line 370: If the package exists in the global cache then we don't even have to ask the question - it's cacheable because it was put there previously

src/Spago/Command/Fetch.purs
```diff
@@ -357,31 +357,67 @@ writeNewLockfile reason allTransitiveDeps = do
...
-  let localPackageLocation = Config.getPackageLocation name (GitPackage package)
-  tempDir <- mkTemp' (Just $ printJson Config.gitPackageCodec package)
-  logDebug $ "Cloning repo in " <> tempDir
-  Git.fetchRepo package tempDir >>= case _ of
-    Left err -> die err
-    Right _ -> do
+  let gitCacheLocation = Config.getGlobalGitCacheForRepo package
+      localPackageLocation = Config.getPackageLocation name (GitPackage package)
+  existsGitCache <- FS.exists gitCacheLocation
+  Tuple copyOrMove sourceDir <-
```
f-f, line 367: Don't think we need this `CopyOrMove` type - if the package can be cached globally then we need to copy it, if it can't then we need to move it.
Moreover, it's generally more idiomatic to use a Record rather than an N-tuple when returning multiple values:
```suggestion
  { globallyCached, sourceDir } <-
```

(where `globallyCached` is a boolean)

src/Spago/Command/Fetch.purs
```diff
@@ -357,31 +357,67 @@ writeNewLockfile reason allTransitiveDeps = do
...
+        if shouldCache
+          then
+            logInfo $ "Using global git cache in " <> gitCacheLocation
+          else do
+            fetchRes <- Git.fetchRepo package gitCacheLocation
+            case fetchRes of
+              Left err -> die err
+              Right _ -> pure unit
+        pure (Tuple Copy gitCacheLocation)
+      else do
```
f-f, line 380: Informative comment to keep track of what's happening
```suggestion
      -- Not cached in the global cache, so we download it and then see if it can be cached
      else do
```

src/Spago/Command/Fetch.purs
```diff
@@ -357,31 +357,67 @@ writeNewLockfile reason allTransitiveDeps = do
...
+        tempDir <- mkTemp' (Just $ printJson Config.gitPackageCodec package)
+        logDebug $ "Cloning repo in " <> tempDir
+        fetchRes <- Git.fetchRepo package tempDir
+        case fetchRes of
+          Left err -> die err
+          Right shouldCache -> do
```
f-f, line 386: Following my suggestion in the `Git` module to take the caching check our of the `fetchRepo`:
```suggestion
          Right _ -> do
            shouldCache <- not <$> Git.isBranch { ref: package.ref, path: tempDir }
```

src/Spago/Command/Fetch.purs
```diff
@@ -357,31 +357,67 @@ writeNewLockfile reason allTransitiveDeps = do
...
+            if shouldCache
+              then do
+                -- Concurrent threads may have created the directory now,
+                -- especially since fetching can take a long time
+                existsGitCache2 <- FS.exists gitCacheLocation
+                when (not existsGitCache2) do
+                  FS.moveSync { src: tempDir, dst: gitCacheLocation }
+                pure (Tuple Copy gitCacheLocation)
```
f-f, line 394: And following the suggestion to return a record:
```suggestion
                pure { globallyCached: true, sourceDir: gitCacheLocation }
```

f-f (commented):
Ah yes, this patch looks much better! The overall structure is good and I left a few minor comments about code organisation and naming

CI fails one test, but I'm hoping it's a glitch 😄

f-f:
Ah right, CI fails because this change is correct 🙂
In the `can't install (uncached) dependencies if **offline**` test we were verifying that git repos never get cached, but with this change we hit the cache when trying to clone the same repo twice. We'll have to clone a different version of `either` in there, and/or add a test where we clone a branch.

ysangkok:
@f-f I applied your suggested changes, and also an `mkdirp` for the deeper directory hierarchy. The reason I joined `package.git` with `package.ref` is because there is no synchronization outside the filesystem, so if there is a subdirectory, I am worried that the more work happen between the call to `exists` and the `mkdirp` and the `moveSync`, the more likely threads are to interleave any of this, which wouldn't work. Ideally, the move should be called unconditionally, and it fails because of an existing destination, the failure could be ignored. That would be correct on many Linux filesystems since moves are atomic. Now that we have three file system operations, it is not atomic. Not sure what to do about this.

src/Spago/Command/Fetch.purs
```diff
@@ -360,28 +360,75 @@ toAllDependencies = foldl (Map.unionWith (\l _ -> l)) Map.empty
...
+  { globallyCached, sourceDir } <-
+    if existsGitCache
+      then do
+        logInfo $ "Using global git cache in " <> gitCacheLocation
+        isBranch <- Git.isBranch { ref: package.ref, path: gitCacheLocation }
+        when isBranch do
+          fetchRes <- Git.fetchRepo package gitCacheLocation
+          case fetchRes of
+            Left err -> die err
+            Right _ -> pure unit
```
f-f, line 373: If the package is in cache we just copy it over, no need to check if it's a branch since it can't be:
```suggestion
        logDebug $ "Using global git cache in " <> gitCacheLocation
```

src/Spago/Command/Fetch.purs
```diff
@@ -360,28 +360,75 @@ toAllDependencies = foldl (Map.unionWith (\l _ -> l)) Map.empty
...
+              then do
+                -- Concurrent threads may have created the directory now,
+                -- especially since fetching can take a long time
+                existsGitCache2 <- FS.exists gitCacheLocation
+                when (not existsGitCache2) do
+                  FS.mkdirp globalGitCache
+                  FS.moveSync { src: tempDir, dst: gitCacheLocation }
+                pure { globallyCached: true, sourceDir: gitCacheLocation }
+              else do
+                pure { globallyCached: false, sourceDir: gitCacheLocation }
```
f-f, line 394: Instead of checking again we can just catch errors - if the move operation fails then other fiber likely put the repo there (we don't have true concurrency though, so I'm not sure how likely that is):
```suggestion
            when shouldCache do
              FS.mkdirp globalGitCache
              try (FS.moveSync { src: tempDir, dst: gitCacheLocation }) >>= case _ of
                Left _err -> logDebug $ "Couldn't move cloned repo to " <> gitCacheLocation
                Right _ -> pure unit
            pure { globallyCached: shouldCache, sourceDir: gitCacheLocation }
```

f-f (commented):
> Now that we have three file system operations, it is not atomic. Not sure what to do about this.

We can catch filesystem errors, see the suggestions

ysangkok:
I don't know how to change the version of `either`, so I tried to put `identity` instead, but it doesn't seem to have worked. I wonder if `identity` is somehow already in the cache. Not sure what other package to use.

test/Spago/Install.purs
```diff
@@ -104,7 +104,7 @@ spec = Spec.around withTempDir do
     Spec.it "can't install (uncached) dependencies if offline" \{ spago, fixture } -> do
       spago [ "init" ] >>= shouldBeSuccess
       writeConfigWithEither
-      spago [ "install", "--offline", "either" ] >>= shouldBeFailureErr (fixture "offline.txt")
+      spago [ "install", "--offline", "identity" ] >>= shouldBeFailureErr (fixture "offline.txt")
```
f-f, line 107: The trick to make this work is in the line above - `writeConfigWithEither` will write a `spago.yaml` that points `either` to a specific commit that we are certain is not in cache.
We could modify `writeConfigWithEither` (defined in line 237) to also point `identity` to its git repo at a specific commit, then this would work.

Once we do that we should duplicate this test and call it `"can install cached dependencies if offline"`, where we try to install `either`, which the previous test brought into the cache.

ysangkok: Addressed in [5ac3611](https://github.com/purescript/spago/pull/1240/commits/5ac36114da694d230575fb86d993a9f7d21c1a2f)

ysangkok:
@f-f Let me know if there are any remaining tasks for this, please! I have fixed the formatting.

f-f:
CI fails because on Windows (and only there) the folder in the cache is not found when we go copy it:
```
Error: ENOENT: no such file or directory, lstat 'C:\\Users\\runneradmin\\AppData\\Local\\spago-nodejs\\Cache\\git\\https%3a%2f%2fgithub.com%2fpurescript%2fpurescript-either.git\\af655a04ed2fd694b6688af39ee20d7907ad0763'
    at Object.lstatSync (node:fs:1666:3)
    at Object.lstatSync (D:\\a\\spago\\spago\\node_modules\\graceful-fs\\polyfills.js:318:34)
    at statFunc (D:\\a\\spago\\spago\\node_modules\\fs-extra\\lib\\util\\stat.js:24:20)
    at getStatsSync (D:\\a\\spago\\spago\\node_modules\\fs-extra\\lib\\util\\stat.js:25:19)
    at Object.checkPathsSync (D:\\a\\spago\\spago\\node_modules\\fs-extra\\lib\\util\\stat.js:64:33)
    at Object.copySync (D:\\a\\spago\\spago\\node_modules\\fs-extra\\lib\\copy\\copy-sync.js:27:38)
    at file:///D:/a/spago/spago/output/Spago.FS/foreign.js:9:54
    at runSync (file:///D:/a/spago/spago/output/Effect.Aff/foreign.js:88:20)
    at run (file:///D:/a/spago/spago/output/Effect.Aff/foreign.js:326:22)
    at file:///D:/a/spago/spago/output/Effect.Aff/foreign.js:346:19 {
  errno: -4058,
  syscall: 'lstat',
  code: 'ENOENT',
  path: 'C:\\\\Users\\\\runneradmin\\\\AppData\\\\Local\\\\spago-nodejs\\\\Cache\\\\git\\\\https%3a%2f%2fgithub.com%2fpurescript%2fpurescript-either.git\\\\af655a04ed2fd694b6688af39ee20d7907ad0763'
}
```

I had weird encounters with windows not writing files to disk in a timely manner, but it would be strange if it would do it for entire folders.
I wonder if we're writing to a place and trying to read from another, or if some of the characters we use in the escaping are not good

ysangkok:
I can't seem to reproduce this failure locally. So not sure how to proceed. I could insert some delays between tests, but that wouldn't be an acceptable permanent solution.

image

f-f:
Ah that's strange - I wonder if this is about Node versions? In CI we're using a pretty old one, 18:
https://github.com/purescript/spago/blob/52999702842304b360b24f57571523564d261823/.github/workflows/build.yml#L41

I wonder if bumping it up to 22 will fix it?
If not I'll try to take a look on my Windows machine sometime in the next days

ysangkok:
Is there a way to get the test runner to just run a single test? That would help with diagnosing whether it's the state carried over between tests that is messing up the failing test.

f-f:
Yes! You can swap the `Spec.it` of that test for `Spec.itOnly`

f-f:
I can't reproduce locally either unfortunately. I'd really like to merge this ASAP - but adding debug prints to CI seems to be the only way to debug this, so it will be slow

f-f:
Superseded by #1275

PR #1244, opened 2024-07-05
Fix glob performance on large monorepos

### LINKED ISSUE #1242: Stuck on "Gathering all the spago configs in the tree"
The build hangs (or it's taking more than 1 minute) when running `Glob.gitignoringGlob`, so the latest patches there must be spinning in some corner case - this issue doesn't show up e.g. when building spago itself.

cc @cakekindel 

## Repro

- clone https://github.com/purescm/purescript-core
- run `spago build`
- wait forever

f-f:
Fix #1242, using [purescript-core](https://github.com/purescm/purescript-core) as the benchmark.

The final patch goes back to the approach we had before #1234, where we'd compose the glob-matching functions instead of keeping a list of ignores, and recompute the matchers when needed.

The patch in #1234 was not optimised, as the matchers were being recreated for every file encountered, and optimising that in the first two commits of this PR got us down to 2x of the performance pre-1234.
That is unfortunately still not acceptable, so I reintroduced the function-composition approach, which is still prone to blowing the stack - the change here should reduce that risk, since instead of composing every line of gitignores as a separate matcher in the chain, we instead nest them in a single `or` block. That should dramatically reduce the size of the call chain.

cc @Blugatroff @cakekindel

### DIFF
--- a/src/Spago/Glob.purs
+++ b/src/Spago/Glob.purs
@@ -11,11 +11,10 @@ import Data.Foldable (any, fold)
 import Data.String as String
 import Data.Traversable (traverse_)
 import Effect.Aff as Aff
+import Effect.Class.Console as Console
 import Effect.Ref as Ref
 import Node.FS.Sync as SyncFS
 import Node.Path as Path
-import Record as Record
-import Type.Proxy (Proxy(..))
 
 type Glob =
   { ignore :: Array String
@@ -30,8 +29,15 @@ splitGlob { ignore, include } = (\a -> { ignore, include: [ a ] }) <$> include
 type Entry = { name :: String, path :: String, dirent :: DirEnt }
 type FsWalkOptions = { entryFilter :: Entry -> Effect Boolean, deepFilter :: Entry -> Effect Boolean }
 
+-- https://nodejs.org/api/fs.html#class-fsdirent
 foreign import data DirEnt :: Type
 foreign import isFile :: DirEnt -> Boolean
+
+foreign import direntToString :: DirEnt -> String
+
+instance Show DirEnt where
+  show = direntToString
+
 foreign import fsWalkImpl
   :: (forall a b. a -> Either a b)
   -> (forall a b. b -> Either a b)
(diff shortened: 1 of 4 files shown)


f-f:
The current changes bring the runtime down from 3min 41s to 53s

f-f:
And back to 6s.

cakekindel:
I wonder if it would be possible to stack-safely and performantly accumulate into a regex pattern, digging a bit into micromatch's source and the core implementation picomatch I realize we could avoid reparsing the patterns (which is where I assume the perf cost is)

f-f:
@cakekindel that would be cool!
I'll now merge&release this so that whoever is affected by the issue is unblocked, let's add improvements in a followup PR

src/Spago/Glob.purs
```diff
@@ -74,41 +74,63 @@ fsWalk cwd ignorePatterns includePatterns = Aff.makeAff \cb -> do
 
   -- Pattern for directories which can be outright ignored.
   -- This will be updated whenver a .gitignore is found.
-  ignoreMatcherRef :: Ref Glob <- Ref.new { ignore: [], include: ignorePatterns }
+  ignoreMatcherRef :: Ref (String -> Boolean) <- Ref.new (testGlob { ignore: [], include: ignorePatterns })
```
cakekindel, line 77: Not sure why this never occurred to me before but if we always boolean OR `||` the matcher functions could we not accumulate into an `Array (String -> Boolean)` and flatten at the end with `||`, avoiding the stack unsafety?

cakekindel: Not necessarily, because `or (_ :: Array (_ -> Boolean))` still uses `HeytingAlgebra (a -> Boolean)`

cakekindel: trying this out on `purescm/purescript-core` and seeing a negligible difference. I'll keep it around in case someone opens a stack overflow issue again (or I can open the PR anyway)

https://github.com/purescript/spago/compare/master...cakekindel:fix-glob-perf?expand=1

f-f: Ah I see - if the performance is comparable and we can make it stack safe then that's a welcome addition

PR #1250, opened 2024-07-08
Update tools used in GitHub Actions

ysangkok:
### Description of the change

I am worried about whether the cache gets updated correctly. So I wanted to bump its version. But many other tools are also out of date, so I bumped them too.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Setup node and npm
         uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 22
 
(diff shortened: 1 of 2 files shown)


f-f:
Let's update the node version here while we're at it

f-f (approved):

ysangkok:
@f-f Looks like this failed on after merging to main... Is it because of this work or is it expected to be flaky?

> #1161 regression: 'subpackage with disjoint git dependency' builds:

>  test timed out after 90000ms

f-f:
It's not consistently flaky, and it's the first time that I see it fail. I rerun that job and now it's passing - I'll keep an eye on it to see if it will fail again

PR #1251, opened 2024-07-09
glob/fsWalk: early exclusion of non-matching directories

### LINKED ISSUE #1243: Stuck at `purs graph`
Hello,

I'm using spago 0.93.35 with purs 0.15.15. My project was building recently, but currently for some unknown reason, in `--verbose` mode it's stuck at:
```
[   94065ms] Called `purs graph`, decoding..
```
I tried to repeat that `purs graph` command, it does spit out a very large JSON object (containing 1185 keys).

Thing is, it's stuck for very long and CPU does crunch something.

Looking at source code, I think it might be decoding that large JSON for so long -- is it possible? And if so, is there a fix for this?

Blugatroff:
### Description of the change

For glob patterns with a fixed prefix, fsWalk will now no longer traverse directories whose contents obviously cannot match, because the directory path doesn't match the fixed prefix of the pattern.

For example, previously, globbing for `.spago/p/tailrec-6.1.0/**/*.purs` would traverse every directory (apart from those in a .gitignore, although in the case of globbing for something in .spago, that's not of much help).
Now only directories whose relative path starts with `.spago/p/tailrec-6.1.0` would be recursed into.

To determine the longest fixed prefix of a glob pattern, i used the [scan](https://www.npmjs.com/package/picomatch#scan) function from `picomatch`, which is already a dependency of micromatch.

Fix #1243 as per https://github.com/purescript/spago/issues/1243#issuecomment-2211732428

### Checklist:

### DIFF
--- a/src/Spago/Glob.js
+++ b/src/Spago/Glob.js
@@ -1,4 +1,5 @@
 import mm from 'micromatch';
+import picomatch from 'picomatch';
 import * as fsWalk from '@nodelib/fs.walk';
 
 export const testGlob = glob => mm.matcher(glob.include, { ignore: glob.ignore });
@@ -13,3 +14,6 @@ export const fsWalkImpl = Left => Right => respond => options => path => () => {
 };
 
 export const isFile = dirent => dirent.isFile();
+
+export const scanPattern = picomatch.scan
+
--- a/src/Spago/Glob.purs
+++ b/src/Spago/Glob.purs
@@ -32,6 +32,22 @@ type FsWalkOptions = { entryFilter :: Entry -> Effect Boolean, deepFilter :: Ent
 foreign import data DirEnt :: Type
 foreign import isFile :: DirEnt -> Boolean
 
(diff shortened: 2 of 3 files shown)


src/Spago/Glob.js
```diff
@@ -1,4 +1,5 @@
 import mm from 'micromatch';
+import picomatch from 'picomatch';
```
f-f, line 2: Now that we import explicitly from `picomatch` we should add it as a dependency to the `package.json`

f-f (approved):
Excellent patch @Blugatroff ❤️ 

With this the running time on the [purescript-core](https://github.com/purescm/purescript-core) repo goes from 30s to less than 5s, which is way nicer.

I left a small note about npm dependencies, but otherwise this is good to merge

src/Spago/Glob.js
f-f: I'll just merge it in and add this while I release 🙂

PR #1252, opened 2024-07-14
documentation tweak about bundle

peterbecich:
### Description of the change

One more example in the documentation about how to bundle Spago and use this bundle.

### Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,7 +47,11 @@ spago bundle -p spago-bin
 ./bin/bundle.js bundle -p spago-bin
 
 # From now on you can use the bootstrapped build to see the changes you make:
-./bin/index.dev.js some-new-command
+./bin/index.dev.js some-spago-command
+
+# Example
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -47,7 +47,11 @@ spago bundle -p spago-bin
 ./bin/bundle.js bundle -p spago-bin
 
 # From now on you can use the bootstrapped build to see the changes you make:
-./bin/index.dev.js some-new-command
+./bin/index.dev.js some-spago-command
+
+# Example
+./bin/index.dev.js --version
+# Currently prints: 0.93.35
```
f-f, line 54: The next example uses `build`, so that might be more contextual:

```suggestion
./bin/index.dev.js build --pedantic-packages --some-new-build-flag
```

peterbecich: Good idea

f-f (approved):

Issue #1237, opened 2024-06-25
In monorepos, allow running commands from package directories


roryc89:
In a monorepo it would more convenient to allow some commands (`build`, `install`, `uninstall`, `run`, `test`, `bundle`) to be run from the package, instead of having to run them from the workspace directory with the package flag. 

As our project is very large, I often work on individual packages as the IDE is slow when working on the whole workspace. It is then a bit cumbersome to have to change directory and add the package flag in order to run any spago commands on the package. 

Currently to install a dependency when working in a monorepo the commands look something like this: 

```sh
cd ../..
spago install -p lib-components some-dep
cd lib/components
```

ideally I would just like to run: 
```sh
spago install some-dep
```

f-f:
I have wished for this before, and we don't have it yet because of UX concerns - I couldn't find a satisfactory answer to "what happens when there's no workspace in the current folder".

That is, Spago will need workspace info for all the commands you mentioned (since it has to find the monorepo output folder, the lockfile, possibly resolve a build plan, etc).
Would we start looking up into parent directories? Is there a stopping point for how high to climb? Do we traverse links? (which might lead to loops). And what if there's no workspace in any parent folder? (Stack has a "default project" in the user's home folder, that is used when this happens. Not sure I'm a fan of that)

roryc89:
I think having a reasonably low max height to search would be reasonable (8?), perhaps with the option to increase it in the package config. 

If there's no workspace found, throw an informative error. 

This might not be perfect but I think it would be an improvement on the current UX in these situations. 

I agree that I wouldn't follow Stack's default project in home folder setup.

f-f:
> I think having a low max height to search would be reasonable (8?), perhaps with the option to increase it in the package config.
> If there's no workspace found, throw an informative error.

Let's try this out, but keep it at 4 levels of directories (following symlinks), and with no additional configuration - if someone needs more than that we can discuss bumping up the limit, but the priority is to keep the amount of available configuration knobs as low as possible.

peterbecich:
> Is there a stopping point for how high to climb?

How about `.git` as the stopping point?

Emacs `projectile` does it this way; `.git`, build tool file, or `.projectile` marker file. Obviously second point is not helpful.
[...]

PR #1253, opened 2024-07-14
search parent dirs for workspace

peterbecich:
https://github.com/purescript/spago/issues/1237

### Description of the change

Search up to 4 levels up the file tree for a `workspace`

Stop at `.git`

This search only happens if a `workspace` not found in current working directory

**Test of finding a `workspace` in parent dir successfully**

`example/spago.yaml` has no workspace
`example/../spago.yaml` has a workspace

```

# current working directory is 'example'

$ ~/purescript/libraries/spago/bin/index.dev.js build
Looking for Spago workspace configuration higher in the filesystem, up to project root (.git)...
Checking for workspace: /home/peterbecich/haskell/libraries/purescript-bridge/example/../spago.yaml
Found workspace definition in /home/peterbecich/haskell/libraries/purescript-bridge/example/../spago.yaml

✅ Selecting package to build: example

Downloading dependencies...
Building...
           Src   Lib   All
Warnings     0     0     0
Errors       0     0     0

✅ Build succeeded.



```

**Test of failing to find a workspace in parent directories**

`example/spago.yaml` and `example/../spago.yaml` have no workspace

```
$ ~/purescript/libraries/spago/bin/index.dev.js build
Looking for Spago workspace configuration higher in the filesystem, up to project root (.git)...
Checking for workspace: /home/peterbecich/haskell/libraries/purescript-bridge/example/../spago.yaml

❌ No workspace definition found in this directory
or in any directory up to root of project.
Root determined by '.git' file.
See the relevant documentation here: https://github.com/purescript/spago#the-workspace



```

I have not carefully tested the search stops at `.git`.

### Checklist:


peterbecich:
@f-f , please review

It does work.

It should stop searching at the directory which contains `.git`, but I have not carefully tested this.

peterbecich:
I think the feature to stop at .git is broken. The use of Glob is probably incorrect.

f-f:
My understanding of this patch is that it adds functionality for finding the workspace if it's reasonably close - I have read through the code but won't review it further until we add a few important test cases that might prompt for a redesign of the patch.

In particular, there are two situations that we should cover - these are good starting points to add as tests:
1. running `spago build` from a folder `project/b`, where the package `b` depends on a package that is in `project/a`, and the workspace is in `project`. I suspect this will fail with the current patch, since we find the workspace file, but we don't have a complete workspace! The packages in the subdirectories from the workspace are all part of the workspace.
2. read a file in the workspace root when running `spago run` both from the workspace folder, and from a subdirectory. The idea is that `spago run` (and `spago test` too) need to always run from the same place, and through user testing we have found that the root of the workspace is the least surprising place

Both of these suggest that a good course of action might be to:
1. find out which package we have at hand in the `spago.yaml`
2. find out where the workspace is, if any
3. move the `cwd` there

That's what I would try, and I think it will solve most of our issues, but there might be corner cases that we'll need to address then

f-f:
@peterbecich anything I can do to help move this forward?

peterbecich:
I have fixed this: https://github.com/purescript/spago/pull/1253#issuecomment-2241522568
It stops at `.git` successfully.  Here is a test.

The `cwd` is `purescript-bridge/test/RoundTripArgonautAesonGeneric/app`. The non-workspace `spago.yaml` is there.
The workspace `spago.yaml` is 3 levels up in `purescript-bridge`.


```
Reading spago.yaml...
Looking for Spago workspace configuration higher in the filesystem.
Search limited to 4 levels, or project root (.git)...
Project root (.git) found at: 
/home/peterbecich/haskell/libraries/purescript-bridge/test/RoundTripArgonautAesonGeneric/app/../../../.git
Found workspace definition in: 
/home/peterbecich/haskell/libraries/purescript-bridge/test/RoundTripArgonautAesonGeneric/app/../../../spago.yaml

✓ Selecting package to build: argonaut-round-trip-test

Downloading dependencies...
Building...
           Src   Lib   All
Warnings     0     0     0
Errors       0     0     0

✓ Build succeeded.


```

Regarding the improvements to be done,


> move the cwd there

I can't figure out how to implement this.

The PR does not fix the two cases you have identified. However it does cover the simplest case, where the workspace is several levels up from the `cwd` non-workspace `spago.yaml`.
Can we merge it as-is and cover the two cases later?
My reasoning is that it does not really introduce any regression. It is merely the feature is half-implemented.

src/Spago/Config.purs
```diff
@@ -164,11 +165,22 @@ type ReadWorkspaceOptions =
...
+  , extraPackages :: Maybe (Map PackageName Core.ExtraPackage)
+  , packageSet :: Maybe Core.SetAddress
+  }
+
 -- | Reads all the configurations in the tree and builds up the Map of local
 -- | packages to be integrated in the package set
 readWorkspace :: ∀ a. ReadWorkspaceOptions -> Spago (Registry.RegistryEnv a) Workspace
 readWorkspace { maybeSelectedPackage, pureBuild, migrateConfig } = do
-  logInfo "Reading Spago workspace configuration..."
+  logInfo "Reading spago.yaml..."
```
peterbecich, line 183: i.m.o. this is more correct because it is reading a `spago.yaml` that may not contain a workspace.

f-f:
> > move the cwd there
> I can't figure out how to implement this.

See [`Node.Process.chdir`](https://pursuit.purescript.org/packages/purescript-node-process/11.2.0/docs/Node.Process#v:chdir)

> The PR does not fix the two cases you have identified. However it does cover the simplest case, where the workspace is several levels up from the cwd non-workspace spago.yaml.
> Can we merge it as-is and cover the two cases later?
> My reasoning is that it does not really introduce any regression. Instead, the feature is half-implemented.

The two tests I mention are in fact potential regressions that I believe this patch can trigger, and we will not know if this patch introduces these regressions until we explicitly have these tests.

src/Spago/Config.purs
f-f: Sure

f-f:
Superseded by #1310

PR #1255, opened 2024-07-25
`spago repl` writes a `.purs-repl` file unless already there

### LINKED ISSUE #1192: Create a `.purs-repl` file when running `spago repl`
The old spago [does that](https://github.com/purescript/spago-legacy/blob/4626a2a140b4fe0adc1445ff91912ab4a687e4da/src/Spago/Build.hs#L215) ([with this content](https://github.com/purescript/spago-legacy/blob/master/templates/purs-repl)), and the new spago [doesn't](https://github.com/purescript/spago/blob/master/src/Spago/Command/Repl.purs)

fsoikin:
### Description of the change

Fixes #1192

If there is not already a `.purs-repl` in the current directory, `spago repl` will write a new one. 
The template is shared with `spago init` to avoid divergence in the future.

### Checklist:


f-f (approved):
This is great, thank you!

PR #1259, opened 2024-07-28
spago publish verifies that the publish location matches one of the current repo's remotes

### LINKED ISSUE #1179: Check existing of repo before contacting registry
This is a reminder issue for comment https://github.com/purescript/registry-dev/issues/684#issuecomment-1904330124

Check that a git repository exists before sending it to the registry.

fsoikin:
### Description of the change

Fixes #1179

Rather than checking that the repository exists on GitHub, I opted to proxy that by checking that the publish location matches one of the remotes in the current local repo. There are several reasons to do this:
1. Works in offline mode.
2. Works for (potentially) non-GitHub repositories.
3. One fewer network requests.
4. Actually a better check: even if the publish location does exist, it may be a completely different, unrelated repository, which may happen, for example, when copy&pasting config to bootstrap a new project. Checking against local remotes makes it more likely that the publish location is actually location of this project.
5. An escape hatch in case of bugs. If something doesn't work, you can shoehorn it by adding a dummy remote with the right names.

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Other improvements:
 - Added support for `--package-set` options for `spago upgrade`.
 - `spago repl` now writes a `.purs-repl` file, unless already there, containing `import Prelude`.
 - Added typo suggestions upon failing to find a package by name.
+- `spago publish` now checks that the publish location matches one of the remotes in the current Git repository.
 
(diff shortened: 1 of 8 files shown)


f-f (approved):
I quite like this 🙂

PR #1260, opened 2024-07-29
Refuse to overwrite existing target bundle file unless it was generated by Spago in the first place

fsoikin:
### Description of the change

Fixes #1921

But this is a somewhat breaking change. Not completely breaking, but every existing project will need to either delete the bundle or build it at least once with `--force-overwrite`.

### Checklist:


src/Spago/Command/Bundle.purs
```diff
@@ -35,7 +41,7 @@ type RawBundleOptions =
   , extraArgs :: Array String
   }
 
-run :: forall a. Spago (BundleEnv a) Unit
+run :: ∀ a. Spago (BundleEnv a) Unit
```
fsoikin, line 44: I reorganized this function a little to reduce noise, I hope it's ok.

test-fixtures/bundle-app-node.js
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-import __module from 'module';import __path from 'path';import __url from 'url';const require = __module.createRequire(import.meta.url);const __dirname = __path.dirname(__url.fileURLToPath(import.meta.url));const __filename=new URL(import.meta.url).pathname
+/* Generated by Spago v0 */import __module from 'module';import __path from 'path';import __url from 'url';const require = __module.createRequire(import.meta.url);const __dirname = __path.dirname(__url.fileURLToPath(import.meta.url));const __filename=new URL(import.meta.url).pathname
```
fsoikin, line 2: Unfortunately the comment cannot be on a separate line, because the banner cannot contain newlines, which is in turn because on Windows command-line parameters cannot contain newlines. 

This can be worked around by importing `esbuild` as a Node dependency and calling it directly from PureScript via JS API rather than via command line. But that seemed like too much work for such insignificant gain.

bin/src/Flags.purs
```diff
@@ -69,9 +69,16 @@ outfile =
...
+            <> O.help "Destination path for the bundle. Defaults to ./index.js"
         )
 
+forceOverwrite :: Parser Boolean
+forceOverwrite =
+  O.switch
+    ( O.long "force-overwrite"
+        <> O.help "Overwrite the target bundle file even if it was manually created or generated by a tool other than Spago."
+    )
+
```
f-f, line 81: I think we could shorten this to `--force`, while at the same time having a more specific name to imply the flag is only used for bundling (as other parts of the CLI could use `--force` for other things)
```suggestion
forceBundle :: Parser Boolean
forceBundle =
  O.switch
    ( O.long "force"
        <> O.help "Overwrite the target bundle file even if it was manually created or generated by a tool other than Spago."
    )

```

src/Spago/Command/Bundle.purs
f-f: Yes, it looks great!

src/Spago/Command/Bundle.purs
```diff
@@ -47,38 +53,85 @@ run = do
...
+        ]
+      , opts.extraArgs
+      , minify
+      , sourceMap
+      , entrypoint
+      ]
+
+  unless opts.forceOverwrite
+    $ whenM (isNotSpagoGeneratedFile outfile)
+    $ die [ "Target file " <> opts.outfile <> " was not previously generated by Spago. Use --force-overwrite to overwrite anyway." ]
```
f-f, line 100: Let's make this change be not breaking in the immediate term by commenting out this check for now, and reintroducing it 6 months down the line. In this way as folks upgrade to the new versions they will start producing bundles with the banner, which will pass the test once we put it back in.

We should leave the tests in too (fixtures and everything), and comment them out too, so they are ready to go when the time comes

```suggestion
  _refuseToOverwriteBundle <- isNotSpagoGeneratedFile outfile
  -- when (refuseToOverwriteBundle && not opts.forceOverwrite) do
  --   die [ "Target file " <> opts.outfile <> " was not previously generated by Spago. Use --force to overwrite anyway." ]
```

src/Spago/Command/Bundle.purs
```diff
@@ -47,38 +53,85 @@ run = do
...
-      , "--outfile=" <> outfile
-      , "--platform=" <> show opts.platform
-      -- See https://github.com/evanw/esbuild/issues/1051
-      , "--loader:.node=file"
-      , format
-      ] <> opts.extraArgs <> minify <> sourceMap <> entrypoint <> nodePatch
+    banner = fold
+      [ spagoGeneratedFilePrefix
+      , " v"
+      , BuildInfo.packages."spago-bin"
```
f-f, line 79: I am confused why this version is 0 in the fixtures

f-f: Ah I think it's because the "packages" field will contain the version for the packages being compiled. None of the packages in the fixtures is called "spago-bin", so it comes out as 0.

Since we want to print Spago's version in here, I think we should add a `spagoVersion` field in the `BuildInfo` (I was entirely convinced it was there already). Probably in a separate PR.

test-fixtures/bundle-app-node.js
f-f: We can't import esbuild as a node dependency, because we won't be able to bundle it in and distribute that: it's written in Go so it relies on funny JS packaging to be distributed. In any case, this change is fine as is 🙂

src/Spago/Command/Bundle.purs
fsoikin: It doesn't "come out" as 0. I manually replaced it with 0.

The problem is, it's supposed to be the "current" version of Spago, which means it can't be any particular version. The current version has to be inserted there during the test run, as that's the only way to get the "current" version. So the "v0" bit is just a placeholder, and it's replaced [here](https://github.com/purescript/spago/pull/1260/files#diff-d4d0ded0a864b48583b9f1fdb5c79023288c9df5fc4bd4499b8d1233b78b5dd9R73-R82), as the comment says.

test-fixtures/bundle-app-node.js
fsoikin: Well, an alternative could be to write a JS file that imports and runs esbuild, and then run that file.

src/Spago/Command/Bundle.purs
fsoikin: I'm worried that the tests and/or the main code may rot over time from disuse.

How about instead make this check dependent on the presence of a special marker file, and then make the tests write that file so that the check will run during tests.

In addition to the marker file, we could make the check dependent on time (e.g. start running it in 2025) or on current version (e.g. start running after 1.0.0). We'd still have to have the marker file for the tests though.

fsoikin: I added the marker file logic. Let me know if you would like the time or version condition.

bin/src/Flags.purs
fsoikin: Done

src/Spago/Command/Bundle.purs
f-f: Oh yeah makes sense - thanks for the clarification!

src/Spago/Command/Bundle.purs
f-f: I wouldn't like to create more files. Let's make this dependent on the spago version (+5 patch versions from now or `1.0.0`, whichever comes first) instead. We won't be able to test this now, but we can already check for the watermark, and we can have tests break once we get past the release threshold - I think it's good enough to prevent rot

f-f: Oh I see - for some reason I entirely missed that comment. Makes sense now!

src/Spago/Command/Bundle.purs
```diff
@@ -47,38 +53,103 @@ run = do
...
+        , "--loader:.node=file" -- See https://github.com/evanw/esbuild/issues/1051
+        , format
+        ]
+      , opts.extraArgs
+      , minify
+      , sourceMap
+      , entrypoint
+      ]
+
+  whenM (FS.exists checkWatermarkMarkerFileName)
```
f-f, line 97:
```suggestion
  -- FIXME: remove this after 2024-12-01
  whenM (FS.exists checkWatermarkMarkerFileName)
```

test/Spago/Bundle.purs
```diff
@@ -2,54 +2,94 @@ module Test.Spago.Bundle where
...
+import Spago.Generated.BuildInfo as BuildInfo
 import Test.Spec (Spec)
 import Test.Spec as Spec
+import Test.Spec.Assertions (shouldNotEqual)
+import Test.Spec.Assertions.String (shouldStartWith)
 
 spec :: Spec Unit
 spec = Spec.around withTempDir do
-  Spec.describe "bundle" do
+  Spec.focus $ Spec.describe "bundle" do
```
f-f, line 16:
```suggestion
  Spec.describe "bundle" do
```

f-f (approved):

test/Spago/Bundle.purs
f-f: No worries - I have merged those in quite a few times already, so extra sensitive to them now.

I wonder how we could prevent merging those in at all - maybe spec should allow for focusing using environment variables like many other test runners do, that would remove the need to edit the source at all

f-f: Yes please! I have been wishing for this in spec for some years already, but my todo list always ends up too long 😄

PR #1261, opened 2024-08-01
Slow lockfile gen

finnhodgkin:
### Description of the change

#1262 
Lockfile generation on my large work project takes over 3 minutes.
Adding a bunch of memoisation and combining many individual sqlite queries into one gets this down to 1.5 seconds.

### Checklist:


f-f:
@finnhodgkin I am interested in merging this, perf is something we care about a lot. You can go ahead with the cleanup and we can get this merged. But I can also review before that if that's something you'd like.

finnhodgkin:
In my head I was going to clean this up straight after pushing but got distracted. I'll sort it tomorrow. 

The thing I could do with a hand on is the sqlite query. I _think_ it's safe to leave un-paramatised because it's called with `Array PackageName` which is sanitised, but coming from postgres it just feels wrong. Spreading parameters hit the upper limit of params and a single array param just doesn't seem to be possible. Currently:

```
export const getMetadatasImpl = (db, names) => {
  let query = 'SELECT * FROM package_metadata WHERE name IN (@names)';
  return db.prepare(query).all({ names: names.map((n) => `'${n}'`).join(',') });
};
```

finnhodgkin:
@f-f should be good to review now as long as CI goes smoothly.

src/Spago/Command/Fetch.purs
```diff
@@ -279,16 +281,39 @@ type LockfileBuilderResult =
...
+
+updateCache :: forall a k v. Ord k => k -> v -> Ref.Ref (Map k v) -> Spago a Unit
+updateCache key value cacheRef = liftEffect $ Ref.modify_ (Map.insert key value) cacheRef
+
 writeNewLockfile :: forall a. String -> PackageTransitiveDeps -> Spago (FetchEnv a) PackageTransitiveDeps
 writeNewLockfile reason allTransitiveDeps = do
   logInfo $ reason <> ", generating it..."
   { workspace } <- ask
+
+  packageDependenciesCache <- liftEffect $ Ref.new Map.empty
```
f-f, line 295:
```suggestion
  -- All these Refs are needed to memoise Db and file reads
  packageDependenciesCache <- liftEffect $ Ref.new Map.empty
```

src/Spago/Command/Fetch.purs
```diff
@@ -303,28 +328,83 @@ writeNewLockfile reason allTransitiveDeps = do
...
+            _ -> pure Nothing
+        )
+        $ allDependencies
+    )
+
   ({ packages, workspacePackages } :: LockfileBuilderResult) <-
-    Array.foldM processPackage
-      { workspacePackages: Map.fromFoldable $ map Config.workspacePackageToLockfilePackage (Config.getWorkspacePackages workspace.packageSet), packages: Map.empty }
-      (foldMap sequence $ toArray $ map toArray allTransitiveDeps)
+    -- Array.foldM was significantly slower (~10ms vs 6s on a very large project)
```
f-f, line 402:
```suggestion
    -- NOTE! We used to have `Array.foldM` here, but it was significantly slower
    -- (~10ms now vs 6s then, on a very large project)
```

src/Spago/Db.js
```diff
@@ -117,3 +117,9 @@ export const getMetadataImpl = (db, name) => {
 export const insertMetadataImpl = (db, name, metadata, last_fetched) => {
   db.prepare("INSERT OR REPLACE INTO package_metadata (name, metadata, last_fetched) VALUES (@name, @metadata, @last_fetched)").run({ name, metadata, last_fetched });
 }
+
+export const getMetadatasImpl = (db, names) => {
+  // there can be a lot of package names here, potentially hitting the max number of sqlite parameters, use json to bypass this
```
f-f, line 122:
```suggestion
  // There can be a lot of package names here, potentially hitting the max number of sqlite parameters, so we use json to bypass this
  // See https://github.com/WiseLibs/better-sqlite3/issues/283#issuecomment-1044105886
```

src/Spago/Registry.purs
```diff
@@ -62,6 +63,7 @@ type RegistryEnv a = Record (RegistryEnvRow a)
 type RegistryFunctions =
   { getManifestFromIndex :: PackageName -> Version -> Spago (LogEnv ()) (Maybe Manifest)
   , getMetadata :: PackageName -> Spago (LogEnv ()) (Either String Metadata)
+  , getMetadatas :: Array PackageName -> Spago (LogEnv ()) (Either String (Map PackageName Metadata))
```
f-f, line 66: Is there any point in keeping `getMetadata` as well now that we have `getMetadatas`?

As an aside, `data` is a latin word that's already plural, so adding the `s` suffix doesn't feel right. How about renaming to `getMetadataForPackages`?

f-f (approved):
This is great now 👏 

We should adjust a few comments and names, but it's otherwise good to go

src/Spago/Registry.purs
finnhodgkin: The singular version is still used in Command.Repl, Command.Publish and Command.Registry so I just left it alone.

Thanks, will swap to suggested name, had the same thought re datas but wasn't sure about a good replacement.

f-f: > The singular version is still used in Command.Repl, Command.Publish and Command.Registry so I just left it alone.

Sure, but we can now call the new implementation passing in a singleton array `[packageName]`? Having two implementations for the same thing guarantees that they will go out of sync in the future, so I'd rather remove the singular version now.

finnhodgkin: Oh sure. Like this?

finnhodgkin: I could replace the function entirely but it feels useful being able to look up a single package without `Map.lookup`ing the results.

f-f:
@finnhodgkin could you merge with latest?

src/Spago/Registry.purs
f-f: Yeah that makes sense, thanks!

f-f:
Thank you!

PR #1263, opened 2024-08-04
Migrate tests to the spec-node runner

fsoikin:
### Description of the change

So here it is, as [promised](https://github.com/purescript/spago/pull/1260#discussion_r1698535771), the shiny new spec runner that persists results and takes command-line options.

Full docs here: https://purescript-spec.github.io/purescript-spec/running/#running-under-node

But here are some examples of possible use:

```
 λ npx spago test -- -E "ModuleParser|bundle"

Running tests for package: docs-search-index
ModuleParser
  ✓︎ test #0
  ✓︎ test #1
  ✓︎ test #2
  ✓︎ test #3
  ✓︎ test #4
  ✓︎ test #5
  ✓︎ test #6
  ✓︎ test #7
  ✓︎ multiline comment #1
  ✓︎ multiline comment #1
  ✓︎ multiline comment #1
  ✓︎ single line comment #1
  ✓︎ single line comment #1
Modules in .spago: 1113
  ✓︎ Parses every module in .spago/

Summary
14/14 tests passed


✓ Test succeeded for package "docs-search-index".

Running tests for package: spago
spago » bundle
  ✓︎ bundles into an app (browser)
  ✓︎ bundles into an app (node)
  ✓︎ bundles into a module
  ✓︎ bundles an app with source map (browser)
  ✓︎ bundles an app with source map (node)
  ✓︎ bundles a module with source map
  ✓︎ bundles a module with extra esbuild arguments
  ✓︎ refuses to overwrite existing bundle that is not Spago-generated
  ✓︎ overwrites non-Spago-generated bundle when there is no magic marker file

Summary
9/9 tests passed


✓ Test succeeded for package "spago".

 λ npx spago test -- -e "into an app (browser)" --timeout 1

Running tests for package: docs-search-index

Summary
0/0 tests passed


✓ Test succeeded for package "docs-search-index".

Running tests for package: spago
spago » bundle
  ✗ bundles into an app (browser):

  test timed out after 1000ms

Summary
0/1 test passed


✘ Tests failed for package "spago".

 λ npx spago test -- --only-failures

Running tests for package: docs-search-index

Summary
0/0 tests passed


✓ Test succeeded for package "docs-search-index".

Running tests for package: spago
spago » bundle
  ✓︎ bundles into an app (browser)

Summary
1/1 test passed


✓ Test succeeded for package "spago".
```

### Checklist:

- [x] Added the change to the "Unreleased" section of the changelog
- ~[ ] Added some example of the new feature to the `README`~
[...]

Issue #573, opened 2020-02-17
Add `--template` flag to `init` command


f-f:
As mentioned in [this comment on Discourse](https://discourse.purescript.org/t/recommended-tooling-for-purescript-applications-in-2019/948/17) it would be really nice if Spago could have templates.

Examples from the post for how this could look like:
```
spago init --template halogen
spago init --template react-basic-hooks
spago init --template cli
... etc.
```

Opening this issue to track this, I don't have an opinion on how this should be implemented, but looking at previous art (e.g. Stack templates), I feel we could:
- have each template as a git submodule in the `templates` folder
- ..and include all templates in the binary (this is so that you can still make a project if you're offline)
- ..and optionally try to fetch them if we are online and they have been updated

JordanMartinez:
Just a "way out there" question. How would this work if a project could be implemented by multiple backends? For example, the `cli` one. I see a few options:
1. Specify the backend in the template name (e.g. `cli-js`, `cli-go`, `cli-c`)
2. Use a separate flag to indicate the backend (e.g. `--template cli --template-backend js`)
3. Specify the backed in the template flag itself (e.g. `--template-js cli` vs `--template-go cli`)

I think first option makes the most sense.

milesfrain:
Is `spago` the right tool for managing templates, especially with multiple backends?

I've had such a great experience with `spago init`, that `--template` seemed like a natural extension.

But after reading "[Why can't `spago` also install my npm dependencies?](https://github.com/purescript/spago#why-cant-spago-also-install-my-npm-dependencies)", it seems like spago should be a dependency of the templates (rather than the templates being a dependency of spago).

So instead of:
```
spago init --template <template-name>@<template-version>
```
we could just suggest:
```
git clone --depth 1 --branch <template-version-tag> <template-url>
```

Starting with the `clone` approach is a low-barrier way to start deciding how we want to organize these templates and what the template guidelines should be. We could always feed this listing into the `--template` strategy later with minimal throwaway work.

Curating an initial list of templates seems outside the scope of this spago task though. I'm wondering where this effort should be tracked. Considering making another Discourse thread, or working on a page in documentation repo where we can start building a table of templates.

JordanMartinez:
I agree with @milesfrain. This feels more like a documentation issue (i.e. "What are various starter/template projects that I can use to get started on X app?") than a feature that Spago should support as it avoids a number of issues.

[...]

PR #1264, opened 2024-08-10
spago init template based on Spec

fsoikin:
### Description of the change

It occurred to me that, since [we're recommending `spec-node` in the docs](https://github.com/purescript/spago/pull/1263#pullrequestreview-2221076069) now, it would make sense to include it in the `spago init` template. So the user doesn't have to jump through extra hoops.

I understand this may not be 100% desirable because it introduced extra moving parts. Feel free to throw away. Just an idea.

### Checklist:


test/Prelude.purs
```diff
@@ -26,7 +26,7 @@ import Spago.Command.Init as Init
...
... bottom, catchError, clamp, compare, comparing, compose, conj, const, degree, die, die', discard, disj, div, either, eq, except, flap, flip, foldMap, foldl, for, forWithIndex, for_, fromMaybe, fst, gcd, genericShow, hush, identity, ifM, indent, indent2, isJust, isLeft, isNothing, isPrefix, isRight, join, justOrDieWith, justOrDieWith', lcm, liftA1, liftAff, liftEffect, liftM1, lmap, local, logDebug, logError, logInfo, logSuccess, logWarn, map, max, maybe, mempty, min, mkTemp, mkTemp', mod, mul, negate, not, notEq, on, one, or, otherwise, output, parTraverseSpago, parallelise, parseJson, parseLenientVersion, parseUrl, parseYaml, partition, partitionEithers, partitionMap, printJson, printYaml, pure, recip, rightOrDieWith, rightOrDieWith', rmap, runSpago, shaToHex, show, snd, sub, toDoc, top, traverse, try, typoSuggestions, unit, unless, unlessM, unsafeFromJust, unsafeFromRight, unsafeLog, unsafeStringify, unsafeThrow, unwrap, void, when, whenM, withBackoff', withForwardSlashes, zero, (#), ($), ($>), (&&), (*), (*>), (+), (-), (..), (/), (/=), (/\), (:), (<), (<#>), (<$), (<$>), (<*), (<*>), (<<<), (<=), (<=<), (<>), (<@>), (<|>), (=<<), (==), (>), (>=), (>=>), (>>=), (>>>), (||)) as X
```
f-f, line 29: Can we rollback this line? It's just confusing to have this huge import like this

f-f (approved):
Ah yeah, this makes sense!

test/Prelude.purs
fsoikin: Oh yeah, sorry, that's accidental, from IDE support.

f-f:
I think this is a good change, but after all the updates to the tests the CI times have ballooned - they are now 2x slower than the current trunk.

That's a bit of a deal breaker, so let's think about an alternative approach; the no-hassle one is to add some `--no-tests` flag to `spago init`, to skip the `spec-node` addition and use the current setup - we could then use this flag for most of the tests, and only add the new testing setup to a few of them, while still keeping it as the default for spago.
However, that would not be extensible. A more general approach could be to introduce templates for `spago init` - then we could have a `basic` template that would contain the bare minimum (basically as it is now), and a `default` template that would add `spec-node`. This would open the door to more complex things, e.g. a `halogen` template, and so on.
Does this make any sense?

fsoikin:
Yes, I agree, this became quite unmanageable. I was hoping it would literally just be changing the template, but it turns out all the tests depend on it, and it's quite a ball of yarn. 

I do like the idea of multiple templates. Other tools have similar things (e.g. dotnet) and they're very popular. But I don't think it is critical for the current milestone. So I guess we should just table this change. Or throw it away altogether.

f-f:
Yeah let's not move forward with this as is, and I agree with you that it's not critical for the current milestone. I have reopened #573 to track this effort.

Issue #1262, opened 2024-08-01
Super slow lockfile parsing and generating


finnhodgkin:
At my work we have a very large workspace with many libs.

Adding or removing any modules from any of the libs takes more than three minutes. I've narrowed most of this time down to lockfile generation. 

Adding a bunch of extra caching and combining queries gets that time down to <1.5s.

finnhodgkin:
A somewhat linked issue to this is that the lockfile can become insanely large, and even reading and parsing it becomes pretty slow. My work codebase has a lockfile with 123k lines and takes this long to parse & decode:

<img width="589" alt="image" src="https://github.com/user-attachments/assets/4aea9448-46d5-41f8-b105-12c7e5dc59c8">

This wouldn't be too bad if it was occasional but every spago command triggers a parse+decode. 3 seconds on every command is definitely noticeable.

finnhodgkin:
This may be a me problem and not something that's going to impact the average user...

Our lockfile is made up of 120k lines of `workspace.package.[package with build plan]` and then a few thousand lines for everything else. 

@f-f I know very little about the internals so this might be a stupid question, but are the build plans necessary or could they be worked out at runtime by looking up the transient deps and nubbing the results? Here's one example from our lockfile:

```
dependencies:
  - argonaut-codecs
  - bifunctors
  - console
  - effect
  - either
  - foreign
  - foreign-generic
  - graphql-client
  - oa-make-fixture
  - prelude
  - transformers
test_dependencies:
  - oa-no-tests
build_plan:
  - aff
  - aff-promise
  - affjax
  - affjax-node
  - affjax-web
  - ansi
  - argonaut
  - argonaut-codecs
  - argonaut-core
  - argonaut-traversals
  - arraybuffer-types
  - arrays
  - avar
  - bifunctors
  - catenable-lists
  - colors
  - console
  - const
  - contravariant
  - control
  - datetime
  - debug
  - distributive
  - effect
  - either
  - enums
  - exceptions
  - exists
  - filterable
  - foldable-traversable
  - foreign
  - foreign-generic
  - foreign-object
  - fork
  - form-urlencoded
  - free
  - functions
  - functors
  - gen
  - graphql-client
  - halogen-subscriptions
  - heterogeneous
  - http-methods
  - identity
  - integers
  - invariant
  - js-date
  - js-uri
  - lazy
  - lcg
  - lists
  - maybe
  - media-types
  - mmorph
  - newtype
  - nonempty
  - now
  - nullable
[...]

PR #1270, opened 2024-08-19
Allow to publish packages with test dependencies not in the registry 

fsoikin:
### Description of the change

The primary motivation for this is allowing circular dependencies between packages for testing only. In this scenario one of the packages is added to `extra_packages` section of the other with a patched `dependencies` property (see also [this Discord discussion](https://discord.com/channels/864614189094928394/873274004704747600/1269052656467443812)). For example:

```
# github/collegevine/purescript-elmish/spago.yaml
workspace:
  extra_packages:
    elmish-testing-library:
      repo: https://github.com/collegevine/purescript-elmish-testing-library.git
      version: v2.0.0
      dependencies: [] # <-- no Elmish in dependencies, avoiding the cycle

package:
  name: purescript-elmish
  dependencies: ...
  test:
    main: ...
    dependencies: ["elmish-testing-library"]
```

To achieve this, the following changes have been made:
* `Fetch.run` no longer returns a single list of dependencies for every package, but instead returns them in two buckets - "core" and "test".
* The lockfile format has been changed to support the same structure: 
    * from: `{ dependencies, build_plan, test_dependencies }` 
    * to: `{ core: { dependencies, build_plan }, test: { dependencies, build_plan } }`
* Most call sites of `Fetch.run` ultimately call `toAllDependencies`, which flattens the whole tree and combines `core` with `test`.
* `Publish.run` specifically picks only `core` and drops `test`, so as not to perform all the rigorous checks for the test dependencies.
* `Build.getBuildGlobs` no longer adds test globs (i.e. `test/**/*.purs`) for all workspace packages, but only for the "selected" ones. This is because only "selected" packages will have their test dependencies added to the build plan.

### Checklist:


src/Spago/Command/Fetch.purs
```diff
@@ -274,66 +279,68 @@ run { packages: packagesRequestedToInstall, ensureRanges, isTest, isRepl } = do
       Right _lockfile -> pure dependencies
       Left reason -> writeNewLockfile reason dependencies
 
-type LockfileBuilderResult =
-  { workspacePackages :: Map PackageName Lock.WorkspaceLockPackage
-  , packages :: Map PackageName Lock.LockEntry
-  }
-
 writeNewLockfile :: forall a. String -> PackageTransitiveDeps -> Spago (FetchEnv a) PackageTransitiveDeps
 writeNewLockfile reason allTransitiveDeps = do
```
fsoikin, line 283: I took the liberty to reorganize a bit how this function worked. It took me quite a while to understand the meaning, and once I did, it occurred to me that the implementation was needlessly convoluted. What it's really doing is separating "workspace packages" from "non-workspace packages" and then converting each set to their respective types of lockfile entry.

src/Spago/Command/Fetch.purs
```diff
@@ -615,3 +671,13 @@ cycleError :: PackageName -> TransitiveDepsResult -> TransitiveDepsResult
 cycleError dep result = result
   { errors { cycle = Set.insert dep result.errors.cycle } }
 
+-- | When we work with dependencies, we need to keep track of them separately
+-- | for core and test, because they need to be treated differently in some
+-- | contexts.
+type ByEnv a = { core :: a, test :: a }
```
fsoikin, line 677: I'm not a fan of the name "env" for this, but I couldn't come up with a better one.

src/Spago/Command/Fetch.purs
```diff
@@ -175,13 +178,16 @@ run { packages: packagesRequestedToInstall, ensureRanges, isTest, isRepl } = do
     let
       allTransitiveDeps = case isRepl of
         false -> dependencies
-        true -> map (\packageMap -> Map.union packageMap supportPackage) dependencies
+        true -> map (onEachEnv \packageMap -> Map.union packageMap supportPackage) dependencies
```
f-f, line 181:
```suggestion
        -- It should be enough to add the support package to the core dependencies only,
        -- but that's more code, so nevermind
        true -> map (onEachEnv \packageMap -> Map.union packageMap supportPackage) dependencies
```

src/Spago/Command/Fetch.purs
f-f: Good point! I didn't mind the previous shape, but I don't mind the new one either, so all good 🙂

src/Spago/Command/Fetch.purs
```diff
@@ -423,11 +439,13 @@ getPackageDependencies packageName package = case package of
...
 
-getWorkspacePackageDeps :: WorkspacePackage -> Dependencies
+getWorkspacePackageDeps :: WorkspacePackage -> ByEnv Dependencies
 getWorkspacePackageDeps pkg =
-  if pkg.hasTests then
-    pkg.package.dependencies <> fromMaybe mempty (map _.dependencies pkg.package.test)
-  else pkg.package.dependencies
+  { core: pkg.package.dependencies
+  , test: fromMaybe (Dependencies Map.empty) do
+      guard pkg.hasTests
```
f-f, line 446: I find `guard` to be confusing more often than not, and this is one of such cases. Could we have an explicit `case` here?

src/Spago/Command/Fetch.purs
f-f: Yeah. It feels like we could come up with a better one, but I can't think of anything as of now - it's good enough

test-fixtures/publish-extra-package-core-dependency.txt
```diff
@@ -0,0 +1,18 @@
...
+Errors       0     0     0
+
+✓ Build succeeded.
+
+Your package "aaa" is not ready for publishing yet, encountered 1 error:
+
+
+✘ Some of the packages you specified as `extra_packages` do not point to the Registry.
+To be able to publish a package to the registry, all of its dependencies have to be packages registered on the Registry.
+Please replace the following packages with versions that are present in the Registry:- "maybe"
```
f-f, line 18: I guess we need newlines for this list of package names

src/Spago/Command/Build.purs
```diff
@@ -202,7 +207,7 @@ getBuildGlobs { selected, dependencies, withTests, depsOnly } =
   dependencyGlobs = (Tuple.uncurry $ Config.sourceGlob NoTestGlobs) =<< dependencyPkgs
   monorepoPkgGlobs
     | depsOnly = []
-    | otherwise = (Tuple.uncurry $ Config.sourceGlob testGlobs) =<< monorepoPkgs
+    | otherwise = (Tuple.uncurry $ Config.sourceGlob NoTestGlobs) =<< monorepoPkgs
```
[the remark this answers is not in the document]
fsoikin: This is explained in the description: 

* `Build.getBuildGlobs` no longer adds test globs (i.e. `test/**/*.purs`) for all workspace packages, but only for the "selected" ones. This is because only "selected" packages will have their test dependencies added to the build plan.

spago.lock
```diff
@@ -2,840 +2,977 @@ workspace:
...
+          - unsafe-coerce
+          - variant
+          - versions
+      test:
+        dependencies:
+          - exceptions
+          - spec
+          - spec-node
+        build_plan:
+          - aff
```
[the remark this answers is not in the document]
fsoikin: The intended semantics is that `test.build_plan` does contain packages that are "additional" to the ones in `core.build_plan`. Sometimes it can happen that the test and core build plans intersect. This is because we're not doing an explicit diff, we're just building transitive closures separately - first starting with core dependencies, and then starting with test dependencies. If the test transitive closure also happens to touch some dependencies from core - so what, no big deal.

If you really insist, I can certainly add an explicit diff right before writing to the lockfile. This would be cleaner, but would incur additional code complexity and CPU cycles.

src/Spago/Command/Fetch.purs
fsoikin: Sure

spago.lock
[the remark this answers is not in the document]
f-f: Oh I see, we are already doing what I describe, minus the deduplication. I guess my concern is that the more lines there are in the yaml, the more it takes to parse it (see #1262); let's leave this as is, and sort out the yaml parsing performance otherwise

src/Spago/Command/Build.purs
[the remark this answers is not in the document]
f-f: Yeah I read that, but I'm still confused and worried that things will subtly break. Let's move on with this and see what happens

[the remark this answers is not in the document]
fsoikin: Maybe an example would help.

```
package:
  name: a
  test: { dependencies: [ x, y ] }
package:
  name: b
  test: { dependencies: [ p, q, a ] }
```

Note that `b`'s tests use `a` in some way. Perhaps `a` is a shared library with some cool utility functions that are useful in `b`'s tests.

When we run `spago test -p b`, we have to build `p`, `q`, and `a`, because, well, `b`'s tests depend on them.

Question: should we also build `x` and `y`?

My answer is "no". They are used in `a`'s tests, not in `a` itself, and it's only `a` itself that is used in `b`'s tests.

test-fixtures/publish-extra-package-core-dependency.txt
f-f: @fsoikin we should adjust this error then we are good to merge

fsoikin: Done. Sorry, I didn't realize that you implied that I should do it here. Looked like an idle musing kind of statement.

f-f: Ah yeah, apologies for not being clearer!

test-fixtures/monorepo/case3-shared-dependencies2/package-c/spago.yaml
```diff
@@ -7,4 +7,3 @@ package:
     dependencies:
       - console
       - effect
-      - prelude
```
f-f, line 10: Why remove these? They are testing a specific scenario (more info in the test code)

src/Spago/Command/Fetch.purs
```diff
@@ -276,18 +281,23 @@ run { packages: packagesRequestedToInstall, ensureRanges, isTest, isRepl } = do
...
-  { workspacePackages :: Map PackageName Lock.WorkspaceLockPackage
-  , packages :: Map PackageName Lock.LockEntry
-  }
-
-lookupInCache :: forall a k v. Ord k => k -> Ref.Ref (Map k v) -> Spago a (Maybe v)
+lookupInCache :: ∀ a k v. Ord k => k -> Ref.Ref (Map k v) -> Spago a (Maybe v)
 lookupInCache key cacheRef = liftEffect $ Ref.read cacheRef >>= Map.lookup key >>> pure
 
-updateCache :: forall a k v. Ord k => k -> v -> Ref.Ref (Map k v) -> Spago a Unit
+fetchFromCache :: ∀ a k v. Ord k => k -> Ref.Ref (Map k v) -> Spago a v -> Spago a v
```
f-f, line 287: This name gives me the idea that we are just fetching from the cache, but we are also computing the value if it's not there - maybe rename to something like `runWithCache`?

f-f (commented):
Brilliant refactoring of the new caching code in `Fetch` - @finnhodgkin could you test this code to see if the perf still holds up?

test-fixtures/monorepo/case3-shared-dependencies2/package-c/spago.yaml
fsoikin: But these are _test_ dependencies. The comment on the test suggests it's testing the core dependencies, and I didn't touch those.

The reason I removed these is that I wanted to make it also test how the transitive test dependencies are computed.

f-f: Sounds good 👍

src/Spago/Command/Fetch.purs
f-f: I think it's important that we reference that there's a cache involved.

fsoikin: Ok, how about `cacheFetchOrCreate`?

fsoikin: Maybe just `cached`?

f-f: `withCache`?

fsoikin: Yeah, that works

f-f (approved):
Stellar work! Thanks 👏

f-f:
Yeah I will merge #1275 now then cut a new version!

f-f:
@fsoikin 0.93.39 is out 🚀

PR #1272, opened 2024-08-23
Fail upon encountering unrecognized fields in the config

### LINKED ISSUE #1165: Issue warnings when the configuration has unrecognised fields

fsoikin:
### Description of the change

(partially) fixes #1165.

The codec library doesn't have such feature, so I basically had to reimplement the relevant functions.

Unfortunately, this change doesn't affect some sections of the config, because their types and codecs are defined in the Registry library. I am planning to contribute the change back to the codec library and then migrate the Registry library to it.

### Checklist:


docs-search/client-halogen/spago.yaml
```diff
@@ -38,7 +38,6 @@ package:
   bundle:
     type: "app"
     minify: true
-    sourceMaps: true
```
fsoikin, line 41: And immediately we have a winner!

Turns out `spago bundle` has a command-line parameter for source maps, but not a config parameter. Another bug.

fsoikin: Or not a bug? Technically this can be specified in extra args, as I did here, but this is asymmetry. Not sure what the overarching philosophy is here.

f-f: Great catch!

Overarching philosophy is that we try to keep flags and config params as close to each other as possible. There are some notable exceptions, but this one is a bug: we should have a config parameter as well.

f-f:
P.S. your commit messages for the autoformatter are extremely amusing, please keep them coming ❤️

f-f:
Code looks good - we shall wait for the upstream codec-json lib to merge the patch so we don't have to keep the library code in here

fsoikin:
Alright, upstream libraries finally updated, this should be ready now.

Sorry for the force-push, somehow my history wasn't matching GitHub, I must have rebased long ago, when I touched this PR last time.

I decided not to port:
* Codecs under `Docs.Search`, because the JSON files there are not written by the user.
* `Spago.Config.remotePackageSetCodec` - this is used to read package set from both local files and network responses, and I think that making it strict would break backward compatibility: if the server adds new fields, old clients won't be able to read responses anymore. Plus, there is no danger of typos here, because all fields are mandatory.
* `Spago.Purs.moduleGraphNodeCodec` and codecs in `Spago.Purs.Graph` - same reason. This is used for reading `purs` responses.
* `Spago.Command.Ls.formatPackagesJson.packageCodec` - used only for output.
* `Spago.Command.Registry.search.infoDataCodec` - used only for output.
* Codecs in `Spago.Psa.Types` - used to parse PSA output, same reason as for `purs` and remote registry.

spago.lock
fsoikin, line 1: I'm not sure what's going on here. Looks like it used to be JSON, but now it's YAML. I don't think I changed anything in that regard.

fsoikin: Oh, I see, this is https://github.com/purescript/spago/pull/1280

fsoikin, line 1: Ok, this is looking better now.

core/src/Config.purs
```diff
@@ -43,31 +43,33 @@ import Data.Array.NonEmpty as NonEmptyArray
...
 import Data.Codec.JSON.Record as CJ.Record
+import Data.Codec.JSON.Strict as CJS
 import Data.Codec.JSON.Sum as CJ.Sum
 import Data.Either as Either
 import Data.List as List
 import Data.Map as Map
 import Data.Profunctor as Profunctor
 import Partial.Unsafe (unsafeCrashWith)
 import Registry.Internal.Codec as Internal.Codec
+import Registry.Internal.Codec as Reg.Internal.Codec
```
thomashoneyman, line 54: This is a duplicated import

fsoikin: Oh indeed, so it is.

thomashoneyman (approved):

f-f (approved):

PR #1277, opened 2024-08-28
In PSA output label warnings and errors as such

### LINKED ISSUE #1148: Spago doesn't tell what is a warning and what is an error
A warning
```
[395/395 UnusedName] src/Websocket.purs:377:7

  377        otherwise      -> pure unit
             ^^^^^^^^^

  Name otherwise was introduced but not used.
  in value declaration receiveServerEvent
```

An error
```
[2/2 UnknownImport] src/Websocket.purs:39:26

  39  import Data.Array.Extra (firstMaybe)
                               ^^^^^^^^^^

  Cannot import value firstMaybe from module Data.Array.Extra
  It either does not exist or the module does not export it.
```

The only way to tell that it is a warning or an error is by the color. But now we immediately see the problem, as soon as you copy paste and lose terminal colors the information is gone. I would suggest the following:
`395 Warning UnusedName in src/Websocket.purs:377:7`

The `/395` and the brackets `[`, `]` don't really add much but i would be ok if they stayed

fsoikin:
### Description of the change

Fixes #1148 

The outputs now look like:

```
[WARN 1/4 ImplicitImport] src/Main.purs:3:1

[ERR 1/2 TypesDoNotUnify] src/Foo.purs:4:5
```

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ Other improvements:
   cached locally.
 - `spago publish` now allows to publish a package with some test (but only
   test!) dependencies not present in the registry.
+- errors and warnings are now explicitly labeled as "ERR" and "WARN" in Spago
+  build output.
 
 ## [0.21.0] - 2023-05-04
 
--- a/src/Spago/Psa/Printer.purs
+++ b/src/Spago/Psa/Printer.purs
@@ -5,9 +5,7 @@
 --   Copyright © Nathan Faubion
 --   https://opensource.org/license/mit/
 module Spago.Psa.Printer
-  ( renderWarning
-  , renderError
-  , renderStats
+  ( renderStats
   , renderVerboseStats
   , printDefaultOutputToErr
   , printJsonOutputToOut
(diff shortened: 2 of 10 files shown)


f-f:
Patch is good, but I wonder if we should expand the texts to be `WARNING` and `ERROR`? It's no strong opinion though, just reads better to me.

fsoikin:
I have no strong opinion either. Changed to "ERROR" and "WARNING".

test/Prelude.purs
```diff
@@ -396,5 +396,11 @@ escapePathInErrMsg = case Process.platform of
 assertWarning :: forall m. MonadThrow Error m => Array String -> Boolean -> String -> m Unit
 assertWarning paths shouldHave stdErr  = do
```
fsoikin, line 397: This function is super confusingly named for what it does and has confusing parameters too. I fixed the error message at least, but refactoring the function would be a bigger undertaking.

test/Spago/Build/Monorepo.purs
```diff
@@ -142,8 +142,8 @@ spec = Spec.describe "monorepo" do
       FS.copyTree { src: fixture "monorepo/strict-true-uncensored-warnings", dst: "." }
       let
         errs =
-          [ "[1/2 UnusedName] " <> escapePathInErrMsg [ "package-b", "src", "Main.purs:6:13" ]
-          , "[2/2 UnusedName] " <> escapePathInErrMsg [ "package-b", "test", "Main.purs:6:13" ]
+          [ "[ERROR 1/2 UnusedName] " <> escapePathInErrMsg [ "package-b", "src", "Main.purs:6:13" ]
+          , "[ERROR 2/2 UnusedName] " <> escapePathInErrMsg [ "package-b", "test", "Main.purs:6:13" ]
```
fsoikin, line 146: Not here: in strict mode warnings are presented as errors. Is that ok? Not sure.

f-f: Yeah it's good - this is sort of the point of the strict mode.

PR #1278, opened 2024-08-30
Turn off optparse backtracking to show help for the current command

### LINKED ISSUE #1146: Invalid option for a command should give the help for that command instead of the general help
```
» spago build --watch
Invalid option `--watch'

Usage: spago (COMMAND | (-v|--version))
  PureScript package manager and build tool

Available options:
  -h,--help                Show this help text
  -v,--version             Show the current version

Available commands:
  init                     Initialise a new project
  fetch                    Downloads all of the project's dependencies
  install                  Compile the project's dependencies
  uninstall                Remove dependencies from a package
  build                    Compile the project
  run                      Run the project
  test                     Test the project
  bundle                   Bundle the project in a single file
  sources                  List all the source paths (globs) for the
                           dependencies of the project
  repl                     Start a REPL
  publish                  Publish a package
  upgrade                  Upgrade to the latest package set, or to the latest
                           versions of Registry packages
  docs                     Generate docs for the project and its dependencies
  registry                 Commands to interact with the Registry
  ls                       List packages or dependencies
  graph                    Generate a graph of modules or dependencies
```

Expected the help for the `build` command instead

```
» spago build --help
Usage: spago build [--monochrome|--no-color] [--offline] [-q|--quiet]
                   [-v|--verbose] [--backend-args ARGS] [--ensure-ranges]
                   [--json-errors] [--output DIR] [--pedantic-packages] [--pure]
                   [--purs-args ARGS] [-p|--package PACKAGE] ([--verbose-stats]
                   | [--censor-stats]) [--strict]
  Compile the project

Available options:
  --monochrome,--no-color  Force logging without ANSI color escape sequences
  --offline                Do not attempt to use the network. Warning: this will
                           fail if you don't have the necessary dependencies
                           already cached
  -q,--quiet               Suppress all spago logging
  -v,--verbose             Enable additional debug logging, e.g. printing `purs`
                           commands
  --backend-args ARGS      Arguments to pass to the running script
  --ensure-ranges          Add version bounds for all the dependencies of the
                           selected project
  --json-errors            Output compiler warnings/errors as JSON
  --output DIR             The output directory for compiled files
  --pedantic-packages      Check for redundant or missing packages in the config
                           and fail the build if any
  --pure                   Use the package information from the current
                           lockfile, even if it is out of date
  --purs-args ARGS         Arguments to pass to purs compile. Wrap in quotes.
                           `--output` and `--json-errors` must be passed to
                           Spago directly.
  -p,--package PACKAGE     Select the local project to build
  --verbose-stats          Show counts for each warning type
  --censor-stats           Censor warning/error summary
  --strict                 Promotes project sources' warnings to errors
  -h,--help                Show this help text
```

fsoikin:
### Description of the change

Fixes #1146

This is kind of by design for `optparse`, even though it ends up being nonsensical.

This happens for commands that don't have any non-optional options. When the parser parses such command and encounters a mistyped option, the command parsing has already succeeded. It doesn't need any options, so it succeeds as soon as its name is parsed. And then the parser tries to parse remaining options as "global". They still fail, but by that time we're already back in "global" context. The command parsing is over. So `optparse` shows global help.

This behavior - trying to parse remaining options as "global" - is called "backtracking", and it's governed by an option. Turning backtracking off has a subtle, but not insignificant effect. It manifests for programs that have truly "global" options as well as commands.

Suppose you have a parser like this:

```
rootParser = fromRecord 
  { command: commandParser
  , anotherOption: strOption (long "foo")
  }
```

So you can run your program as `prog --foo bar command`, which will result in `{ command: ..., anotherOption: "bar" }`.
With backtracking enabled, you can also run `prog command --foo bar`. The `--foo` option will be parsed after the command.
But with backtracking disabled you can't. Supplying `--foo` after `command` will be an error, and you will be shown help for the command.

Fortunately, Spago's parser is setup in a way that doesn't have any global options. What we call `GlobalArgs` are actually options of every individual command from `optparse`'s point of view. And this means that we don't care for backtracking, so it can be safely turned off. Which is what this PR does.

### Checklist:

### DIFF
--- a/bin/src/Main.purs
+++ b/bin/src/Main.purs
@@ -20,6 +20,7 @@ import Node.Path as Path
 import Node.Process as Process
 import Options.Applicative (CommandFields, Mod, Parser, ParserPrefs(..))
 import Options.Applicative as O
+import Options.Applicative.Types (Backtracking(..))
 import Optparse as Optparse
 import Record as Record
 import Registry.PackageName as PackageName
@@ -508,6 +509,7 @@ parseArgs = do
         ( p
             { prefShowHelpOnError = true
             , prefShowHelpOnEmpty = true
+            , prefBacktrack = NoBacktrack
             }
         )
     )
--- a/bin/src/Main.purs
+++ b/bin/src/Main.purs
@@ -509,6 +509,7 @@ parseArgs = do
         ( p
             { prefShowHelpOnError = true
             , prefShowHelpOnEmpty = true
+            -- Needed to avoid things like https://github.com/purescript/spago/issues/1146
             , prefBacktrack = NoBacktrack
             }
         )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ Other improvements:
 - always using forward slash as path separator in lockfile, regardless of the
   platform, so that the lockfile doesn't keep changing when team members run
   Spago on different platforms.
+- when encountering a mistyped option for a command, Spago will show help for
+  that command, not root help.
 
 ## [0.21.0] - 2023-05-04
 
--- a/test-fixtures/1146-cli-help/build.txt
+++ b/test-fixtures/1146-cli-help/build.txt
@@ -1,0 +1,36 @@
+Invalid option `--bogus'
+
+Usage: index.dev.js build [--migrate] [--monochrome|--no-color] [--offline]
+                          [-q|--quiet] [-v|--verbose] [--backend-args ARGS]
+                          [--ensure-ranges] [--json-errors] [--output DIR]
+                          [--pedantic-packages] [--pure] [--purs-args ARGS]
+                          [-p|--package PACKAGE] ([--verbose-stats] |
+                          [--censor-stats]) [--strict]
+  Compile the project
+
+Available options:
+  --migrate                Migrate the spago.yaml file to the latest format
+  --monochrome,--no-color  Force logging without ANSI color escape sequences
+  --offline                Do not attempt to use the network. Warning: this will
+                           fail if you don't have the necessary dependencies
+                           already cached
+  -q,--quiet               Suppress all spago logging
+  -v,--verbose             Enable additional debug logging, e.g. printing `purs`
+                           commands
+  --backend-args ARGS      Arguments to pass to the running script
+  --ensure-ranges          Add version bounds for all the dependencies of the
+                           selected project
+  --json-errors            Output compiler warnings/errors as JSON
+  --output DIR             The output directory for compiled files
+  --pedantic-packages      Check for redundant or missing packages in the config
+                           and fail the build if any
+  --pure                   Use the package information from the current
+                           lockfile, even if it is out of date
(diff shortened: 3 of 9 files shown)


f-f:
> Fortunately, Spago's parser is setup in a way that doesn't have any global options. What we call `GlobalArgs` are actually options of every individual command from optparse's point of view. And this means that we don't care for backtracking, so it can be safely turned off. Which is what this PR does.

I always had an issue with all of this - the context is:
- in the Haskell implementation, we had "real global options" - they were printed out nicely and separately from the command options, but needed to be parsed _before_ the command, which was very confusing as it meant you couldn't e.g. call `spago build -v`
- so in this PS implementation I structured the parser so that there are no global options, so you can put flags wherever. But the formatting of the help message is very confusing

So I'd like to end up in a place where:
- we can use global options in any position of the command line parser
- and we also get a help message which separates the command options from the global options

The concrete scenarios would be:
1. we disable the backtracking but find a way to separate the global options in the help output
2. or we keep the backtracking, go back to global options, and find a way to enable their parsing anywhere

I have the suspicion that none of these scenarios is possible 🥲 But thought I'd bring this up with you in case you have any idea how to go at it

fsoikin:
> So in this PS implementation I structured the parser so that there are no global options, so you can put flags wherever

That's not how it works currently. With the current parser structure you can't put global args wherever - only after the command name. So you can do `spago build -q`, but you can't do `spago -q build`.

> we keep the backtracking, go back to global options, and find a way to enable their parsing anywhere

We don't have to find a way. If we go back to global options and keep backtracking, the global options will be parseable anywhere. That's what backtracking is for. It enables parsing of global options after the command name.

But then issue #1146 will remain. The only way to fix #1146 is to disable backtracking. Backtracking is what makes the parser "forget" the current command after it's been parsed satisfactorily.

> But thought I'd bring this up with you in case you have any idea how to go at it

I think this might be possible by hijacking how the help is printed. Not sure though.

Also, these limitations ultimately stem from the particular way `optparse` is implemented. I think it is needlessly waaaaay too complicated. I haven't actually tried yet, but I'm pretty sure option parsing can be done way simpler and straightforward. From what I can see, `optparse` is trying to do "regular" parsing, similar to how you'd parse a program source code, but CLI options are not sequential by their very nature, so the parser would be better off with random access model.

f-f:
Good thread, thanks for the details. You're right that we need this to fix #1146, so I'll merge this now and we'll eventually figure out the rest

bin/src/Main.purs
```diff
@@ -508,6 +509,7 @@ parseArgs = do
         ( p
             { prefShowHelpOnError = true
             , prefShowHelpOnEmpty = true
+            , prefBacktrack = NoBacktrack
```
f-f, line 512:
```suggestion
            -- Needed to avoid things like https://github.com/purescript/spago/issues/1146
            , prefBacktrack = NoBacktrack
```

f-f (approved):

PR #1283, opened 2024-09-06
In spago publish, when registry solver returns package versions not in local cache, install them before trying to build

### LINKED ISSUE #1110: Publish rebuilds with another version

fsoikin:
### Description of the change

Fixes #1110

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ Other improvements:
   that command, not root help.
 - a new `spago init --subpackage foo` option to initialize a sub-project in the
   current workspace.
+- #1110: `spago publish` will now install packages returned by the registry solver
+  before trying to build with them.
 
 ## [0.21.0] - 2023-05-04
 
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -36,6 +36,8 @@ Other improvements:
   that command, not root help.
 - a new `spago init --subpackage foo` option to initialize a sub-project in the
   current workspace.
+- #1110: `spago publish` will now install packages returned by the registry solver
```
fsoikin, line 39: I'm not sure if it's a good idea to mention issue numbers here. I didn't before, but in this particular one it seems like the change description doesn't make sense without it.

f-f: Tbh we are not really paying attention to the changelog since the spago rewrite, because so many things changed from the previous stable version. I appreciate that you have been keeping track of the latest changes, but we might just start with an empty changelog once we get the new spago out of the alpha stage, so feel free to skip these entirely.

f-f (approved):
This is great, thanks!

PR #1285, opened 2024-09-16
fix: bail if purs exits non-ok

### LINKED ISSUE #1284: "failed to decode PsaResult" when purs compile exits due to OOM
Spent several hours yesterday trying to figure out why my ci job running `spago build` / `spago test` was failing despite nothing visibly changing in the environment, and the `purs compile` step exiting at random points in the build with:

```text
$ spago build --pure --strict --pedantic-packages && bun run scripts/fmt.js --check
✓ Selecting 13 packages to build:
    <snip>
Downloading dependencies...
Refreshing the Registry Index...
Building...
[  1 of 941] Compiling Unsafe.Coerce
[  2 of 941] Compiling Web.Event.Internal.Types
[  4 of 941] Compiling Type.Proxy
[  3 of 941] Compiling Web.DOM.Internal.Types
[  5 of 941] Compiling Type.Function
[  6 of 941] Compiling Type.Equality
[  7 of 941] Compiling Type.Row
<snip>
[xxx of 941] Compiling xxx
‼ Failed to decode PsaResult at index '0': Unexpected end of JSON input
Json was: 
✓ Build succeeded.
```

Eventually I had the thought "maybe the `--memory=1G` I added to the runner is making `purs compile` exit early" which turned out to be exactly the case. Not sure if this is necessarily a bug in spago, just documenting this as an issue here for people who run into this in the future.

cakekindel:
closes #1284

### Description of the change

TODO

### Checklist:


cakekindel:
need to test this

cakekindel:
```
root@3a9a77b24532:/app# node ../spago/bin/bundle.js build

✓ Selecting 14 packages to build:
  ...
  
Downloading dependencies...
Building...
‼ Failed to decode PsaResult at index '0': Unexpected end of JSON input
Json was: 

✘ purs exited with non-ok status code: BySignal "SIGKILL"
```

cakekindel:
hmm maybe dying isn't the right thing to do?

f-f:
Uh, it looks like this uncovered a funny bug - CI is failing on [this test](https://github.com/purescript/spago/blob/b7c04832ca363cd1c43b06acf8fd3a3af9100409/test/Spago/Build.purs#L37), which should have never passed in the first place [because we are supposed to fail in that condition](https://github.com/purescript/spago/blob/b7c04832ca363cd1c43b06acf8fd3a3af9100409/src/Spago/Command/Build.purs#L70)

src/Spago/Psa.purs
```diff
@@ -60,9 +60,12 @@ psaCompile globs pursArgs psaArgs = do
 
         pure $ FO.isEmpty out'.stats.allErrors
 
-  if Array.all identity arrErrorsIsEmpty then do
+  if Array.all identity arrErrorsIsEmpty && Cmd.exitedOk result then do
```
f-f, line 63: I don't think checking for an exit code of 0 is the right thing to do here - when compilation fails because of errors in the build `purs` will exit with `1`.
Let's try to catch this earlier - e.g. I think we should be able to prevent the JSON decoding error: let's try to run the json parsing separately, and print a sensible error there

src/Spago/Psa.purs
```diff
@@ -60,9 +60,12 @@ psaCompile globs pursArgs psaArgs = do
...
     logSuccess "Build succeeded."
     pure true
+  else if Array.all identity arrErrorsIsEmpty && not (Cmd.exitedOk result) then do
```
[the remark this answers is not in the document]
cakekindel: ok no problem, ended up changing the output traversal to collect into an array of eithers to defer that logging and changed to a case expression

f-f (commented):
Code looks good - could we add a test for this? The best way I can think of is to try to pass a nonexistent flag to purs.

cakekindel:
Sorry forgot about this. Will do tomorrow!

f-f:
@cakekindel did you get around to finish this or shall I take care of the last bits? I'd like to get this merged ASAP before it gets out of date

cakekindel:
> @cakekindel did you get around to finish this or shall I take care of the last bits? I'd like to get this merged ASAP before it gets out of date

Sorry I forgot again, updated & added test!

test/Spago/Build.purs
```diff
@@ -32,6 +32,12 @@ spec = Spec.around withTempDir do
       spago [ "init", "--name", "aaa", "--use-solver" ] >>= shouldBeSuccess
       spago [ "build" ] >>= shouldBeSuccess
 
+    Spec.after (rmRf <<< _.testCwd)
+      $ Spec.it "exits when purs exits non-ok"
+      $ \{ spago, fixture } -> do
+          spago [ "init", "--name", "aaa" ] >>= shouldBeSuccess
+          spago [ "build", "--purs-args", "--non-existent" ] >>= shouldBeFailureErr (fixture "purs-not-ok.txt")
```
f-f, line 39: A couple of things:
- we can remove the `rmRf` for the temp dir since we don't need it anymore
- we can patch the fixture using the `sanitize` hook in the fixture machinery, so that we can get the Windows CI happy as well 🙂 

```suggestion
    Spec.it "exits when purs exits non-ok" \{ spago, fixture } -> do
      spago [ "init", "--name", "aaa" ] >>= shouldBeSuccess
      spago [ "build", "--purs-args", "--non-existent" ] >>=
        checkOutputs'
          { stdoutFile: Nothing
          , stderrFile: Just (fixture "purs-not-ok.txt")
          , result: isLeft
          , sanitize: String.trim >>> String.replace (String.Pattern "Usage: purs.bin") (String.Replacement "Usage: purs")
          }
```

f-f (approved):
This is good, thanks! 👏 

Let's see if I can merge with just applying the suggestion or if we need to fight CI any further

test/Spago/Build.purs
```diff
@@ -32,6 +32,16 @@ spec = Spec.around withTempDir do
...
       spago [ "build" ] >>= shouldBeSuccess
 
+    Spec.it "exits when purs exits non-ok" \{ spago, fixture } -> do
+      spago [ "init", "--name", "aaa" ] >>= shouldBeSuccess
+      spago [ "build", "--purs-args", "--non-existent" ] >>=
+        checkOutputs'
+          { stdoutFile: Nothing
+          , stderrFile: Just (fixture "purs-not-ok.txt")
+          , result: isLeft
+          , sanitize: String.trim >>> String.replace (String.Pattern "Usage: purs.bin") (String.Replacement "Usage: purs")
```
f-f, line 42:
```suggestion
          , sanitize: String.trim
              >>> String.replace (String.Pattern "Usage: purs.bin") (String.Replacement "Usage: purs")
              >>> String.replace (String.Pattern "\r\n") (String.Replacement "\n")
```

cakekindel:
Is it maybe CRLF? No diff in log output :confused:

cakekindel:
```diff
diff --git a/test/Prelude.purs b/test/Prelude.purs
index 9477ac5..a9595fc 100644
--- a/test/Prelude.purs
+++ b/test/Prelude.purs
@@ -7,6 +7,7 @@ import Spago.Prelude
 
 import Data.Array as Array
 import Data.Map as Map
+import Data.Newtype (wrap)
 import Data.String (Pattern(..), Replacement(..))
 import Data.String as String
 import Effect.Aff as Aff
@@ -102,11 +103,11 @@ shouldEqualStr v1 v2 =
     fail $ Array.intercalate "\n"
       [ ""
       , "===== (Actual)"
-      , v1
+      , String.replaceAll (wrap "\r\n") (wrap "CRLF\n") v1
       , "====="
       , "  ≠"
       , "===== (Expected)"
-      , v2
+      , String.replaceAll (wrap "\r\n") (wrap "CRLF\n") v2
       , "====="
       , ""
       ]
```

VirtualBox_w10_02_11_2024_16_24_19

cakekindel:
why is this not an issue with the other fixtures???

cakekindel:
> why is this not an issue with the other fixtures???

Because purs' error output is printed to stderr, which is inherited for the reasons mentioned in `Spago.Purs.compile` :upside_down_face:

That makes so much sense now

f-f:
@cakekindel could I ask you to not force push? It makes it harder to follow the changeset over the history of the PR.
We already have a way to deal with Windows newlines, and that that's through the `sanitize` hook; I pushed something that should approximate a fix [in this commit](https://github.com/purescript/spago/pull/1285/commits/bab00b5608d495d957373970a0e147375ad11b4a). We use it in a similar way across all the tests, and I'd like to keep that consistent

cakekindel:
> @cakekindel could I ask you to not force push? It makes it harder to follow the changeset over the history of the PR.
Understood! I can definitely respect this, but also ask in return for patience / understanding as my instinct is not to prioritize a linear history on working branches so I may rebase onto upstream rather than merge by mistake.

> We already have a way to deal with Windows newlines, and that that's through the sanitize hook; I pushed something that should approximate a fix in this commit. We use it in a similar way across all the tests, and I'd like to keep that consistent (we do use separate fixtures for windows in other places, but it's best to avoid those if possible, as they are annoying to regenerate)
I saw this after I realized I was out of date but the test still failed for some reason.

I'll revert my commits on top of yours and try to get your solution passing on windows, sorry 😔

test/Prelude.purs
```diff
@@ -98,15 +98,20 @@ shouldEqualStr
   -> String
   -> m Unit
 shouldEqualStr v1 v2 =
+  let
+    renderNonPrinting =
+      String.replaceAll (String.Pattern "\r") (String.Replacement "␍")
+      >>> String.replaceAll (String.Pattern "\t") (String.Replacement "␉-->")
+  in
```
cakekindel, line 105: If this is contentious i can undo, just thought it would be a good unobtrusive addition to catch failures early in CI

f-f: Ah yeah, I like it! (for now, let's see how it fares over time 😂)

f-f (approved):
OMG CI passes 🎉  thank you again @cakekindel for getting this over the finish line! ❤️

PR #1288, opened 2024-09-21
Automatically fill in 'publish.location' in config when absent

### LINKED ISSUE #1060: Location field autofill
Would be nice if the location field could be filled as far as it is possible. One would have to run this command `git ls-remote --get-url origin`, if the remote is not `origin` then i would further not support automatic detection.

fsoikin:
### Description of the change

Fixes #1060

* When `publish.location` is missing in the config, Spago will try to figure it out from the `origin` remote.
* If figured out successfully, the location is written back to `spago.yaml`, but the publish operation is still aborted, because now `spago.yaml` needs to be committed to Git.
* If there is no remote named `origin`, Spago gives up.
* If the `origin` remote has a GitHub URL, Spago uses the owner+repo format, otherwise - the sole URL format.

### Checklist:


src/Spago/Command/Publish.purs
```diff
@@ -476,5 +481,32 @@ locationIsInGitRemotes location = do
           Location.Git { url } -> r.url == url
           Location.GitHub { owner, repo } -> r.owner == owner && r.repo == repo
 
+inferLocationAndWriteToConfig :: ∀ a. WorkspacePackage -> Spago (PublishEnv a) Boolean
+inferLocationAndWriteToConfig selectedPackage = do
+  Git.getRemotes Nothing >>= case _ of
+    Left err ->
+      die $ toDoc err
```
f-f, line 488:
```suggestion
    Left err ->
      die $ toDoc $ "Couldn't parse the Git remotes: " <> err
```

src/Spago/Command/Publish.purs
```diff
@@ -476,5 +481,32 @@ locationIsInGitRemotes location = do
...
+              | Config.isRootPackage selectedPackage = Nothing
+              | otherwise = Just $ withForwardSlashes selectedPackage.path
+            configPath
+              | Config.isRootPackage selectedPackage = "spago.yaml"
+              | otherwise = Path.concat [ selectedPackage.path, "spago.yaml" ]
+            location
+              | String.contains (String.Pattern "github.com") origin.url =
+                  Location.GitHub { owner: origin.owner, repo: origin.repo, subdir }
+              | otherwise =
+                  Location.Git { url: origin.url, subdir }
```
f-f, line 505: All of this is "correct" in the sense that eventually we're going to support it in this way - however right now the registry only supports `GitHub` locations, with `subdir: Nothing` (for backwards compatibility with Bower and Pursuit). So I'd like to keep this code, but error out in unsupported cases (and record that in the fixtures too, so they will change once we start supporting this)

fsoikin: Oh, I didn't realize that. Does the publish command already error out if you specify a subdir? Or would you like me to add that too?

fsoikin: Done

src/Spago/Command/Publish.purs
fsoikin: Done

src/Spago/Command/Publish.purs
```diff
@@ -155,7 +155,12 @@ publish _args = do
       , "See the configuration file's documentation: https://github.com/purescript/spago#the-configuration-file"
       ]
     Just publishConfig -> case publishConfig.location of
-      Nothing -> addError $ toDoc "Need to specify a publish.location field."
+      Nothing -> do
+        addedToConfig <- inferLocationAndWriteToConfig selected
+        if addedToConfig then
+          addError $ toDoc "The `publish.location` field was empty. Spago filled it in, but the config file needs to be committed."
```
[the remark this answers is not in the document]
fsoikin: Done

f-f (approved):
Thanks!

PR #1296, opened 2024-10-25
Stricter path types for stricter path-related logic

fsoikin:
### Description of the change

**Summary of prior discussion:**
* Paths in Spago are handled in a bit of an ad-hoc way. To name a few examples: some places use local/partial paths and rely on current directory being the right one; other places construct paths out of pieces; the `cwd` value is `unsafePerformEffect`ed at init and then ends up baked into various values around the code.
* This would be a problem for implementing `spago script`.
* This would be a problem for implementing https://github.com/purescript/spago/issues/1237
* We agreed to introduce stricter types to represent paths in Spago code.

**In this pull request:**
* New module `Spago.Path`
* Three new types for logically different kinds of paths - `RootPath`, `LocalPath`, and `GlobalPath`. See comments on the types for explanation.
* All lower-level utilities, such as `Spago.FS`, take strongly typed paths as parameters. 
* This creates ripple effects, forcing the whole codebase to use strongly typed paths.
* Most command environments now have a `rootPath :: RootPath` and use that path as root for all operations, no longer relying on CWD for this purpose.
    * Some commands, such as `registry`, don't need a root path.
* Halogen apps are not using the new types. I'm not yet familiar enough with them to make the changes.

### Checklist:


src/Spago/Glob.purs
```diff
@@ -105,11 +105,10 @@ fsWalk cwd ignorePatterns includePatterns = Aff.makeAff \cb -> do
...
       let
-        gitignorePath = entry.path
         -- directory of this .gitignore relative to the directory being globbed
-        base = Path.relative cwd (Path.dirname gitignorePath)
+        base = Path.dirname entry.path `Path.relativeTo` root
 
-      try (SyncFS.readTextFile UTF8 entry.path) >>= traverse_ \gitignore -> do
+      try (FS.readTextFileSync entry.path) >>= traverse_ \gitignore -> do
         let
           gitignored = testGlob <$> (splitGlob $ gitignoreFileToGlob base gitignore)
```
fsoikin, line 113: I found this slightly concerning place here. We're checking globs against other globs in an effort to verify if this gitignore conflicts with any of the patterns in parameters, and this doesn't always work as you would expect it to.

```
> mm = require('micromatch')
> mm.matcher("**/foo/bar")("**/foo/bar")
true
> mm.matcher("**/foo/bar")("**/foo/*")
false
```

Obviously `"**/foo/*"` overlaps `"**/foo/bar"`, but the matcher can't see it, because it expects the second parameter to be a file name, not a pattern.

CONTRIBUTING.md
```diff
@@ -111,3 +111,50 @@ Learn by doing and get your hands dirty!
 [f-f]: https://github.com/f-f
 [discord]: https://purescript.org/chat
 [spago-issues]: https://github.com/purescript/spago/issues
+
+## Working with file paths
```
fsoikin, line 115: I'd like this documentation to exist, but I'm not sure this is the right place for it and I couldn't find a better one.

src/Spago/Config.purs
```diff
@@ -143,9 +144,9 @@ type PackageSet =
   }
 
 type WorkspacePackage =
-  { path :: FilePath
+  { path :: LocalPath
   , package :: Core.PackageConfig
-  , doc :: YamlDoc Core.Config
+  , doc :: Maybe (YamlDoc Core.Config)
```
fsoikin, line 149: This change is not actually part of the path types work, but is accidentally left over from my work on `spago script`, where I'd like to have synthetic workspaces without a YAML doc on the file system, so I made it optional. Use sites that actually need it will `die` if it's absent, explaining that it's an internal error (see `configDocMissingErrorMessage` in this file).

I would like to keep it here if you don't mind, since it's a small change and already made. If you insist, I can, of course, take it out and add back in when I get back to `spago script`.

test/Spago/Publish.purs
```diff
@@ -81,129 +82,129 @@ spec = Spec.around withTempDir do
...
+        (fixture "publish/1060-autofill-location/scenario-non-github/expected-spago.yaml")
+
+    Spec.it "prints error when no origin remote" \{ fixture, spago, testCwd } -> do
+      prepareProject spago fixture testCwd
+      git [ "remote", "remove", "origin" ]
+      git [ "remote", "add", "upstream", "git@github.com:foo/bar.git" ]
+      spago [ "publish", "-p", "aaa", "--offline" ] >>=
+        shouldBeFailureErr (fixture "publish/1060-autofill-location/scenario-no-origin/expected-stderr.txt")
+      checkFixture (testCwd </> "spago.yaml")
+        (fixture "publish/1060-autofill-location/project/spago.yaml")
```
fsoikin, line 207: These tests got accidentally nested under "transfer", so I unnested them.

core/src/Path.purs
```diff
@@ -0,0 +1,180 @@
...
+-- | something "global", such as registry cache, temp directory, and so on.
+newtype GlobalPath = GlobalPath String
+
+instance Show GlobalPath where
+  show (GlobalPath p) = p
+
+derive newtype instance Eq GlobalPath
+derive newtype instance Ord GlobalPath
+
+type AdHocFilePath = String
```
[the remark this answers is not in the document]
fsoikin: I did it this way on purpose. The name `FilePath` implies that it applies to any kind of path, but the whole point of this change is to institute distinctions.

spago.yaml
```diff
@@ -12,6 +12,7 @@ package:
       - WildcardInferredType
       - ImplicitQualifiedImportReExport
   dependencies:
+    - debug
```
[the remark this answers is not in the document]
fsoikin: I use it for debugging a lot, and sometimes forget to remove before committing. I already removed it while you were reviewing.

test/Spago/Publish.purs
```diff
@@ -81,129 +82,129 @@ spec = Spec.around withTempDir do
...
+
+  Spec.it "#1110 installs versions of packages that are returned by the registry solver, but not present in cache" \{ spago, fixture, testCwd } -> do
+    let
+      shouldBeFailureErr' file = checkOutputs'
+        { stdoutFile: Nothing
+        , stderrFile: Just file
+        , result: isLeft
+        , sanitize:
+            String.trim
+              >>> String.replaceAll (String.Pattern "\\") (String.Replacement "/")
```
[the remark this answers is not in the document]
fsoikin: It applies to paths now, not to strings.

src/Spago/Glob.purs
fsoikin: There has already been a fix for this. It didn't make this globbing business completely bulletproof, but it removed this particular problem that I raised in this comment.

src/Spago/Config.purs
fsoikin: What makes it dubious in your eyes?

core/src/Path.purs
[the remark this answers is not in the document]
f-f: Sure, but the current name is a bit of a mouthful. How about `RawFilePath`?

src/Spago/Config.purs
f-f: The fact that we carry a maybe and then need to discharge it at the point of use, potentially crashing.
I generally try to limit the times we terminate for not being able to proceed anymore; in that sense, carrying a Doc and only figuring out that we can't write it to disk because it has no path seems like a better way to arrange this, as it's possible to avoid crashing: no path, can't write, warn and move on

fsoikin: But that's the same thing! The only reason we need the doc at all is that we want to modify it sometimes and write it back to disk. It's not needed in any other circumstances.

Although I now realize that the way I implemented it will result in a much wider area of failure when the doc is not present.

Would it be satisfactory if I change the implementation to move the actual crash sites to just before the modification attempts?

f-f: Nevermind - let's go with the current source, but for future patches let's try to keep them self contained so they can be reviewed in the right context

.github/workflows/build.yml
```diff
@@ -45,7 +45,9 @@ jobs:
         env:
           cache-name: cache-node-modules
         with:
-          path: ~/.npm
+          path: |
+            ~/.npm
+            node_modules
```
fsoikin, line 50: While I'm debugging the test failure (have to do it on GH, it doesn't fail on my local Windows), I noticed that the `node_modules` directory isn't cached. Is this on purpose? Or an oversight?

core/src/Path.purs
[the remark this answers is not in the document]
fsoikin: Renamed to `RawFilePath`

fsoikin:
Alright, I finally figured out the Windows failure, and it turned out to be an actual potential problem, not just a silly forward vs. back slash issue. I added special handling for it and some tests.

core/src/Path.purs
```diff
@@ -0,0 +1,215 @@
...
+--     -- Windows case on the same drive:
+--     (GlobalPath "C:\\a\\b") `relativeTo` (RootPath "C:\\a\\x") == LocalPath { root: RootPath "C:\\a\\x", local: "..\\b" }
+--
+--     -- But Windows case on different drives:
+--     (GlobalPath "C:\\a\\b") `relativeTo` (RootPath "D:\\a\\x") == LocalPath { root: RootPath "C:\\a\\b", local: "" }
+--
+-- In the last case the root path `D:\\a\\x` could not be used as the root of
+-- the resulting `LocalPath`, because there is no way to build a relative path
+-- from it to `C:\\a\\b`.
+areBothAbsolutePathsOnDifferentDrives :: String -> String -> Boolean
```
fsoikin, line 208: This here is the special handling for the Windows issue.

.github/workflows/build.yml
f-f: Possibly an oversight? Let's keep it in and see if it breaks things

fsoikin: I also added an attempt to cache the Windows equivalent of `~/.npm`, though I'm not 100% sure it worked. But the lack of caching here is big part of the reason that Windows build is so much slower.

f-f (approved):
Looks great now - awesome job 👏

PR #1309, opened 2025-01-19
Docs: get package names from purs graph

### LINKED ISSUE #1065: Docs command: Fix how we get package names
Spun off from this conversation: https://github.com/purescript/spago/pull/1063/files#r1352609435

Here's what the docs html page looks like:

Screen Shot 2023-10-10 at 10 19 46 AM

There's a few issues, we'd like to address:

(1) Instead of throwing all the workspace modules under the heading `<local package>`, we ought to be able to use the actual package names for each package in the current workspace. They could still live in a separate section at the top of the page

(2) We don't need to clutter the view with version numbers.

(3) We'd like to make the implementation more robust than it is at the moment. Right now it's just directory matching on `.spago/packages`.

f-f:
Fix #1065 as described in [here](https://github.com/purescript/spago/issues/1065#issuecomment-1868942919)

This patch started with just "let's thread the graph through the docs-search machinery", but getting that done snowballed into a few more things.
Brief of changes:
- we used to get package names from folders on disk, which didn't quite work. We now compute the `purs graph` instead, and pass it in
- that entailed splitting out the graph types into their own separate module so we could avoid circular imports
- however, the `PackageName` used in those graph types (the one from the Registry) was not the same `PackageName` used in the docs-search codebase (the one from Bower), so this second one is no more
- however that Bower stuff was getting in from somewhere; we were in fact parsing `bower.json` files to figure out package dependencies. That's just wrong, so that's no more, and we do instead parse `purs.json` files, which are accurate
- in passing I noticed that we were parsing the module names from the local modules. That probably worked ok in the old architecture, but with the new monorepo architecture it's just wrong. So the code for parsing the files is no more, and we instead just pass in the package names of the workspace packages and read the modules from the graph


bin/docs-search-app.js.map
fsoikin, line 1: Should this file be committed?

docs-search/client-halogen/src/Docs/Search/App/SearchResults.purs
```diff
@@ -72,7 +73,7 @@ mkComponent
   -> MD.MarkdownIt
   -> Meta
   -> H.Component Query i o Aff
-mkComponent initialEngineState contents markdownIt { localPackageName } =
+mkComponent initialEngineState contents markdownIt {} =
```
fsoikin, line 76:
```suggestion
mkComponent initialEngineState contents markdownIt _ =
```

docs-search/client-halogen/src/Docs/Search/App/Sidebar.purs
```diff
@@ -64,7 +63,7 @@ mkComponent
   -> IsIndexHTML
   -> Meta
   -> Aff (H.Component Query i Action Aff)
-mkComponent moduleIndex@{ packageModules } isIndexHTML { localPackageName } = do
+mkComponent moduleIndex@{ packageModules } isIndexHTML {} = do
```
fsoikin, line 66:
```suggestion
mkComponent moduleIndex@{ packageModules } isIndexHTML _ = do
```

docs-search/common/src/Docs/Search/Declarations.purs
```diff
@@ -47,15 +51,18 @@ mkDeclarations scores = Declarations <<< foldr (insertDocModule scores) mempty
...
-  :: Scores
+  :: Graph.ModuleGraphWithPackage
+  -> Set PackageName
+  -> Scores
   -> ModuleName
   -> Declaration
   -> Trie Char (List SearchResult)
   -> Trie Char (List SearchResult)
-insertDeclaration scores moduleName entry@(Declaration { title: _ }) trie = foldr insertSearchResult trie (resultsForDeclaration scores moduleName entry)
+insertDeclaration moduleGraph workspacePackages scores moduleName entry@(Declaration { title: _ }) trie =
```
fsoikin, line 64: I understand this is not new, but what's up with the `title: _` pattern?

```suggestion
insertDeclaration moduleGraph workspacePackages scores moduleName entry@(Declaration _) trie =
```

docs-search/common/src/Docs/Search/Declarations.purs
```diff
@@ -201,19 +211,15 @@ getLevelAndName (Declaration { info, title }) =
...
-    topLevelDir <- dirs !! 0
-    if topLevelDir == ".spago" then Package <<< PackageName <$> dirs !! 2
-    else do
-      bowerDirIx <- Array.findIndex (_ == "bower_components") dirs
-      Package <<< PackageName <$> dirs !! (bowerDirIx + 1)
-  where
-  dirs = String.split (Pattern "/") name
+extractPackageName :: Graph.ModuleGraphWithPackage -> Set PackageName -> ModuleName -> PackageInfo
+extractPackageName moduleGraph workspacePackages (ModuleName moduleName) =
+  case String.split (Pattern ".") moduleName !! 0 == Just "Prim" of
```
fsoikin, line 216:
```suggestion
  case moduleName # startsWith "Prim." of
```

docs-search/common/src/Docs/Search/Meta.purs
```diff
@@ -6,23 +6,21 @@ import Data.Codec.JSON as CJ
 import Data.Codec.JSON.Record as CJ.Record
 import Docs.Search.Config as Config
 import Docs.Search.Loader as Loader
-import Docs.Search.Types (PackageName)
-import Docs.Search.Types as Package
 import Effect.Aff (Aff, catchError)
 
 type Meta =
-  { localPackageName :: PackageName
+  {
```
fsoikin, line 12: Is this going to be used at a later point? Or can it be removed?

docs-search/index/src/Docs/Search/IndexBuilder.purs
```diff
@@ -58,33 +62,32 @@ import Node.FS.Stats (isDirectory, isFile)
...
 
   checkDirectories cfg
 
   liftEffect do
     log "Building the search index..."
 
-  docsJsons /\ moduleNames /\ packageMetas <- sequential $
-    (\d h m -> d /\ h /\ m)
+  docsJsons /\ packageMetas <- sequential $
+    (\d m -> d /\ m)
```
fsoikin, line 88:
```suggestion
    Tuple
```

fsoikin (approved):
Looks sparkly!

bin/docs-search-app.js.map
f-f: Good question. It was already committed so I just updated it, but I don't know if it's of any use, tempted to remove it

docs-search/common/src/Docs/Search/Meta.purs
f-f: I would like to remove it but didn't want to expand the refactoring any further. I can have a go at it if you'd like

docs-search/common/src/Docs/Search/Declarations.purs
f-f: I committed this but I'm reverting it - `Prim` is a builtin module and it's not matched by the `startsWith` implementation

docs-search/common/src/Docs/Search/Meta.purs
f-f: Done

bin/docs-search-app.js.map
f-f: Done

Issue #951, opened 2023-03-22
Ignore nested workspaces


JordanMartinez:
Below is my understanding of this request based on the corresponding conversation in Discord. https://discord.com/channels/864614189094928394/865401625249448006/1084363652238413955

A monorepo is a group of packages that can compile together via some master build. A multirepo/polyrepo is a group of packages that cannot be built together; rather, they are just stored together. For context, see https://monorepo.tools/.

In the new spago, a workspace is defined by some folder that contans a `spago.yml` file. For example, this repo structure is a monorepo:
```
foo/Foo.purs
bar/Bar.purs
baz/Baz.purs
spago.yml
```

However, this repo structure is a multirepo:
```
client/
  Client.purs
  spago.yml -- multirepo package - uses JS backend
server/
  Server.purs
  spago.yml -- multirepo package - uses Erlang backend
shared/
  Shared.purs
spago.yml -- Root workspace - backend depends on usage
```

When building the root `spago.yml`, should it also consider things found in the `server/spago.yml`? There are a few ways to resolve this issue:
1. Throw an error - disable this entirely.
2. Use the workspace defined in the subdirectory? - use the root for root things and the server for server things
3. Merge the two workspaces?
4. Ignore the nested workspace

The answers are:
1. No. Throwing an error shouldn't be done so that multirepos can be supported. 
2. No. But it's not clear to me why. Perhaps because one doesn't always want to build `server` when building `shared`?
3. No. Merging the two shouldn't be done because different backends may be used between workspace and it prevents a situation where one workspace uses version A of some package and another workspace version B of the same package. PureScript doesn't support compiling two different versions of the same package as it affects type class resolution and whatnot. 
4. Yes. 

See https://github.com/purescript/spago/blob/master/spaghetto/src/Spago/Config.purs#L154-L155 for where this todo originated.

considerate:
> 2. No. But it's not clear to me why. Perhaps because one doesn't always want to build server when building shared?

The reason I'm inferring is that it's because you can only know that packages built with the same compiler and backend together. Hence, if you have the following directory structure and `package-b` declares a different workspace than the client workspace `spago.yaml` it's not obvious what compiler to use when compiling `package-a` assuming `package-a` is a dependency of `package-b`.

- If you use the compiler specified in the client workspace you can't be sure it will work with `package-b`.
[...]

PR #1310, opened 2025-01-19
Run Spago from a nested directory other than workspace root

### LINKED ISSUE #1237: In monorepos, allow running commands from package directories

fsoikin:
### Description of the change

Fixes #1237
Allowing Spago to be launched from a directory nested within the workspace.

* When launched, Spago will walk directories up until it finds a `spago.yaml` containing a workspace.
* As it goes, it remembers config docs it finds and loads along the way.
    * We have to load them anyway in order to see if they contain a workspace. So we remember them in order to avoid loading them twice.
* After finding the root, it then globs for `spago.yaml` files back down the tree and loads those that it hasn't loaded while walking up.
* As it walks up the tree, it also gathers and remembers any `spago.yml` (wrong extension) files.
* As it walks up the tree, it also remembers the first `spago.yaml` it sees. If this is not the root project, it will become "selected".
    * I'm not 100% on this. Previous behavior has been to never "select" a project unless there is a `-p` option. But on the other hand, it's very likely that if your CWD is a particular project, you're probably working on just that one.
* If it goes all the way up to the file system root, it prints an error saying no config found and lists all the `yml` files found along the way, if any, suggesting that they may have the wrong extension.

### Checklist:


test/Prelude.purs
```diff
@@ -49,9 +49,10 @@ withTempDir = Aff.bracket createTempDir cleanupTempDir
   where
   createTempDir = do
     oldCwd <- Paths.cwd
-    temp <- Path.mkRoot =<< mkTemp' (Just "spago-test-")
-    FS.mkdirp temp
-    Paths.chdir temp
+    temp' <- mkTemp' (Just "spago-test-")
+    FS.mkdirp temp'
+    Paths.chdir temp'
+    temp <- Path.mkRoot =<< Paths.cwd
```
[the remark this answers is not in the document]
fsoikin: I asked one of my coworkers to check on his Mac, same thing happened to him 🤔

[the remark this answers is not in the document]
fsoikin: We actually narrowed it down to just `process.chdir("/var"); console.log(process.cwd())`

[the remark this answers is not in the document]
f-f: Ah yeah. We had to deal with this stuff at work - handling of tmp folders on macOS is weird.

It looks like your change fixes it. Shall we have a dedicated small test about it?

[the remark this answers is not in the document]
fsoikin: Well, this code is already test code, so if it breaks, _some_ test will break. And if no tests break, then it wasn't necessary in the first place. I don't think there needs to be a dedicated test.

src/Spago/Config.purs
```diff
@@ -174,130 +175,212 @@ type ReadWorkspaceOptions =
...
+        , doc: Just workspace.doc
+        , workspaceConfig: workspace.config
+        , rootPackage: workspace.rootPackage
+        }
+    }
+  where
+  readConfig' :: ∀ s. _ -> State.StateT s _ _
+  readConfig' = State.lift <<< readConfig
+
+  walkDirectoriesUpFrom dir = do
```
f-f, line 239: I find the State monad always super confusing - let's have full type signatures on these internal functions so they are easier to figure out

src/Spago/Config.purs
```diff
@@ -174,130 +175,212 @@ type ReadWorkspaceOptions =
...
 
-  -- We prune any configs that use a different workspace.
-  -- For reasoning, see https://github.com/purescript/spago/issues/951
-  let configPathsWithWorkspaces = otherPackages # Array.mapMaybe \readResult -> readResult.packagePath <$ readResult.config.yaml.workspace
-  unless (Array.null configPathsWithWorkspaces) do
-    logDebug $ "Found these paths with workspaces: " <> String.joinWith ", " (Path.quote <$> configPathsWithWorkspaces)
+    -- Traversing directories (not files) and doing it in sorted order ensures
+    -- that parent directories come before their subdirectories. That way we
+    -- can remember workspaces that we find along the way and avoid trying to
+    -- load their subprojects that come later.
```
f-f, line 278: We should still mention https://github.com/purescript/spago/issues/951 somewhere in this area.

src/Spago/Config.purs
```diff
@@ -174,130 +175,212 @@ type ReadWorkspaceOptions =
...
+
+  { workspace, rootPath } /\ upTree@{ closestPackage } <-
+    State.runStateT (walkDirectoriesUpFrom cwd)
+      { loadedPackages: Map.empty, misnamedConfigs: [], closestPackage: Nothing }
+
+  { loadedPackages } <-
+    State.execStateT (loadSubprojectConfigs rootPath)
+      { loadedPackages: upTree.loadedPackages, blockedSubtrees: [] }
+
+  migrateConfigsWhereNeeded rootPath loadedPackages
```
f-f, line 193: There was a comment about this in the previous version - would be still good to mention that we only migrate the configs after we chose which subtrees/configs to load

src/Spago/Config.purs
```diff
@@ -174,130 +175,212 @@ type ReadWorkspaceOptions =
...
+              State.modify_ \s -> s { loadedPackages = Map.insert dir { package, config } s.loadedPackages }
+            Right _ -> do
+              logWarn $ "Neither workspace nor package found in " <> Path.quote configFile
+
+  tryReadConfigAt path = do
+    exists <- FS.exists path
+    if exists then
+      Just <$> do
+        logDebug $ "Loading spago.yaml at " <> Path.quote path
+        readConfig' path >>= rightOrDieWith \errLines ->
```
f-f, line 309: This function is called in `walkDirectoriesUpFrom`, and the way I'm reading this is that a parse failure will terminate execution? Currently (in `readWorkspaceConfig`) we just log a warning and keep going, don't think we should be this harsh

src/Spago/Config.purs
```diff
@@ -174,130 +175,212 @@ type ReadWorkspaceOptions =
...
+     }
+  -> Spago (Registry.RegistryEnv a) (Maybe WorkspacePackage)
+determineSelectedPackage { explicitlySelected, inferredFromCwd, rootPackage, loadedPackages } = do
+  let
+    inferredFromCwd' =
+      -- Do not auto-select the root package if Spago is being run from the root directory.
+      if inferredFromCwd == rootPackage then Nothing else inferredFromCwd
+
+    selectedName =
+      explicitlySelected <|> inferredFromCwd'
```
f-f, line 368: As you said in the initial comment, I'm not sure we should infer the selected package from the cwd. On one hand we have always refrained from auto-selecting packages (except in very specific cases), on the other hand this looks quite sensible?

We do not auto-select if we are in the workspace folder, where it makes sense to build the whole thing. If you are in a subdirectory I think it's safe to assume that you want to refer to the package where you currently are.

All of this to say: yes sure why not

src/Spago/Config.purs
```diff
@@ -334,18 +439,26 @@ readWorkspace { maybeSelectedPackage, pureBuild, migrateConfig } = do
...
-  -- Read in the package database
+loadPackageSet
+  :: ∀ a
+   . { workspaceConfig :: Core.WorkspaceConfig
+     , loadedPackages :: Map PackageName WorkspacePackage
+     , rootPath :: RootPath
+     , lockfile :: Either String Lockfile
+     }
+  -> Spago (Registry.RegistryEnv a) { packageSet :: PackageSet, compiler :: Range }
+loadPackageSet { lockfile, workspaceConfig, loadedPackages, rootPath } = do
```
f-f, line 450:
```suggestion
loadPackageSet { lockfile: eitherLockfile, workspaceConfig, loadedPackages, rootPath } = do
```

src/Spago/Config.purs
```diff
@@ -334,18 +439,26 @@ readWorkspace { maybeSelectedPackage, pureBuild, migrateConfig } = do
...
   { offline } <- ask
-  packageSetInfo <- case maybeLockfileContents, workspace.packageSet of
+  packageSetInfo <- case lockfile, workspaceConfig.packageSet of
```
f-f, line 452: (we tend to use [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) for ease of brain-typechecking)
```suggestion
  packageSetInfo <- case eitherLockfile, workspaceConfig.packageSet of
```

src/Spago/Config.purs
```diff
@@ -334,18 +439,26 @@ readWorkspace { maybeSelectedPackage, pureBuild, migrateConfig } = do
...
     _, Nothing -> do
       logDebug "Did not find a package set in your config, using Registry solver"
       pure Nothing
 
     -- If there's a lockfile we don't attempt to fetch the package set from the registry
     -- repo nor from the internet, since we already have the whole set right there
-    Right lockfile, _ -> do
+    Right lf, _ -> do
```
f-f, line 459: (sorry, low-letter-count variables trip my brain)
```suggestion
    Right lockfile, _ -> do
```

src/Spago/Config.purs
fsoikin: I don't understand this one. Mention how?

src/Spago/Config.purs
fsoikin: I don't think that's the right behavior. Would it really be ok to potentially operate on a whole different workspace just because we couldn't read one of the configs in this one?

fsoikin: Also, this is a bit apples to oranges. In the previous version we do, in fact, error out if we can't read the root config: https://github.com/purescript/spago/blob/c723832d1474afac9efe6256a8e8feb4c85b567e/src/Spago/Config.purs#L198-L206

And only on nested configs we do the warning+ignore thing.

But here we don't know which one is the root one until we have read it successfully. So it's a different situation.

On the other hand, after we discover the root one and start reading back down the tree, we keep the previous behavior: warning+ignore. Except it was changed as a result of discussion in #1316 to warning+ignore the whole subtree.

src/Spago/Config.purs
fsoikin: I will change this since you asked, but this is actually how it's in the final result: https://github.com/fsoikin/spago/blob/3c2b1ce92a5d9d1edbd236181eb821cabafd9c6a/src/Spago/Config.purs#L136-L139

src/Spago/Config.purs
fsoikin: And this was due to potential shadowing of the parameter that you asked to rename to `eitherLockfile`

src/Spago/Config.purs
fsoikin: I added the signatures, but I am not at all sure they make it any clearer. If anything, I think they just create noise. Let me know if you like them.

src/Spago/Config.purs
fsoikin: Added the comment.

src/Spago/Config.purs
fsoikin: Renamed.

src/Spago/Config.purs
fsoikin: Renamed.

src/Spago/Config.purs
```diff
@@ -174,130 +175,212 @@ type ReadWorkspaceOptions =
...
+
+  migrateConfigsWhereNeeded rootPath loadedPackages
+
+  packagesByName <-
+    Map.fromFoldable <$>
+      for (Map.toUnfoldable loadedPackages :: Array _) \(path /\ { package, config }) -> do
+        hasTests <- FS.exists (path </> "test")
+        let
+          wsp :: WorkspacePackage
+          wsp = { package, path: path `Path.relativeTo` rootPath, doc: Just config.doc, hasTests }
```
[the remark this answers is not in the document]
fsoikin: Renamed.

fsoikin:
All feedback addressed except the request to mention #951, because I didn't understand what it meant.

peterbecich:
Thank you @fsoikin for implementing this feature!

src/Spago/Config.purs
f-f: There currently is a comment that links to that one - I think it's useful to leave a paper trail if someone gets in the code and comes up with the question "why are we excluding some packages"

fsoikin: Added the previously existing comment. Also added a mention of discussion in https://github.com/purescript/spago/pull/1316#issuecomment-2629011927

src/Spago/Config.purs
f-f: My issue with State is that if the program is large enough then you lose track of what's modifying the State and your functional program becomes a side-effectful one (Refs have the same problem of course). I've been in a codebase like that and it's no fun. Having the type signatures on these funs helps to distinguish what is stateful and what is not.

fsoikin: Ah, well, yes, that is a fair concern, which I share. And this is why these functions are small and limited in scope.

src/Spago/Config.purs
f-f: Ah I see, thanks for the explanation!

test-fixtures/publish/1110-solver-different-version/spago.yaml
```diff
@@ -3,7 +3,7 @@ package:
   dependencies:
     - console: ">=6.0.0 <7.0.0"
     - effect: ">=4.0.0 <5.0.0"
-    - prelude: ">=6.0.1 <7.0.0"
+    - prelude: ">=6.0.2 <7.0.0"
```
fsoikin, line 6: This change and equivalent changes in other test fixtured are not related to the PR as such, but are fixing a different issue.

* Prelude 6.0.2 was published on February 15th. It is included in package set 63.3.0.
* Because of this, when building with registry solver, Spago pulls in this new version of prelude.
* These projects, however, explicitly specified prelude 6.0.1 as well as package set versions that don't have prelude 6.0.2 in them.
* As a result, building with solver and building with package set produces different build plans, which in turn causes `spago publish` to produce the compiler progress output, even if `spago build` was run immediately before it, and this compiler progress output was unexpected in test fixtures.
* To fix the issue, I upgraded the test fixtures to use prelude 6.0.2 and a corresponding package set, so that build plans with and without solver will match for these projects.
* This is not a good solution though, only a bandaid "for now". It will break again as soon as new packages are published. I'm not 100% sure what a good solution would be. Perhaps make the registry solver able to use a past snapshot of the registry? Or, perhaps, this is actually good: this way Spago development never goes stale and keeps up with new stuff in new package sets.

peterbecich: @fsoikin , can this be merged as-is? When an updated package set causes test breakage in the future, I suggest we continue patching until a permanent solution is found. The value-add of this PR is already big, as it is. Thanks again

f-f (approved):
Sorry I got to this only now - thanks for the massive effort! ❤️

peterbecich:
Yes, seconded, thank you @fsoikin.

@f-f , can we tag a new release `0.93.44` including this?

f-f:
@peterbecich published 0.93.44

peterbecich:
thanks @f-f !

Issue #1311, opened 2025-01-20
Feature request: merge workspace `build` section to apply to all packages


noisyscanner:
We would like to set `censorProjectWarnings` for all packages in our monorepo (and probably `strict: true` at a later date as well)
As far as I can see this isn't currently possible? We can only set `censorLibraryWarnings` in `workspace.buildOpts`.

What about having something like a "base" package section in the workspace `spago.yaml` that could specify things like this that would apply to all packages by default, like `build.strict` and `build.censorProjectWarnings`?

Sub-packages could always override it later in their own package `spago.yaml` files..

f-f:
@noisyscanner I think something like this would make sense. I'll mark it as up for grabs since it should be pretty straightforward to implement

noisyscanner:
Am working on this atm, should have something pretty soon

PR #1316, opened 2025-01-26
censorProjectWarnings in monorepo workspace `spago.yaml`

noisyscanner:
Closes https://github.com/purescript/spago/issues/1311

### Description of the change

Allows `censorProjectWarnings` to be specified in a workspaces's `spago.yaml` and apply to all packages in a workspace.
Can override it in a package's own `spago.yaml` either to `all` or `[]` to censor no warnings for that particular workspace.

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -975,6 +975,7 @@ $ node -e "import('./index.js').then(m => console.log(m.main))"
 ```
 
 Spago does not wrap the entirety of the bundler's API (esbuild for JS builds), so it's possible to pass arguments through to it. E.g. to exclude an NPM package from the bundle you can pass the `--external` flag to esbuild:
+
 - either through the command line, with the `--bundler-args` flag, i.e. `--bundler-args "--external:better-sqlite3"`.
 - or by adding it to the configuration file:
   ```yaml
@@ -982,6 +983,8 @@ Spago does not wrap the entirety of the bundler's API (esbuild for JS builds), s
     bundle:
       extra_args:
         - "--external:better-sqlite3"
+  ```
+
 ### Enable source maps
 
 When bundling, you can include `--source-maps` to generate a final source map for your bundle.
@@ -1083,6 +1086,7 @@ Library authors will often build "ecosystems" of small interdependent packages t
 If you're using the registry solver then this is not an issue, but if your project is based on a package set, then that will not contain your newly published package, since well, you just published it!
 
 You should be able to add the newly released version to your build plan by adding it to the `extraPackages` section ([see here](https://github.com/purescript/spago/issues/1215)):
+
 ```yaml
 package:
   name: next-library-to-publish
@@ -1095,6 +1099,7 @@ workspace:
   extraPackages:
     newly-published-library: 0.1.0
 ```
+
 > [!NOTE]\
 > This only works when the package you add to `extraPackages` has been published to the registry. Adding a git dependency will produce an error, as publishing to the Registry only admits build plans that only contain packages coming from the Registry.
 
@@ -1117,9 +1122,10 @@ If you are the owner of a package and you want to transfer it to another user, y
 about the new location of the repository, so that the new owner will be able to publish new versions of the package.
 
 The transfer procedure is automated by Spago commands, and goes as follows:
-* Add your (or the new owner's) SSH public key to the `spago.yaml` through `spago auth` if they are not there already (see previous section)
-* Transfer the repository to the new owner using the hosting platform's transfer mechanism (e.g. GitHub's transfer feature)
-* Depending on whose key is present in the `owners` field, either you or the new owner will update the `publish.location` field in the `spago.yaml`, and call `spago registry transfer` to initiate the transfer. If all goes well you'll now be able to publish a new version from the new location.
+
+- Add your (or the new owner's) SSH public key to the `spago.yaml` through `spago auth` if they are not there already (see previous section)
+- Transfer the repository to the new owner using the hosting platform's transfer mechanism (e.g. GitHub's transfer feature)
+- Depending on whose key is present in the `owners` field, either you or the new owner will update the `publish.location` field in the `spago.yaml`, and call `spago registry transfer` to initiate the transfer. If all goes well you'll now be able to publish a new version from the new location.
 
 ### Know which `purs` commands are run under the hood
 
@@ -1373,7 +1379,7 @@ workspace:
       # Value 1: "all" - All warnings are censored
       all
 
-      # Value 2: `NonEmptyArray (Either String { byPrefix :: String })`
+      # Value 2: `Array (Either String { byPrefix :: String })`
       # - String values:
       #      censor warnings if the code matches this code
       # - { byPrefix } values:
@@ -1384,6 +1390,12 @@ workspace:
       # see https://yaml-multiline.info/
       - byPrefix: >
         "Data.Map"'s `Semigroup instance`
+
+    # Specify whether to censor warnings coming from the compiler
+    # for files in workspace projects and tests
+    # Optional - takes the same values as censorLibraryWarnings above
+    censorProjectWarnings: all
+    censorTestWarnings: all
 
     # Specify whether to show statistics at the end of the compilation,
     # and how verbose they should be.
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -975,13 +975,16 @@ $ node -e "import('./index.js').then(m => console.log(m.main))"
 ```
 
 Spago does not wrap the entirety of the bundler's API (esbuild for JS builds), so it's possible to pass arguments through to it. E.g. to exclude an NPM package from the bundle you can pass the `--external` flag to esbuild:
+
```
noisyscanner, line 978: My editor auto-formatted this markdown on save - can revert if preferred.

src/Spago/Psa.purs
```diff
@@ -144,23 +145,24 @@ toPathDecisions { rootPath, allDependencies, selectedPackages, psaCliFlags, cens
...
         , strict: fromMaybe false $ psaCliFlags.strict <|> (package.build >>= _.strict)
-        , censorWarnings: package.build >>= _.censorProjectWarnings
+        , censorWarnings: (package.build >>= _.censorProjectWarnings) <|> censorProjectWarnings
         }
     , toPathDecision
         { pathIsFromPackage: (testPath `Path.isPrefixOf` _)
         , pathType: IsSrc
         , strict: fromMaybe false $ psaCliFlags.strict <|> (package.test >>= _.strict)
-        , censorWarnings: package.test >>= _.censorTestWarnings
+        , censorWarnings: (package.test >>= _.censorTestWarnings) <|> censorProjectWarnings
```
noisyscanner, line 165: _to discuss_ I looked into pulling this out into a separate param like `censorTestWarnings`, but maybe it doesn't make sense having it under `buildOpts` as this is akin to `build:` in the package `spago.yaml`, maybe we'd need `testOpts` as `censorTestWarnings` is under `test:`.

If you get me

f-f: I think we should have `censorTestWarnings` under `buildOpts`. I now realise that `censorTestWarnings` is misplaced under `test`, since it's a build option.

README.md
f-f: Yes please, reformatting of the markdown documents (both the README and the CHANGELOG) are not relevant/desired and should be reverted

src/Spago/Psa.purs
noisyscanner: Thanks @f-f - have moved it under `buildOpts`.
I think moving `test.censorTestWarnings` to `build.censorTestWarnings` is probably out of scope for this PR tho, sounds like a breaking change?

noisyscanner:
Re the CI - this is failing on our fork at `spago install`  due to the current spago version not understanding these new config options.
Kinda want it to ignore these fixtures until we actually run the tests no?

Screenshot 2025-01-31 at 18 44 45

src/Spago/Psa.purs
f-f: > I think moving test.censorTestWarnings to build.censorTestWarnings is probably out of scope for this PR tho, sounds like a breaking change?

It doesn't need to be a breaking change if we migrate the config as we do [here](https://github.com/purescript/spago/blob/1060f4654b16047c3a74e34e5f0be1fd5d17b9ab/src/Spago/Config.purs#L652). But up to you, let's open a new ticket to track that if you would prefer to have it in a different patch

noisyscanner:
Isn't that error caused by it not being able to parse `package-b/spago.yaml` and therefore it doesn't think `package-b` exists at all?

```
Could not read config at path "test-fixtures/monorepo/strict-true-workspace-censored-warnings/package-b/spago.yaml"
```

src/Spago/Psa.purs
noisyscanner: Created a ticket to track it [here](https://github.com/purescript/spago/issues/1322).
I don't mind having a go at this once this PR is in - need to wrap my head around the config migration logic a bit first.

f-f:
Almost: Spago is supposed to ignore configurations that are in subfolders with respect to configurations that contain a `workspace` section. Somehow it doesn't see that config in `test-fixtures/monorepo/strict-true-workspace-censored-warnings/spago.yaml`, so it thinks that the configs for `package-a` and `package-c` belong to the current workspace.

We might have to fix this detection bug here.

src/Spago/Psa.purs
f-f: Sounds great!

f-f:
> Maybe here for `configPathsWithWorkspaces` we just need to "soft-parse" spago.yaml to know if it's a workspace or not?
> Although I suspect this would harm performance as we'd have to parse it twice in that case.
> Otherwise maybe we could make the decoder ignore unknown fields and log the warning in `readConfig`, instead of returning Left configErrors?

Hmm good points. @fsoikin any preferences/suggestions on this?

fsoikin:
Let me make sure I understand the problem correctly. In a subdirectory there may be a config file belonging it a different (presumably newer) version of Spago, which we would fail to read, and as a result consider all its projects part of the top workspace?

f-f:
@fsoikin yes - more specifically, here "fail to parse" means that we reject extra fields that are present in new versions of the config file, which we parsed correctly before #1272.

I wouldn't like to roll that patch back, so @noisyscanner mentioned, we have two options here:
1. use a non-strict parser when collecting configs in the tree
2. or just parse the yaml enough to detect the `workspace` key

I am leaning towards option 2 since it would be cheap both performance-wise and code-wise, but looking for things I might have missed here.

fsoikin:
I see a few more options.

1. Do what we currently do. The CI failure can be fixed by adding a dummy `spago.yaml` in `./test-fixtures/` with a workspace in it. That will stop Spago from parsing down into the fixture directories, which it shouldn't be doing anyway. The advantage of this option is that it's free.
2. Distinguish between "strict" and "soft" parsing errors, returning `Tuple SoftErrors (Either HardErrors Config)` from `readConfig`, and then consider only "hard" ones for the purposes of detecting a workspace. This would be quite a big lift. I'm not even sure if `Data.Codec.JSON` can support that. _(this is your option 1)_
3. Distinguish between "complete" and "preliminary" parsing, returning something like `Either HardErrors { package :: Maybe (Either Errors PackageConfig), workspace :: Maybe (Either Errors WorkspaceConfig) }`, and then only check up to `isJust _.workspace` for the purposes of detecting a workspace. This would be a smaller lift, but still would make the code quite awkward. _(this is your option 2)_
4. Die upon encountering an unparseable config while globbing down the tree. This is easy to implement: just replace `logWarn` with `die`.
5. Warn on unparseable config, but still don't include any projects down the tree from it, as if it contained a workspace. This is also easy to implement. At least in #1310, not sure about the previous state of the code.

My choice would be option 4. I think it's totally reasonable to error out in this case. What would be a real-world scenario where there is an unreadable `spago.yaml` in a subdirectory, apart from using different versions of Spago? And I think it's totally reasonable not to support different versions within the same directory subtree.

My second choice would be option 5. It's kind of like option 4, in that it may lead to errors due to unparseable `spago.yaml`, but it's more forgiving, in that it might not lead to errors. It would depend on what the current command is trying to do. Like, if you're trying to build a project that is nested down the tree from an unparseable `spago.yaml`, Spago would refuse, claiming that such project doesn't exist, and hopefully, together with the warning about the failure to parse, you'd be able to figure out what's up.

Options 2 and 3 I don't like at all. Not only are they very labour-intensive, but they also feel like band-aids, fixing just the current symptom, and not addressing the deeper problem.

noisyscanner:
Makes a lot of sense - 2 and 3 are a lot of work for what should be a fairly uncommon case, and with 4 you are going to find out straight away if your config is broken for whatever reason.

Sounded like 1 is worth doing as well either way so I added that to this PR, like you say it shouldn't be going into those dirs anyway. That should sort the CI.

As for 4, happy to also add that here or open another PR if you'd prefer

fsoikin:
There is already a PR that retractors config discovery quite a bit: https://github.com/purescript/spago/pull/1310

It would be trivial to add either option 4 or option 5 there, which I'll do as soon as we come to a consensus.

noisyscanner:
@f-f as the CI is now green due to adding the dummy `test-fixtures/spago.yaml` (thanks @fsoikin!) shall we merge this?
Since that bug isn't really related to this PR.

f-f:
@noisyscanner the bug is related to this PR: we are changing the configuration format in a way that is not forwards-compatible.

I would like Spago to be robust against this kind of configuration format evolution (which is something that we didn't do in the previous codebase, and came back to bite us), so we should handle that case. In this direction I would like to have the dummy `test-fixtures/spago.yaml` reverted, so that we can verify that whatever patch we add to fix this will handle this scenario.

@fsoikin:
> My choice would be option 4. I think it's totally reasonable to error out in this case. What would be a real-world scenario where there is an unreadable `spago.yaml` in a subdirectory, apart from using different versions of Spago? And I think it's totally reasonable not to support different versions within the same directory subtree.

My preferred option would be 5, because as I said above we should be able to fail gracefully. The "automatic traversal of subfolder" is a feature that is not welcome by everyone, so we should be careful to handle all the edge cases gracefully.
Option 4 does not feel as reasonable: why can't I build my project just because Spago found a file that it didn't like? If the config is unrelated to the current project then Spago should be able to continue execution - ignoring the whole subtree where the parse error happened while warning the user about it feels like the most reasonable option.

A real world scenario (that happened to me at some point) could be multi-backend projects that need different spago versions (and compiler versions, etc) because they are dealing with different enough ecosystems.

fsoikin:
Then it is agreed. Option 5 shall be implemented in #1310 before next sunrise.

fsoikin:
Done: https://github.com/purescript/spago/pull/1310/commits/171b93f0824473ac6b4a45eb41e9468d1a5110f0

noisyscanner:
Thanks both - I've reverted adding `test-fixtures/spago.yaml`. I guess this means this PR is now blocked by #1310 - but that's fine, it's not urgent.

f-f:
@noisyscanner some type errors from the rebase

f-f:
The main branch has moved quite a bit since last time so I took these commits to be merged in #1378

PR #1325, opened 2025-02-09
Restore type-directed search

### LINKED ISSUE #1093: `spago docs`: Fix type directed search
Type directed search doesn't work, for the browser UI. 

Searching for `a -> a` should return a list including 
`identity :: forall a. a -> a`, and as well as various monorphized functions that match like
`Int -> Int` or `Boolean -> Boolean`.

Screen Shot 2023-10-16 at 8 55 51 PM

fsoikin:
### Description of the change

Fixes #1093

The problem was that `IndexBuilder` was writing a `Maybe` value into index files, and it came out like `{ tag: "Just", value: [ ... ] }`, while the UI was expecting to read an `Array` value, not wrapped in a `Maybe`, silently failing at that, and returning empty results. I suspect the bug was introduced during migration from Argonaut to Codec, because they have different `Maybe` semantics: Argonaut writes the value itself for `Just` or `null` for `Nothing`, while Codec apparently wraps it like any other ADT.

To fix, I made the `IndexBuilder` write a naked `Array` value, coercing `Nothing` to an empty array.

I took the opportunity to also clean up and straighten the code a bit. See comments in the code.

I did not add a test for this particular issue, because the test would have to be an integration one: build docs, open browser under Selenium, try to search. I think it would be valuable to add such test, but it felt like too big of a scope for this fix.

type-search

### Checklist:

- ~[ ] Added a test for the contribution (if applicable)~

### DIFF
--- a/docs-search/common/src/Docs/Search/TypeIndex.purs
+++ b/docs-search/common/src/Docs/Search/TypeIndex.purs
@@ -7,14 +7,16 @@ module Docs.Search.TypeIndex
 
 import Prelude
 
+import Codec.JSON.DecodeError as DecodeError
 import Control.Promise (Promise, toAffE)
 import Data.Array as Array
+import Data.Bifunctor (lmap)
 import Data.Codec.JSON as CJ
-import Data.Either (hush)
+import Data.Either (Either(..))
 import Data.Foldable (fold, foldr)
 import Data.Map (Map)
 import Data.Map as Map
-import Data.Maybe (Maybe(..), fromMaybe')
+import Data.Maybe (Maybe(..))
 import Data.Newtype (class Newtype, over)
 import Data.Set (Set)
 import Docs.Search.Config as Config
@@ -27,6 +29,8 @@ import Docs.Search.TypeQuery (TypeQuery)
 import Docs.Search.TypeShape (shapeOfType, shapeOfTypeQuery, stringifyShape)
 import Effect (Effect)
 import Effect.Aff (Aff, try)
+import Effect.Aff as Error
+import Effect.Class.Console as Console
 import JSON (JSON)
 import Language.PureScript.Docs.Types (DocModule(..))
 import Registry.PackageName (PackageName)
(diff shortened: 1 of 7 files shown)


docs-search/common/src/Docs/Search/TypeIndex.purs
```diff
@@ -80,15 +84,22 @@ lookup
...
+        eitherResults = do
+          json <- eitherJson # lmap Error.message
+          CJ.decode (CJ.array SearchResult.searchResultCodec) json # lmap DecodeError.print
+
+      case eitherResults of
+        Right results ->
           pure { index: insert key (Just results) index, results }
+        Left err -> do
+          Console.error $ "Error reading type index: " <> err
+          pure { index: insert key Nothing index, results: [] }
```
fsoikin, line 102: This is where the UI is trying to read an array value from the index JSON. I did not change how it functions in principle, but added an error message to the browser console if load or parsing fails. So that at least we can see what's happening.

docs-search/common/src/Docs/Search/TypeIndex.purs
```diff
@@ -102,9 +113,8 @@ query
   :: TypeIndex
   -> TypeQuery
   -> Aff { index :: TypeIndex, results :: Array SearchResult }
-query typeIndex typeQuery = do
-  res <- lookup (stringifyShape $ shapeOfTypeQuery typeQuery) typeIndex
-  pure $ res { results = res.results }
```
fsoikin, line 107: WTF was this?!

docs-search/index/src/Docs/Search/IndexBuilder.purs
```diff
@@ -72,15 +70,16 @@ type Config =
   , generatedDocs :: String
   , workspacePackages :: Set PackageName
   , moduleGraph :: Graph.ModuleGraphWithPackage
+  , log :: String -> Aff Unit
+  , die :: String -> Aff Unit
```
fsoikin, line 74: `IndexBuilder` was logging directly to console with no indirection at all. I added these parameters so that it would respect the `--quiet` option.

Looking at the larger picture, I don't understand why this is a separate project and not part of Spago proper. It's not used from anywhere else, only from the `docs` command implementation.

src/Spago/Purs.purs
```diff
@@ -110,14 +110,21 @@ printDocsFormat = case _ of
...
   Etags -> "etags"
 
-docs :: ∀ a. RootPath -> Set LocalPath -> DocsFormat -> Spago (PursEnv a) (Either ExecaResult ExecaResult)
-docs cwd globs format = do
+docs
+  :: ∀ a
+   . { root :: RootPath
+     , globs :: Set LocalPath
+     , format :: DocsFormat
+     , quiet :: Boolean
```
fsoikin, line 118: Same here: to support the `--quiet` option.

docs-search/index/src/Docs/Search/IndexBuilder.purs
```diff
@@ -224,24 +213,23 @@ decodePursJsons { pursJsonFiles } = do
...
+      (mkHeader typeShape <> JSON.print (CJ.encode codec $ fromMaybe [] results))
   where
   mkHeader typeShape =
     "// This file was generated by docs-search\n"
       <> "window.DocsSearchTypeIndex[\""
       <> typeShape
       <> "\"] = "
 
-  codec = CJ.Common.maybe $ CJ.array SearchResult.searchResultCodec
+  codec = CJ.array SearchResult.searchResultCodec
```
fsoikin, line 232: These lines are the actual fix of the actual problem.

docs-search/common/src/Docs/Search/TypeIndex.purs
f-f: 😂

docs-search/index/src/Docs/Search/IndexBuilder.purs
f-f: It was born as a separate experiment, and was integrated only recently in Spago's codebase. It could be used independently from the command line as well (functionality that we ought to restore as a Spago command)

f-f (approved):
Lovely!!

PR #1334, opened 2025-04-02
When resolving modules in graph, glob their paths relative to their own roots

### LINKED ISSUE #1281: Pedantic packages flag ignoring local packages
With the latest version of spago, 0.93.39, we are having an issue with using pedantic packages. It is reporting that packages, which are defined in the same repo are not needed.

Our setup is 
```
- packageA
|  -- src/
| -- spago.yaml (with a workspace section)
- packageB
| -- src/
| -- spago.yaml (also with a workspace section)
```

Our `packageA` lists `packageB` as a dependency in the `package` section and also includes a reference to `packageB` as an extra package in the workspace, i.e.

```
package:
  dependcies:
    -packageB
workspace:
  extraPackages:
    packageB:
      path: ../packageB
```

With this setup, packageB is now being listed as unused, even though it is definitely used by package A.

fsoikin:
### Description of the change

Fixes #1281

The root of the issue is in how `purs graph` is annotated with package names. For every package we try to glob all its source `*.purs` files using `gitignoringGlob`, which takes a `root` parameter (formerly `cwd`). Why does it take that parameter, pray tell? Turns out that's the directory under which the globbing is to take place. The function will never go outside that directory.

And this in turn means that, if we give it the root of the workspace as the `root` parameter, it will never find any files outside the workspace! Such as, for example, `../another-package/src/File.purs` or something like that.

Initially I considered switching to [the `glob` package](https://www.npmjs.com/package/glob), which can surprisingly glob outside the root fine, but it doesn't support `.gitignore` files. Then I looked at [the `globby` package](https://www.npmjs.com/package/globby), which does support `.gitignore` files, but alas, doesn't support going outside the root.

And then it suddenly occurred to me 🤦 that I don't actually need to go outside the root. I just need to use a different root! Every given package's files are all guaranteed to live under that same package's own root, be it within the current workspace or without it. So that's what I did: every package's files are now globbed relative to its own root, and then the results are made relative to the workspace root again. Works like a charm!

### Checklist:

    - I added tests for both `spago graph` (since that was where the problem actually occurred) and `spago build --pedantic-packages` (because that's where the problem was originally reported).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,9 @@ Other improvements:
   their specified dependency ranges.
 - `spago publish` no longer tries to validate all workspace dependencies, but
   only the (transitive) dependencies of the project being published.
-- Restored broken search-directed search in generated docs.
+- Restored broken type-directed search in generated docs.
+- `spago graph modules` now correctly reports extra-packages located outside the
+  workspace root.
 
 ## [0.21.0] - 2023-05-04
 
--- a/src/Spago/Purs/Graph.purs
+++ b/src/Spago/Purs/Graph.purs
@@ -96,10 +96,22 @@ getModuleGraphWithPackage (ModuleGraph graph) = do
     $ for (Map.toUnfoldable allPackages)
         \(Tuple name package) -> do
           -- Basically partition the modules of the current package by in src and test packages
-          let withTestGlobs = if (Set.member name (Map.keys testPackages)) then OnlyTestGlobs else NoTestGlobs
+          let withTestGlobs = if (Map.member name testPackages) then OnlyTestGlobs else NoTestGlobs
           logDebug $ "Getting globs for package " <> PackageName.print name
-          globMatches :: Array LocalPath <- map Array.fold $ traverse compileGlob (Config.sourceGlob rootPath withTestGlobs name package)
-          pure $ map (\p -> Tuple p name) globMatches
+
+          -- We have to glob this package's sources relative to its own root,
+          -- not to our workspace root, because the package may be located
+          -- outside of the workspace root - e.g. if it's a local-file-system
+          -- package, - in which case the `gitignoringGlob` function won't match
+          -- any files there. That's just how it works: it walks down the tree
+          -- from the given root.
+          packageRoot <- Path.mkRoot $ Config.getLocalPackageLocation rootPath name package
+
+          Config.sourceGlob rootPath withTestGlobs name package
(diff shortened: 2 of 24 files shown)


CHANGELOG.md
```diff
@@ -49,7 +49,9 @@ Other improvements:
   their specified dependency ranges.
 - `spago publish` no longer tries to validate all workspace dependencies, but
   only the (transitive) dependencies of the project being published.
-- Restored broken search-directed search in generated docs.
+- Restored broken type-directed search in generated docs.
```
fsoikin, line 52: Oops, a typo made in a previous pull request.

src/Spago/Purs/Graph.purs
```diff
@@ -96,10 +96,22 @@ getModuleGraphWithPackage (ModuleGraph graph) = do
     $ for (Map.toUnfoldable allPackages)
         \(Tuple name package) -> do
           -- Basically partition the modules of the current package by in src and test packages
-          let withTestGlobs = if (Set.member name (Map.keys testPackages)) then OnlyTestGlobs else NoTestGlobs
+          let withTestGlobs = if (Map.member name testPackages) then OnlyTestGlobs else NoTestGlobs
```
fsoikin, line 99: WTF was this?!

src/Spago/Purs/Graph.purs
```diff
@@ -324,7 +335,7 @@ unusedError isTest selected unused = do
   { errorMessage: toDoc
       [ toDoc $ (if isTest then "Tests for package '" else "Sources for package '")
           <> PackageName.print selected.package.name
-          <> "' declares unused dependencies - please remove them from the project config:"
+          <> "' declare unused dependencies - please remove them from the project config:"
```
fsoikin, line 338: Drive-by fix.

test-fixtures/1281-local-fs-extra-packages/expected-stdout.txt
```diff
@@ -0,0 +1,9 @@
+Consumer:
+  depends:
+    - ExtraPackage
+  package: consumer
+  path: src/Main.purs
+ExtraPackage:
+  depends: []
+  package: extra-package
+  path: ../extra-package/src/Main.purs
```
fsoikin, line 9: Before this change the output of `spago graph modules` wouldn't have had this section. Packages outside the workspace root were simply ignored and didn't make it into the graph.

f-f (approved):
Brilliant, I wonder this ever worked 😄

PR #1339, opened 2025-05-04
monorepo multiple main modules documentation

peterbecich:
### Description of the change

This documents a particular use case of the monorepo style:
multiple executable sub-projects in the monorepo. They cannot all have `module Main`.

This solution is equivalent to
```
ghc-options: -main-is Foo
```
in Haskell
https://stackoverflow.com/a/70857560/1007926

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -831,30 +831,8 @@ The `--package` flag is also available for many more commands, such as `build`, 
 
 An important property of this "monorepo setup" is that the `output` folder will be shared between all the packages: they will share the same build package set (or build plan when using the solver) and they will be all build together.
 
-#### Multiple Main Modules
-
-Given
-```
-.
-├── app1
-│   ├── spago.yaml
-│   └── src
-│       └── Main.purs
-├── app2
-│   ├── spago.yaml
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -831,6 +831,31 @@ The `--package` flag is also available for many more commands, such as `build`,
...
+
+, to avoid the error `Module Main has been defined multiple times`:
+* `app1` `Main.purs` can have `module Main1`
+* `app2` `Main.purs` can have `module Main2`
+
+From the root, run:
+* `spago run -p app1 --main Main1` 
+or 
+* `spago run -p app2 --main Main2`
+
```
f-f, line 858: I find this way too verbose for the amount of information that it is providing - if you have multiple `Main` modules the issue will be apparent. So I'd just leave a note of this sort here:
```suggestion
> [!NOTE]\
> Remember that you can't have multiple modules with the same name in a single project. This usually happens with the `Main` module being defined multiple times. Rename these modules to something unique.

```

f-f (approved):

PR #1342, opened 2025-06-01
Fix byPrefix examples in the README

cwstra:
### Description of the change

Running spago 0.93.44 and purs 0.15.15, when trying out the  `byPrefix` censor example in the configuration section of the README, it doesn't actually work. 

```purescript
import Data.Array as Array
import Data.Map (Map)

--warning still happens at this declaration
fn :: Array (Map String String) -> Map String String
fn = Array.fold
```

Turns out, to properly silence this, we'd need to include the "A custom warning occurred" blurb.

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1391,7 +1391,9 @@ workspace:
       # Note: when using `byPrefix`, use the `>` for block-string:
       # see https://yaml-multiline.info/
       - byPrefix: >
-        "Data.Map"'s `Semigroup instance`
+          A custom warning occurred while solving type class constraints:
+
+              Data.Map's `Semigroup` instance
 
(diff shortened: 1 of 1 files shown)


f-f:
I think that was supposed to work and broke at some point? It looks like we don't have tests for this, and I'd rather add tests (and fix the implementation) than change the README.

f-f:
Sounds good - but to be clear I consider this a bug, in the sense that the README is correct and the implementation is wrong.
So I am not going to merge a documentation change to document the bug, and the only change that I'm willing to merge is fix+tests so that spago behaves like documented.

If you'd like to have a go at fixing it then we can keep this PR open, otherwise we should close it and open a bug report instead

PR #1357, opened 2026-01-04
Use Node.js native sqlite instead of better-sqlite3

i-am-the-slime:
Replace `better-sqlite3` with Node.js native `node:sqlite` module.

- Requires Node.js >= 22.5.0 (see [release schedule](https://nodejs.org/en/about/previous-releases))
- Named parameter syntax changed slightly (`@param` with object argument)
- Suppresses `ExperimentalWarning` in CLI and tests

### DIFF
--- a/src/Spago/NodeVersion.purs
+++ b/src/Spago/NodeVersion.purs
@@ -26,7 +26,7 @@ instance Show NodeVersionCheck where
 checkNodeVersion :: { major :: Int, minor :: Int } -> String -> NodeVersionCheck
 checkNodeVersion minimum version =
   case traverse Int.fromString (Array.take 2 parts) of
-    Just [major, minor]
+    Just [ major, minor ]
       | major > minimum.major -> NodeVersionOk
       | major == minimum.major && minor >= minimum.minor -> NodeVersionOk
       | otherwise -> NodeVersionTooOld version


wclr:
Maybe it is better to move to nodejs build-in sql support https://nodejs.org/api/sqlite.html (available from Node 22)

Spago next is a new tool and still not widely used so I don't think that there need to worry about old node version and ignore node platform development.

bin/index.dev.js
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/env -S node --no-warnings
```
f-f: Removing all the warnings is a terrible idea for a set of reasons 😄 

Let's just exclude the warnings for experimental modules: `--no-warnings=ExperimentalWarning`

src/Spago/Db.js
```diff
@@ -1,118 +1,172 @@
...
+    logger("Connecting to database at " + databasePath);
+
+    // Ensure directory exists
+    try {
+      const dir = path.dirname(databasePath);
+      if (dir) {
+        fs.mkdirSync(dir, { recursive: true });
+      }
+    } catch (err) {
+      // ignore - will fail later if needed
```
f-f: why are we ignoring the error here?

PR #1360, opened 2026-01-18
Add warnings if the lockfile doesn't match the Registry hashes

f-f:
In the same vein as https://github.com/purescript/registry-dev/pull/723, we add hash verification for the lockfile hashes: the check between tarball hash and what we store in the metadata is error-worthy, but we currently say nothing if the lockfile doesn't have good hashes. This makes sense as those hashes were initially put there for tools such as Nix.

In any case, as we are approaching the Registry-regen, it might make sense to start warning about these mismatch and make them actionable. @thomashoneyman the current warning is not very actionable, I would like to tell users to regen the lockfile, do you have any ideas on how to not make it overly scary?

### DIFF
--- a/test-fixtures/build/local-package-set-lockfile/spago.lock.new
+++ b/test-fixtures/build/local-package-set-lockfile/spago.lock.new
@@ -33,7 +33,7 @@
     "console": {
       "type": "registry",
       "version": "6.1.0",
-      "integrity": "sha256-CxmAzjgyuGDmt9FZW51VhV6rBPwR6o0YeKUzA9rSzcM=",
+      "integrity": "sha256-3T5PPz1mATZmV0C4GzkCkpLlCa5xG7u8ZNgnLrrrI1Q=",
       "dependencies": [
         "effect",
         "prelude"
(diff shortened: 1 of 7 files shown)


src/Spago/Command/Fetch.purs
```diff
@@ -246,6 +256,13 @@ fetchPackagesToLocalCache packages = do
           Left err -> die $ "Couldn't read metadata, reason:\n  " <> err
           Right versionMetadata -> do
             logDebug $ "Metadata read: " <> printJson Metadata.publishedMetadataCodec versionMetadata
+            -- Verify that the lockfile integrity matches the registry metadata
+            case Map.lookup name lockfileIntegrities of
+              Just expectedIntegrity | expectedIntegrity /= versionMetadata.hash ->
+                logWarn $ "Package " <> packageVersion
+                  <> " has a different hash in the lockfile (" <> Sha256.print expectedIntegrity
+                  <> ") than in the registry metadata (" <> Sha256.print versionMetadata.hash <> ")"
```
thomashoneyman: Potentially we could provide a link to an issue on Spago or a Discourse thread or something where we can post why this might happen, so it can be followed by users? For instance, an issue that notes the registry regen as the primary reason, and if so you can regenerate your lockfile; if that _isn't_ why it's happening, then they should open an issue so we can understand the situation. This shouldn't really happen.

f-f: I am thinking we release this together with #1362 and together with the new registry, so that lockfiles will be regenerated anyways and no such warning will pop up then.

Then we can make the warning a bit scarier since as you say it shouldn't really happen 🙂

f-f: I have updated the error message pointing to the issue tracker, but I'm not really happy with it

PR #1362, opened 2026-01-19
Remove build_plan from lockfile to improve perf

### LINKED ISSUE #1262: Super slow lockfile parsing and generating

f-f:
Fix #1262 by following the approach I explained [at the time](https://github.com/purescript/spago/issues/1262#issuecomment-2332826147) - we remove `build_plan` from each package of the lockfile since it's faster to recompute it than to decode it, as the files can get massive.

cc @finnhodgkin

### DIFF
--- a/README.md
+++ b/README.md
@@ -1399,7 +1399,7 @@ workspace:
       # Value 1: "all" - All warnings are censored
       all
 
-      # Value 2: `NonEmptyArray (Either String { byPrefix :: String })`
+      # Value 2: `Array (Either String { byPrefix :: String })`
       # - String values:
       #      censor warnings if the code matches this code
       # - { byPrefix } values:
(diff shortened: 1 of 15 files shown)


src/Spago/Command/Fetch.purs
```diff
@@ -585,27 +576,51 @@ getTransitiveDeps workspacePackage = do
...
+          computeTransitiveDeps deps =
+            let
+              transitiveDeps = transitiveClosure $ Set.fromFoldable $ Map.keys $ unwrap deps
+
+              isInTransitiveDeps :: forall v. PackageName -> v -> Boolean
+              isInTransitiveDeps name _ = Set.member name transitiveDeps
+            in
+              Map.union
+                (Map.filterWithKey isInTransitiveDeps allWorkspacePackages)
+                (map fromLockEntry $ Map.filterWithKey isInTransitiveDeps lockfile.packages)
```
fsoikin: I don't quite understand why all of this is necessary. 

Before this change, the transitive dependencies used to be initially calculated in the `Left` branch of this `case`, and then written to the lockfile, and subsequently lifted from there. This suggests that the logic for computing them already exists in some form inside the `Left` branch.

fsoikin (commented):
There is also a comment that refers to storing all transitive dependencies in the lockfile: https://github.com/purescript/spago/blob/782311b1f8e2067d100b653ef816c8d979a9f616/src/Spago/Command/Fetch.purs#L162-L164

I also noticed that there is a copy&paste of computing transitive deps for all workspace packages - [here](https://github.com/purescript/spago/blob/782311b1f8e2067d100b653ef816c8d979a9f616/src/Spago/Command/Fetch.purs#L165-L168) and [here](https://github.com/purescript/spago/blob/782311b1f8e2067d100b653ef816c8d979a9f616/src/Spago/Command/Uninstall.purs#L87-L90). Perhaps these should be consolidated?

PR #1368, opened 2026-01-22
Check if the main function exists and is exported when bundling

### LINKED ISSUE #181: Spago bundles a module without a main function
Steps to reproduce:
```
mkdir explore-bundle-no-main
cd explore-bundle-no-main
spago init
# edit the `src/Main.purs` file to remove the "main" function
spago bundle -m Main -t dist/main.js
```

Spago doesn't fail with an error that `Main.purs` does not have a function called `main`. Thus, the output of the bundle is this:
```javascript
// Generated by purs bundle 0.12.5
var PS = {};
PS["Main"].main();
```
One won't realize that the source of their issue is the compiled output until after exploring a few things more.

I only came across this because I used the wrong module name in the `spago bundle` command.

f-f:
Fix #181 (at last!!) by following my own suggestion from [here](https://github.com/purescript/spago/issues/181#issuecomment-1741506193)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1394,12 +1394,12 @@ workspace:
       #      censor warnings if the code matches this code
       # - { byPrefix } values:
       #      censor warnings if the warning's message
-      #      starts with the given text
+      #      starts with the given text.
       - CodeName
       # Note: when using `byPrefix`, use the `>` for block-string:
       # see https://yaml-multiline.info/
       - byPrefix: >
-        "Data.Map"'s `Semigroup instance`
+          Data.Map's `Semigroup` instance
 
     # Specify whether to show statistics at the end of the compilation,
     # and how verbose they should be.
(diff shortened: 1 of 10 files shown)


src/Spago/Command/Bundle.purs
```diff
@@ -146,3 +159,44 @@ nodeTargetPolyfill = Str.joinWith ";"
...
+                [ "Cannot bundle app: module " <> moduleName <> " does not declare a `main` function."
+                , "If you want to create a bundle without an entry point, use --bundle-type=module instead."
+                ]
+            EntryPoint.MainNotExported ->
+              die
+                [ "Cannot bundle app: module " <> moduleName <> " does not export `main`."
+                , "Add `main` to the module's export list, remove the explicit export list, or use --bundle-type=module."
+                ]
+            EntryPoint.ParseError err ->
+              logWarn $ "Could not verify main export: " <> err
```
fsoikin: Should this also be an error? If the module failed to parse, surely we can't bundle, right?

PR #1376, opened 2026-01-29
Implement spago update for solver projects

### LINKED ISSUE #1001: `spago update` command should work with solver-packages too
When the project build happens with the solver (rather than package sets), we should:
1. try to build a plan with the latest versions on the Registry that are still compatible with the current compiler
2. build with that plan
3. and if all is well put the new ranges in the config file

f-f:
Fix #1001: we add `spago update` support for solver-based projects. The heuristic for doing this is by "extending" the range upwards, while keeping the lower bound intact. The packages that are fixed to a single version will keep being specified as a single (upgraded) version.


thomashoneyman:
Gary had contributed this:
https://github.com/purescript/registry-dev/pull/702

It sounds like you're worried the registry will barf if it sees a range specified as "X.Y.Z" in the manifest instead of `>=X.Y.Z <X.Y.Z`?

src/Spago/Command/Upgrade.purs
```diff
@@ -30,4 +37,70 @@ run args = do
...
+            Nothing -> Nothing -- bare dep stays bare
+            Just (Core.ExactVersion _) -> Just (Core.ExactVersion newVersion) -- exact stays exact
+            Just (Core.VersionRange r)
+              | r == Core.widestRange -> Just (Core.VersionRange Core.widestRange) -- "*" stays "*"
+              | otherwise -> Just (Core.VersionRange (Range.union r (Range.caret newVersion)))
+
+      upgradedCore = upgradeConstraints currentCore corePlan
+      upgradedTest = upgradeConstraints currentTest testPlan
+
+    -- Write back to config
```
thomashoneyman: Not sure how much it matters but this goes ahead and writes the constraints immediately after solving; #1001 describes building the resolved plan first, though, and only persisting changes in the build succeeded. Should we run the build with the solved plan and only then write?

PR #1385, opened 2026-02-10
Do not request the same log line from Registry on every iteration

### LINKED ISSUE #1381: spago publish issues repeated logs while streaming from the registry
When running `spago publish`, we stream logs from the Registry job that is doing the publish.
The way this happens is that we ask for logs from the Registry, by setting a `since` field in the API request, receiving the answer, waiting some time, then asking again by setting the `since` field to the date of the latest log line we have received, so allowing for the next logs to come through.
Now, some operations in the registry pipeline take a while, so what happens is that we only receive one log line, and when asking again with that timestamp we get that lone line again, which creates repeated lines in the output.

fsoikin:
### Description of the change

Turns out the Registry understands the `since` field as "greater or equal", so that sending the last line's timestamp on the next request results in returning that same last line again.

To fix, I'm adding a millisecond to the timestamp, so that the same line doesn't qualify anymore. 

Another option could be to patch the registry to change the meaning of the `since` field to just "greater". But since milliseconds are integer anyway, 1 is the smallest difference, so the result would be the same.

I tested the change manually by publishing a package. But there is no automated test. Turns out, there are no tests for non-offline publishing at all. Adding them would require mocking the Registry API, and I didn't think it was warranted for such small and largely insignificant change.

Fixes #1381 

### Checklist:

- [ ] ~Added a test for the contribution (if applicable)~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 Bugfixes:
 * Preserve TTY properties for child processes in `spago run` (#1341)
+* Do not repeatedly request the same log line from the Registry server (#1381)
 
 ## [1.0.3] - 2026-02-01
 


f-f (approved):
No worries for Windows, it's being flaky and I am looking into it.

Thanks for the patch! 🙂

PR #1409, opened 2026-07-13
Set background-color on search input

### LINKED ISSUE #1408: Illegible search input text in `spago docs` app (low contrast) in system dark mode
The `<input>` used for searching in the `spago docs` app has `color: hsl(221.25, 21.62%, 14.51%);` with no `background-color` set:

Image

The background color for the `<input>` inherits from the default user agent stylesheet, which uses the [`Field` system color](https://drafts.csswg.org/css-color/#css-system-colors) for inputs. Because we also have
```css
:root {
  color-scheme: light dark;
}
```
in our CSS (I think this is coming from the [Pursuit styles defined in `purescript` repo](https://github.com/purescript/purescript/blob/cb3c4965c8468d26c9b14cf0319db6dbd06ee4ff/app/static/pursuit.less#L84-L86)), the resolved value of `Field` will vary depending on color scheme.

In `dark` color scheme, `Field` is a dark gray which I think varies a bit depending on browser, but in any case the dark text `color` does not respond to color scheme and the result is that the input value is illegible. Here's a screenshot from Chrome on Mac OS with system them set to dark (the input value is set to "pure"):

screenshot of illegible text in dark mode

pete-murphy:
Fixes #1408

### Description of the change

Explicitly setting `background-color: white` in the search `<input>`'s styles, to override UA stylesheet default of `background-color: Field` which has poor contrast with the input's `color: hsl(221.25, 21.62%, 14.51%)` in dark mode.

#### Before

Screenshot 2026-07-13 at 1 03 07 PM

#### After

image

### Checklist:

### DIFF
--- a/docs-search/client-halogen/src/Docs/Search/App/SearchField.purs
+++ b/docs-search/client-halogen/src/Docs/Search/App/SearchField.purs
@@ -4,7 +4,7 @@ module Docs.Search.App.SearchField where
 
 import Prelude
 
-import CSS (border, borderRadius, color, em, float, floatLeft, fontWeight, lineHeight, marginBottom, marginLeft, paddingBottom, paddingLeft, paddingRight, paddingTop, pct, px, rgb, solid, weight, width)
+import CSS (backgroundColor, border, borderRadius, color, em, float, floatLeft, fontWeight, lineHeight, marginBottom, marginLeft, paddingBottom, paddingLeft, paddingRight, paddingTop, pct, px, rgb, solid, weight, white, width)
 import Data.Maybe (Maybe(..), maybe)
 import Data.Newtype (wrap)
 import Docs.Search.URIHash as URIHash
@@ -172,6 +172,7 @@ render state =
               pursuitColor = rgb 0x1d 0x22 0x2d
               rds = px 3.0
 
+            backgroundColor white
             border solid (px 1.0) pursuitColor
             borderRadius rds rds rds rds
             color pursuitColor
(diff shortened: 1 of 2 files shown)


docs-search/client-halogen/src/Docs/Search/App/SearchField.purs
```diff
@@ -172,6 +172,7 @@ render state =
               pursuitColor = rgb 0x1d 0x22 0x2d
               rds = px 3.0
 
+            backgroundColor white
```
pete-murphy, line 175: I didn't check until after making this change, but Pursuit has different dark mode styles for its search input:
```css
.top-banner__form input {
  border: 1px solid #1d222d;
  border-radius: 3px;
  background-color: #ffffff;
  color: #1d222d;
  font-weight: 300;
  line-height: 2;
  padding: 0.21em 0.512em;
  width: 100%;
}
@media (prefers-color-scheme: dark) {
  .top-banner__form input {
    border-color: #1d222d;
    background-color: #141417;
    color: #dedede;
  }
}
```

Image

Maybe we should match that? ~~Would mean moving the style declaration to `insertStyle` in `App.purs` so we can use the media query.~~ Nevermind, should be even simpler than that, we should be able to just re-use the same styles from Pursuit CSS (and delete a lot of inline styles here) by using the same classes as Pursuit.

f-f: I think we should move to whatever Pursuit does!

